PR #6, opened 2019-03-07
[fix]http.go time.Millisecond to time.Second

hxmhlt:
Fix the bug i/o timeout when the consumer calling provider.


AlexStocks:
为何close了？

PR #75, opened 2019-06-06
Fix bug

fangyincheng:
**What this PR does**:

- return error should be Response_OK
- fix bug when go-client calling java-client by p2p

### DIFF
--- a/examples/dubbo/go-server/profiles/dev/server.yml
+++ b/examples/dubbo/go-server/profiles/dev/server.yml
@@ -42,7 +42,7 @@ services:
 
 protocols:
 -   name: "dubbo"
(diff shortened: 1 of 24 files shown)


examples/dubbo/go-server/profiles/dev/server.yml
```diff
@@ -41,12 +41,21 @@ services:
         loadbalance: "random"
 
 protocols:
+<<<<<<< HEAD
```
ralf0131: This looks like a conflict.

fangyincheng: Yes, I will fix it!

PR #82, opened 2019-06-06
Fix:issue #68

### LINKED ISSUE #68: Bug: java-client can not get error message when go-server return error
My code is as follows

**go-server**
image

**java-client**
image

The problem  is that  exception  java catched  is not contain the go error message which should be ‘_invalid params: uid, name, cert could not be empty_’.

**In fact, java-client always catch as follows**:

Failed to invoke the method Init in the service com.xxx.service.go.base.certify.api.CertifyService. Tried 1 times of the providers [10.71.8.70:20001] (1/1) from the registry 10.71.8.70:2181 on the consumer 10.71.8.70 using the dubbo version 2.6.3. Last error is: Failed to invoke remote method: Init, provider: dubbo://10.71.8.70:20001/com.xxx.service.go.base.certify.api.CertifyService?anyhost=true&app.version=1.0&application=consumer-of-certify-app&category=providers&check=false&cluster=failover&dubbo=dubbo-provider-golang-2.6.0&environment=dev&group=&interface=com.xxx.service.go.base.certify.api.CertifyService&ip=192.168.99.1&loadbalance=random&methods=Init&methods.Init.loadbalance=random&methods.Init.retries=3&methods.Init.weight=0&module=dubbogo+certify+server&name=CertifyService&organization=test.com&owner=raychen&pid=71023&register.ip=10.71.8.70&remote.timestamp=1559621785&retries=0&service.filter=echo&side=consumer&timestamp=1559622450690&version=&warmup=100, cause: 0

fangyincheng:
**What this PR does**:

Fixes #68 

### DIFF
--- a/examples/dubbo/go-client/app/user.go
+++ b/examples/dubbo/go-client/app/user.go
@@ -25,8 +25,11 @@ import (
 )
 
 import (
-	"github.com/apache/dubbo-go/config"
 	hessian "github.com/dubbogo/hessian2"
+)
+
+import (
+	"github.com/apache/dubbo-go/config"
 )
 
 type Gender hessian.JavaEnum
--- a/examples/dubbo/go-server/app/server.go
+++ b/examples/dubbo/go-server/app/server.go
@@ -26,11 +26,11 @@ import (
 )
 
 import (
-	"github.com/apache/dubbo-go/common/logger"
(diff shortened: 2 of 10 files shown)


AlexStocks:
LGTM

PR #84, opened 2019-06-08
Fix: handle java exception

fangyincheng:
**What this PR does**:

- exception should not be error of `ReadHeader ` and `ReadBody` 
- fix big package


protocol/dubbo/codec.go
```diff
@@ -88,10 +92,12 @@ func (p *DubboPackage) Unmarshal(buf *bytes.Buffer, opts ...interface{}) error {
...
 		}
 
-		p.Body = client.GetPendingResponse(SequenceType(p.Header.ID)).reply
-		if p.Body == nil {
+		pendingRsp := client.GetPendingResponse(SequenceType(p.Header.ID))
+		if pendingRsp == nil {
 			return perrors.Errorf("client.GetPendingResponse(%v) = nil", p.Header.ID)
 		}
+		response := &hessian.Response{RspObj: pendingRsp.reply}
+		p.Body = response
```
AlexStocks: to avoid a temple var, just a assignment may be better

protocol/dubbo/readwriter.go
```diff
@@ -24,7 +24,7 @@ import (
 
 import (
 	"github.com/dubbogo/getty"
-	hessian "github.com/dubbogo/hessian2"
+	"github.com/dubbogo/hessian2"
 	perrors "github.com/pkg/errors"
 )
```
AlexStocks: there should be another import clause

protocol/dubbo/codec.go
fangyincheng: ok， a temple var should be deleted.

protocol/dubbo/readwriter.go
fangyincheng: it's right. Don't need to fix.

AlexStocks: ok

AlexStocks:
LGTM

PR #106, opened 2019-06-27
Fix:Lock bug

### LINKED ISSUE #101: Benchmark test problem
**What happened**:

   There are these problems, when high concurrency and continuous stress testing: 

   - 'read tcp io timeout' is frequent
   - High CPU and memory footprint

fangyincheng:
**What this PR does**:

  fix a bug about lock when benchmark test.

Fixes #101 

  @AlexStocks 

image
When Concurrency, this lock will not be got. The resolve case is to use sync.Map.


AlexStocks:
LGTM

PR #117, opened 2019-07-02
 New Ftr: allow user set custom params to register to registry

hxmhlt:
**What this PR does**:

 New Ftr: allow user set custom params to register to registry

Fixes #

```release-note
User can set references & services params in consumer.yml & provider.yml
```

### DIFF
--- a/remoting/zookeeper/listener.go
+++ b/remoting/zookeeper/listener.go
@@ -129,14 +129,14 @@ func (l *ZkEventListener) handleZkNodeEvent(zkPath string, children []string, li
 			continue
 		}
 		// listen l service node
-		go func(node string) {
+		go func(node, childNode string) {
 			logger.Infof("delete zkNode{%s}", node)
 			if l.ListenServiceNodeEvent(node, listener) {
-				logger.Infof("delete content{%s}", n)
+				logger.Infof("delete content{%s}", childNode)
 				listener.DataChange(remoting.Event{Path: zkPath, Action: remoting.EventTypeDel})
 			}
 			logger.Warnf("listenSelf(zk path{%s}) goroutine exit now", zkPath)
-		}(newNode)
+		}(newNode, n)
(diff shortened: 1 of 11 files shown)


wongoo:
LGTM

remoting/zookeeper/listener.go
```diff
@@ -129,14 +129,14 @@ func (l *ZkEventListener) handleZkNodeEvent(zkPath string, children []string, li
 			continue
 		}
 		// listen l service node
-		go func(node string) {
+		go func(node, childNode string) {
```
gaoxinge: `zkPath` may also add to parameter?

remoting/zookeeper/listener_test.go
```diff
@@ -86,32 +87,36 @@ func TestListener(t *testing.T) {
...
 	client      *ZookeeperClient
 	changedData string
+	wait        *sync.WaitGroup
 }
 
 func (m *mockDataListener) DataChange(eventType remoting.Event) bool {
 	fmt.Println(eventType)
 	m.eventList = append(m.eventList, eventType)
 	if eventType.Content == m.changedData {
-		m.client.Wait.Done()
```
gaoxinge: Why not just change `m.client.Wati.Done` to `m.client.Close`?

hxmhlt: If I want to test another data change event, close client may not be a good idea. Instead , I define a wait group for one data change event

remoting/zookeeper/listener.go
hxmhlt: @AlexStocks why add childNode ,just use in log?

remoting/zookeeper/listener_test.go
gaoxinge: @hxmhlt Ok. I think your solution may be more suitable.

PR #118, opened 2019-07-03
avoid wg counter negative

gaoxinge:
**What this PR does**:

If we use `client.Wait.Done` in `DataChange` in the test, we may encounter the `negative watigroup counter negative`, because we invoke `client.Wait.Add` only once at

https://github.com/apache/dubbo-go/blob/1178d9a142bda19c04f49a9f584c96f68540a029/remoting/zookeeper/listener_test.go#L92

but invoke `client.Wait.Done` twice at 

https://github.com/apache/dubbo-go/blob/1178d9a142bda19c04f49a9f584c96f68540a029/remoting/zookeeper/listener_test.go#L114

and 

https://github.com/apache/dubbo-go/blob/1178d9a142bda19c04f49a9f584c96f68540a029/remoting/zookeeper/client.go#L233

So just change `client.Wait.Done` into `client.Close` in `DataChange`, then we can handle this.

### DIFF
--- a/remoting/zookeeper/listener_test.go
+++ b/remoting/zookeeper/listener_test.go
@@ -18,7 +18,6 @@
 package zookeeper
 
 import (
-	"fmt"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


AlexStocks:
@gaoxinge excellent work.

PR #130, opened 2019-07-16
Feature/Fix ListenDirEvent

aliiohs:
**What this PR does**:
change select time.After to wheel.After

Fixes #

```release-note

```

### DIFF
--- a/remoting/zookeeper/listener.go
+++ b/remoting/zookeeper/listener.go
@@ -18,6 +18,7 @@
 package zookeeper
 
 import (
+	gxtime "github.com/dubbogo/gost/time"
 	"path"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


remoting/zookeeper/listener.go
```diff
@@ -33,6 +34,10 @@ import (
 	"github.com/apache/dubbo-go/remoting"
 )
 
+var (
+	wheel = gxtime.NewWheel(gxtime.TimeSecondDuration(ConnDelay), MaxFailTimes+1) // wheel longest span is 2 minute
```
AlexStocks: If you think your time span is greater than 2 minutes, you can use time.After() directly. Because the go gc will collect a minute time span timer in time. If not, you can use wheel.After.

aliiohs: max is 45 seconds

AlexStocks: use getty.Wheel().After instead.

PR #133, opened 2019-07-22
Ftr: Hystrix filter for circuit break and service downgrade

YGrylls:
**What this PR does**:

Add an implement of filter for circuit break, based on hystrix-go, see filter/impl/hystrix_filter.go
The config of hystrix is exposed to users in the consumer config. Thus a new field FilterConf is added, just like the protocol conf existed. 
The service downgrade can also be customized by users by defining extra filters that handles the error returned from hystrix filter, an example is offered
This filter is designed for both client and server end, now only client works because temporarily filters in server end can not get the return information from callService, see protocol/dubbo/listener OnMessage(). Once they are integrated in the future, the filter can be used in server end just like in client end

Unit test is done, see filter/impl/hystrix_filter_test.go

<ul>
<li>To use the hystrix filter, add "hystrix_consumer" to the filter of references part in the consumer config</li>
<li>Config the hystrix in the filter_conf part of the consumer config file as the example does, the config can be set at either service or method level </li>
<li>An error whitelist can be defined in the config to decide what error will trigger hystrix and what will not</li>
<li>You can define fallback logic by defining customized filter, see how examples/hystrixfilter/dubbo/with-hystrix-go-client/app/example_fallback_filter.go does. Remember to add your fallback filter JUST BEFORE the hystrix filter </li>
</ul>

```release-note

```


hxmhlt:
pls resolve the conflicting files

filter/impl/hystrix_filter_test.go
```diff
@@ -0,0 +1,217 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package impl
+
+import (
+	"github.com/afex/hystrix-go/hystrix"
```
fangyincheng: format

filter/impl/hystrix_filter_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+func mockInitHystrixConfig() {
+	//Mock config
+	confConsumer = &HystrixFilterConfig{
+		make(map[string]*CommandConfigWithError),
+		"Default",
+		make(map[string]ServiceHystrixConfig),
+	}
+	confConsumer.Configs["Default"] = &CommandConfigWithError{
+		1000,
```
fangyincheng: add field name.

filter/impl/hystrix_filter.go
```diff
@@ -0,0 +1,274 @@
...
+			logger.Warnf("[Hystrix Filter]Config load failed for consumer, error is: %v , will use default", err)
+		}
+		isConsumerConfigLoaded = true
+	}
+
+	return &HystrixFilter{COrP: true}
+}
+
+func GetHystrixFilterProvider() filter.Filter {
+	if !isProviderConfigLoaded {
```
fangyincheng: use Once.

protocol/protocolwrapper/protocol_filter_wrapper_test.go
```diff
@@ -30,17 +30,42 @@ import (
 	"github.com/apache/dubbo-go/common"
 	"github.com/apache/dubbo-go/common/constant"
 	"github.com/apache/dubbo-go/common/extension"
-	"github.com/apache/dubbo-go/filter/impl"
+	"github.com/apache/dubbo-go/filter"
+	//"github.com/apache/dubbo-go/filter/impl"
 	"github.com/apache/dubbo-go/protocol"
 )
 
+//To avoid import cycle
```
fangyincheng: import cycle should not be exist.
And, this mock is meaningless, you can modify `impl.ECHO` to `"echo"`.

PR #141, opened 2019-07-27
add task pool for getty

fangyincheng:
**What this PR does**:

  add task pool for getty


config/testdata/consumer_config.yml
```diff
@@ -56,14 +56,16 @@ protocol_conf:
     session_timeout: "20s"
     pool_size: 64
     pool_ttl: 600
+    gr_pool_size: 1000
+    queue_len: 128
+    queue_number: 10
```
AlexStocks: gr_pool_size: 1000
    queue_len: 64
    queue_number: 200

config/testdata/provider_config.yml
```diff
@@ -56,14 +56,16 @@ protocol_conf:
   dubbo:
     session_number: 700
     session_timeout: "20s"
+    gr_pool_size: 1000
+    queue_len: 128
+    queue_number: 10
```
AlexStocks: gr_pool_size: 5000
queue_len: 64
queue_number: 400

config/testdata/consumer_config.yml
wongoo: set `gr_pool_size`=1000, `queue_number`=200 means there are only 5 goroutines to process tasks on one channel, which may cause latency issue. I think `gr_pool_size` should be 50 times greater than the `queue_number`.

PR #170, opened 2019-08-19
zk log change

pantianying:
**What this PR does**:

Fixes #

```release-note

```

### DIFF
--- a/registry/zookeeper/registry.go
+++ b/registry/zookeeper/registry.go
@@ -381,7 +381,7 @@ func (r *zkRegistry) registerTempZookeeperNode(root string, node string) error {
 	}
 	zkPath, err = r.client.RegisterTemp(root, node)
 	if err != nil {
-		logger.Errorf("RegisterTempNode(root{%s}, node{%s}) = error{%v}", root, node, perrors.WithStack(err))
+		logger.Warnf("RegisterTempNode(root{%s}, node{%s}) = error{%v}", root, node, perrors.WithStack(err))
 		return perrors.WithMessagef(err, "RegisterTempNode(root{%s}, node{%s})", root, node)
 	}
 	logger.Debugf("create a zookeeper node:%s", zkPath)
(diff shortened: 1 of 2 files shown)


registry/zookeeper/registry.go
```diff
@@ -381,7 +381,7 @@ func (r *zkRegistry) registerTempZookeeperNode(root string, node string) error {
 	}
 	zkPath, err = r.client.RegisterTemp(root, node)
 	if err != nil {
-		logger.Errorf("RegisterTempNode(root{%s}, node{%s}) = error{%v}", root, node, perrors.WithStack(err))
+		logger.Warnf("RegisterTempNode(root{%s}, node{%s}) = error{%v}", root, node, perrors.WithStack(err))
 		return perrors.WithMessagef(err, "RegisterTempNode(root{%s}, node{%s})", root, node)
```
hxmhlt: I think not all error's log level need set to warn. You can specify the 'node already exists' error type and set its log level to warn.

pantianying: i will fix it

registry/zookeeper/registry.go
```diff
@@ -381,7 +381,11 @@ func (r *zkRegistry) registerTempZookeeperNode(root string, node string) error {
 	}
 	zkPath, err = r.client.RegisterTemp(root, node)
 	if err != nil {
-		logger.Errorf("RegisterTempNode(root{%s}, node{%s}) = error{%v}", root, node, perrors.WithStack(err))
+		if err == zk.ErrNodeExists {
+			logger.Warnf("RegisterTempNode(root{%s}, node{%s}) = error{%v}", root, node, perrors.WithStack(err))
+		} else {
+			logger.Errorf("RegisterTempNode(root{%s}, node{%s}) = error{%v}", root, node, perrors.WithStack(err))
+		}
 		return perrors.WithMessagef(err, "RegisterTempNode(root{%s}, node{%s})", root, node)
```
hxmhlt: Is condition (err==zk.ErrNodeExists )should return perrors.WithMessagef?

pantianying: > This function will return error. And outer function will catch the error ,continue to call logger.Error. If you need to ignore the error, I think you should modify all error catch.

For now, there is no way to print this error except by calling the Interface Methods peripherally.

Issue #146, opened 2019-07-31
fatal error: concurrent map read and map write


wongoo:
**What happened**:

detail see: https://travis-ci.org/apache/dubbo-go/jobs/565782895

```
2019-07-31T03:20:27.734Z	DEBUG	directory/directory.go:126	update service name: dubbo://:@:/TEST2?cluster=mock&group=group2!
fatal error: concurrent map read and map write
goroutine 12 [running]:
runtime.throw(0x7b7b34, 0x21)
	/home/travis/.gimme/versions/go1.12.linux.amd64/src/runtime/panic.go:617 +0x72 fp=0xc0000b36f0 sp=0xc0000b36c0 pc=0x42e932
runtime.mapaccess1_faststr(0x774340, 0xc0000863c0, 0x7ad44f, 0x5, 0x1)
	/home/travis/.gimme/versions/go1.12.linux.amd64/src/runtime/map_faststr.go:21 +0x464 fp=0xc0000b3760 sp=0xc0000b36f0 pc=0x413a44
net/url.Values.Get(...)
	/home/travis/.gimme/versions/go1.12.linux.amd64/src/net/url/url.go:838
github.com/apache/dubbo-go/common.URL.GetParam(...)
	/home/travis/gopath/src/github.com/apache/dubbo-go/common/url.go:286
github.com/apache/dubbo-go/registry/directory.(*registryDirectory).toGroupInvokers(0xc000146000, 0x7ad422, 0x5, 0x7acd31)
	/home/travis/gopath/src/github.com/apache/dubbo-go/registry/directory/directory.go:164 +0x956 fp=0xc0000b3d10 sp=0xc0000b3760 pc=0x6ef9b6
github.com/apache/dubbo-go/registry/directory.(*registryDirectory).refreshInvokers(0xc000146000, 0xc000129500)
	/home/travis/gopath/src/github.com/apache/dubbo-go/registry/directory/directory.go:145 +0x1b3 fp=0xc0000b3f80 sp=0xc0000b3d10 pc=0x6eeeb3
github.com/apache/dubbo-go/registry/directory.(*registryDirectory).update(0xc000146000, 0xc000129500)
	/home/travis/gopath/src/github.com/apache/dubbo-go/registry/directory/directory.go:128 +0x187 fp=0xc0000b3fd0 sp=0xc0000b3f80 pc=0x6eec97
runtime.goexit()
[...]

PR #172, opened 2019-08-26
Fix:url.Values is not safe map

hxmhlt:
**What this PR does**:
Url.Values is not safe map, so add the lock to avoid concurrent map read and map write error

Fixes #

```release-note

```

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -69,9 +69,9 @@ type baseUrl struct {
 	Location string // ip+port
 	Ip       string
 	Port     string
-	Params   url.Values
 	//url.Values is not safe map, add to avoid concurrent map read and map write error
 	paramsLock   sync.RWMutex
+	Params       url.Values
 	PrimitiveURL string
 	ctx          context.Context
 }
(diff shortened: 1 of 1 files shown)


gaoxinge:
Does this fix https://github.com/apache/dubbo-go/issues/146?

common/url.go
```diff
@@ -283,14 +286,18 @@ func (c URL) Service() string {
...
 func (c *URL) AddParam(key string, value string) {
+	c.paramsLock.Lock()
 	c.Params.Add(key, value)
+	c.paramsLock.Unlock()
 }
 
 func (c URL) GetParam(s string, d string) string {
 	var r string
+	c.paramsLock.RLock()
 	if r = c.Params.Get(s); r == "" {
```
AlexStocks: 'len(r) == 0' instead.

common/url.go
```diff
@@ -64,11 +65,13 @@ func (t RoleType) Role() string {
...
-	Ip           string
-	Port         string
-	Params       url.Values
+	Protocol string
+	Location string // ip+port
+	Ip       string
+	Port     string
+	Params   url.Values
+	//url.Values is not safe map, add to avoid concurrent map read and map write error
+	paramsLock   sync.RWMutex
```
AlexStocks: pls place the  paramsLock before Params

AlexStocks:
LGTM

PR #190, opened 2019-09-04
Imp:  remove client from pool before closing it

AlexStocks:
**What this PR does**:

remove client from pool before closing it to ensure other task can not got it.


AlexStocks:
If there is no problems, pls merge it. @hxmhlt @fangyincheng

PR #197, opened 2019-09-10
fix MergeUrl

pantianying:
**What this PR does**:

Fixes #

```release-note

```


pantianying:
> We can not do this because referenceURL is different from serviceUrl in addtion to params , I don't know how many bug will appear after modifying in this way.

ok get it

PR #201, opened 2019-09-11
Fix:fix url params unsafe

### LINKED ISSUE #200: URL.params concurrent map read and map write problem
**What happened**:
program crash due to URL.params concurrent map read and map write problem

A paramsLock (common/url.go baseUrl struct) is introduced in recent commit to fix concurrent map read and write problem of params. However,  this problem has not been completely solved. In  function "MergeUrl", params are set without this lock guard.

hxmhlt:
**What this PR does**:
Fix fix url params unsafe completely.

Fixes #200 

```release-note

```

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -331,8 +331,6 @@ func (c URL) GetParamAndDecoded(key string) (string, error) {
 }
 
 func (c URL) GetRawParam(key string) string {
-	c.paramsLock.RLock()
-	defer c.paramsLock.RUnlock()
 	switch key {
 	case "protocol":
 		return c.Protocol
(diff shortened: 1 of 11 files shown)


common/url.go
```diff
@@ -291,20 +293,38 @@ func (c URL) Service() string {
...
+}
+
+func (c *URL) SetParam(key string, value string) {
+	c.paramsLock.Lock()
+	c.params.Set(key, value)
 	c.paramsLock.Unlock()
 }
 
+func (c *URL) RangeParams(f func(key, value string) bool) {
+	c.paramsLock.RLock()
```
AlexStocks: use defer unlock pls to defeat the for-loop panic.

PR #202, opened 2019-09-11
Add token support

xujianhai666:
**What this PR does**:
Add token support

```release-note

```

### DIFF
--- a/filter/impl/token_filter.go
+++ b/filter/impl/token_filter.go
@@ -27,7 +27,7 @@ import (
 
 import (
 	"github.com/apache/dubbo-go/common/constant"
-	"github.com/apache/dubbo-go/coconfig/service_config.gommon/extension"
+	"github.com/apache/dubbo-go/common/extension"
 	"github.com/apache/dubbo-go/filter"
 	"github.com/apache/dubbo-go/protocol"
 )
--- a/filter/impl/token_filter.go
+++ b/filter/impl/token_filter.go
@@ -1,0 +1,68 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package impl
+
+import (
+	"strings"
+)
+
+import (
+	perrors "github.com/pkg/errors"
+)
+
+import (
+	"github.com/apache/dubbo-go/common/constant"
+	"github.com/apache/dubbo-go/coconfig/service_config.gommon/extension"
+	"github.com/apache/dubbo-go/filter"
+	"github.com/apache/dubbo-go/protocol"
+)
+
+const (
+	token = "token"
+)
+
+func init() {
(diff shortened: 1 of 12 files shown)


filter/impl/token_filter_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/apache/dubbo-go/common"
+	"github.com/apache/dubbo-go/common/constant"
+	"github.com/apache/dubbo-go/protocol"
+	"github.com/apache/dubbo-go/protocol/invocation"
+)
+
+func TestTokenFilter_Invoke(t *testing.T) {
+	filter := GetTokenFilter()
+
+	url := common.NewURLWithOptions(common.WithParamsValue(constant.TOKEN_KEY, "ori_key"))
```
gaoxinge: This code may need to be modified as 

```go
url := common.NewURLWithOptions(
		common.WithParams(url.Values{}), 
		common.WithParamsValue(constant.TOKEN_KEY, "ori_key"),
	)
```

because we should new a `url.Value` before we set a key-value.

xujianhai666: thx, Happy Mid-autumn Day！

protocol/dubbo/dubbo_invoker.go
```diff
@@ -36,16 +36,29 @@ import (
...
 var Err_No_Reply = perrors.New("request need @response")
 
+var (
+	attachmentKey = []string{constant.INTERFACE_KEY, constant.GROUP_KEY, constant.TOKEN_KEY, constant.TIMEOUT_KEY}
+)
+
 type DubboInvoker struct {
 	protocol.BaseInvoker
 	client      *Client
+	attachment  map[string]string
```
fangyincheng: you don't need to add this attachment in the `DubboInvoker`. if you want to set `attachmentKey `, I think you should do it in the `token_filter`. or here:
image

filter/impl/token_filter.go
```diff
@@ -0,0 +1,68 @@
...
+
+func init() {
+	extension.SetFilter(token, GetTokenFilter)
+}
+
+type TokenFilter struct{}
+
+func (tf *TokenFilter) Invoke(invoker protocol.Invoker, invocation protocol.Invocation) protocol.Result {
+	invokerTkn := invoker.GetUrl().GetParam(constant.TOKEN_KEY, "")
+	if len(invokerTkn) > 0 {
```
fangyincheng: I think this may replace the original one. Because `map` can be read when it is not initialized.

```golang
if len(invokerTkn) > 0 {
		attachs := invocation.Attachments()
		remoteTkn, exist := attachs[constant.TOKEN_KEY]
		if !exist || !strings.EqualFold(invokerTkn, remoteTkn) {
			return &protocol.RPCResult{Err: perrors.Errorf("Invalid token! Forbid invoke remote service %s method %s ", invoker., invocation.MethodName())}
		}
	}
```

filter/impl/token_filter.go
```diff
@@ -0,0 +1,68 @@
...
+		attachs := invocation.Attachments()
+		if len(attachs) > 0 {
+			remoteTkn, exist := attachs[constant.TOKEN_KEY]
+			if exist && strings.EqualFold(invokerTkn, remoteTkn) {
+				return invoker.Invoke(invocation)
+			}
+		}
+		return &protocol.RPCResult{Err: perrors.Errorf("Invalid token! Forbid invoke remote service %s method %s ",
+			invoker, invocation.MethodName())}
```
fangyincheng: `invoker` is not a `string`.

filter/impl/token_filter.go
xujianhai666: done

filter/impl/token_filter.go
xujianhai666: done

protocol/dubbo/dubbo_invoker.go
xujianhai666: done

gaoxinge:
LGTM.

Issue #211, opened 2019-09-20
Ftr: AccessLogFilter support


flycash:
We need to support the AccessLogFilter which usually could provide more infomation for monitoring or debuging.

**What would you like to be added**:
AccessLogFilter, which is used to log the invocation request.

**Why is this needed**:
In order to monitor the invocation in Dubbo, we should log the invocation request into file. And we can use these files to debug too.

hxmhlt:
Finished.

PR #214, opened 2019-09-22
Ftr: AccessLogFilter support

flycash:
**What this PR does**:
Add AccessLogFilter

[Ftr: AccessLogFilter support](https://github.com/apache/dubbo-go/issues/211)

Fixes #

Please review the code about channel and file operation carefully. I'm new to Golang.

```release-note

```

### DIFF
--- a/filter/impl/access_log_filter.go
+++ b/filter/impl/access_log_filter.go
@@ -1,0 +1,208 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package impl
+
+import (
+	"github.com/apache/dubbo-go/common/constant"
+	"github.com/apache/dubbo-go/common/extension"
+	"github.com/apache/dubbo-go/common/logger"
+	"github.com/apache/dubbo-go/filter"
+	"github.com/apache/dubbo-go/protocol"
+	"os"
+	"reflect"
+	"strings"
+	"time"
+)
+
+const (
+	//usd in URL.
+	AccessLogKey      = "accesslog"
+	FileDateFormat    = "2006-01-02"
+	MessageDateLayout = "2006-01-02 15:04:05"
+	LogMaxBuffer      = 5000
+	LogFileMode       = 0600
+
+	// those fields are the data collected by this filter
+	Types     = "types"
+	Arguments = "arguments"
+)
+
+func init() {
+	extension.SetFilter(AccessLogKey, GetAccessLogFilter)
+}
+
+type AccessLogFilter struct {
+	logChan chan AccessLogData
+}
+
+func (ef *AccessLogFilter) Invoke(invoker protocol.Invoker, invocation protocol.Invocation) protocol.Result {
+	accessLog := invoker.GetUrl().GetParam(AccessLogKey, "")
+	logger.Warnf(invoker.GetUrl().String())
+	if len(accessLog) > 0 {
+		accessLogData := AccessLogData{data: ef.buildAccessLogData(invoker, invocation), accessLog: accessLog}
+		ef.logIntoChannel(accessLogData)
+	}
+	return invoker.Invoke(invocation)
+}
+
+// it won't block the invocation
+func (ef *AccessLogFilter) logIntoChannel(accessLogData AccessLogData) {
+	select {
+	case ef.logChan <- accessLogData:
+		return
+	default:
+		logger.Warn("The channel is full and the access logIntoChannel data will be dropped")
+		return
+	}
+}
+
+func (ef *AccessLogFilter) buildAccessLogData(invoker protocol.Invoker, invocation protocol.Invocation) map[string]string {
+	dataMap := make(map[string]string)
+	attachments := invocation.Attachments()
+	dataMap[constant.INTERFACE_KEY] = attachments[constant.INTERFACE_KEY]
+	dataMap[constant.METHOD_KEY] = invocation.MethodName()
+	dataMap[constant.VERSION_KEY] = attachments[constant.VERSION_KEY]
+	dataMap[constant.GROUP_KEY] = attachments[constant.GROUP_KEY]
+	dataMap[constant.TIMESTAMP_KEY] = time.Now().Format(MessageDateLayout)
+	dataMap[constant.LOCAL_ADDR], _ = attachments[constant.LOCAL_ADDR]
+	dataMap[constant.REMOTE_ADDR], _ = attachments[constant.REMOTE_ADDR]
+
+	if len(invocation.Arguments()) > 0 {
+		builder := strings.Builder{}
+		// todo(after the paramTypes were set to the invocation. we should change this implementation)
+		typeBuilder := strings.Builder{}
+		first := true
+		for _, arg := range invocation.Arguments() {
+			if first {
+				first = false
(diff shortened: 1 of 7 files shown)


filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/apache/dubbo-go/common/constant"
+	"github.com/apache/dubbo-go/common/extension"
+	"github.com/apache/dubbo-go/common/logger"
+	"github.com/apache/dubbo-go/filter"
+	"github.com/apache/dubbo-go/protocol"
+	"os"
+	"reflect"
+	"strings"
+	"time"
+)
```
AlexStocks: pls see other go files. split this import block into 3 blocks.

filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,208 @@
...
+	case ef.logChan <- accessLogData:
+		return
+	default:
+		logger.Warn("The channel is full and the access logIntoChannel data will be dropped")
+		return
+	}
+}
+
+func (ef *AccessLogFilter) buildAccessLogData(invoker protocol.Invoker, invocation protocol.Invocation) map[string]string {
+	dataMap := make(map[string]string)
```
AlexStocks: pls give the init len for this dataMap. 16 may be good.

filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,208 @@
...
+	dataMap[constant.LOCAL_ADDR], _ = attachments[constant.LOCAL_ADDR]
+	dataMap[constant.REMOTE_ADDR], _ = attachments[constant.REMOTE_ADDR]
+
+	if len(invocation.Arguments()) > 0 {
+		builder := strings.Builder{}
+		// todo(after the paramTypes were set to the invocation. we should change this implementation)
+		typeBuilder := strings.Builder{}
+		first := true
+		for _, arg := range invocation.Arguments() {
+			if first {
```
AlexStocks: this check is so ugly. u can handle the first case specially. And then use `for idx := 1; idx < len(invocation.Arguments(); idx++{`

filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (ef *AccessLogFilter) OnResponse(result protocol.Result, invoker protocol.Invoker, invocation protocol.Invocation) protocol.Result {
+	return result
+}
+
+func (ef *AccessLogFilter) writeLogToFile(data AccessLogData) {
+	accessLog := data.accessLog
+	if isDefault(accessLog) {
+		logger.Info(data.toLogMessage())
+	} else {
```
AlexStocks: u can add return in the if condition. and the delete the else keyword.

filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,208 @@
...
+		logger.Warnf("Can not get the info of access log file: %s, %v", accessLog, err)
+		return nil, err
+	}
+	last := fileInfo.ModTime().Format(FileDateFormat)
+	if now != last {
+		err = os.Rename(fileInfo.Name(), fileInfo.Name()+"."+now)
+		if err != nil {
+			logger.Warnf("Can not rename access log file: %s, %v", fileInfo.Name(), err)
+			return nil, err
+		} else {
```
AlexStocks: delete the else keyword.

filter/impl/access_log_filter_test.go
```diff
@@ -0,0 +1,66 @@
...
+package impl
+
+import (
+	"context"
+	"github.com/apache/dubbo-go/common"
+	"github.com/apache/dubbo-go/protocol"
+	"github.com/apache/dubbo-go/protocol/invocation"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"testing"
```
AlexStocks: pls see other go files. split this import block into 3 blocks.

filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/apache/dubbo-go/common/constant"
+	"github.com/apache/dubbo-go/common/extension"
+	"github.com/apache/dubbo-go/common/logger"
+	"github.com/apache/dubbo-go/filter"
+	"github.com/apache/dubbo-go/protocol"
+)
+
+const (
+	//used in URL.
+	AccessLogKey      = "accesslog"
```
hxmhlt: in dubbo accessLogKey is dubbo.accesslog.
I recommand to change the key.

filter/impl/access_log_filter_test.go
```diff
@@ -0,0 +1,83 @@
...
+ * limitations under the License.
+ */
+
+package impl
+
+import (
+	"context"
+	"testing"
+
+	"github.com/apache/dubbo-go/common/constant"
```
hxmhlt: import format

filter/impl/access_log_filter.go
flycash: It's the ACCESS_LOG_KEY, not the LOG_KEY. The difference is that the LOG_KEY is used to fetch the Logger. But the ACCESS_LOG_KEY is used to fetch the AccessLogFilter configuration from URL.
The code in Java is:
image

image

we don't have the similar mechanism to fetch different logger. So I only need to define the ACCESS_LOG_KEY.

image

image

filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,211 @@
...
+	extension.SetFilter(AccessLogKey, GetAccessLogFilter)
+}
+
+type AccessLogFilter struct {
+	logChan chan AccessLogData
+}
+
+func (ef *AccessLogFilter) Invoke(invoker protocol.Invoker, invocation protocol.Invocation) protocol.Result {
+	accessLog := invoker.GetUrl().GetParam(AccessLogKey, "")
+	logger.Warnf(invoker.GetUrl().String())
```
fangyincheng: Why warnf? `logger.Infof("[AccessLogFilter] %s", invoker.GetUrl().Key())` will is better.

fangyincheng:
How to start this filter？

filter/impl/access_log_filter.go
flycash: Sorry, it's my debug info and it has been removed

flycash:
Now I add the example in general/go-server.
It looks like that:
image

and clients need to do nothing

AlexStocks:
@fangyincheng  should  the examples be placed in [dubbogo-examples](https://github.com/dubbogo/dubbogo-samples)?

flycash:
I change the example to be comments:
image

It would be better.

fangyincheng:
> @fangyincheng should the examples be placed in [dubbogo-examples](https://github.com/dubbogo/dubbogo-samples)?

yes

fangyincheng:
> I change the example to be comments:
> image
> 
> It would be better.

Should it be default or user configuration? What about Java dubbo? And should you add some comments in `access_log_filter.go` instead of examples?

flycash:
It's a default filter in Java. But if the user does not specify the value of "accesslog" to store the access log. It won't do anything. 
So, maybe we could make it as default filter. But if the user doesn't configure the of "accesslog", we do nothing.

fangyincheng:
> It's a default filter in Java. But if the user does not specify the value of "accesslog" to store the access log. It won't do anything.
> So, maybe we could make it as default filter. But if the user doesn't configure the of "accesslog", we do nothing.

you are right. Pls do it.

flycash:
done, pls review that

filter/impl/access_log_filter.go
```diff
@@ -0,0 +1,225 @@
...
+func init() {
+	extension.SetFilter(constant.ACCESS_LOG_KEY, GetAccessLogFilter)
+}
+
+/*
+ * Although the access log filter is a default filter,
+ * you should config "accesslog" in service's config to tell the filter where store the access log.
+ * for example:
+ * "UserProvider":
+    # 可以指定多个registry，使用逗号隔开;不指定默认向所有注册中心注册
```
fangyincheng: pls use english

filter/impl/access_log_filter.go
```diff
@@ -53,10 +53,8 @@ func init() {
  * you should config "accesslog" in service's config to tell the filter where store the access log.
  * for example:
  * "UserProvider":
-    # 可以指定多个registry，使用逗号隔开;不指定默认向所有注册中心注册
```
AlexStocks: Add remark here. pls.

AlexStocks:
LGTM

AlexStocks:
@fangyincheng u recheck it and merge it, pls.

PR #221, opened 2019-09-23
Ftr: protocol_conf default values

joeyzhouy:
Ftr: issue #215


AlexStocks:
@fangyincheng pls check this pr. if there is no problems, u can merge it.

protocol/dubbo/config.go
```diff
@@ -90,6 +90,57 @@ type (
...
+			TcpNoDelay:       true,
+			TcpKeepAlive:     true,
+			KeepAlivePeriod:  "120s",
+			TcpRBufSize:      262144,
+			TcpWBufSize:      65536,
+			PkgWQSize:        512,
+			TcpReadTimeout:   "1s",
+			TcpWriteTimeout:  "5s",
+			WaitTimeout:      "1s",
+			MaxMsgLen:        2014,
```
pantianying: why use 2014?  and I think 2014 is too short.
I use 102400

protocol/dubbo/config.go
```diff
@@ -90,6 +90,57 @@ type (
...
+			TcpNoDelay:       true,
+			TcpKeepAlive:     true,
+			KeepAlivePeriod:  "120s",
+			TcpRBufSize:      262144,
+			TcpWBufSize:      65536,
+			PkgWQSize:        512,
+			TcpReadTimeout:   "1s",
+			TcpWriteTimeout:  "5s",
+			WaitTimeout:      "1s",
+			MaxMsgLen:        1024,
```
pantianying: MaxMsgLen is too small

protocol/dubbo/config.go
```diff
@@ -90,6 +90,57 @@ type (
...
+	return ClientConfig{
+		ReconnectInterval: 0,
+		ConnectionNum:     16,
+		HeartbeatPeriod:   "5s",
+		SessionTimeout:    "20s",
+		PoolSize:          64,
+		PoolTTL:           600,
+		GrPoolSize:        1200,
+		QueueLen:          64,
+		QueueNumber:       60,
```
pantianying: this size of config may cause higher CPU，this is my config I use better
image

protocol/dubbo/config.go
joeyzhouy: Configuration from example.
If the configuration is not reasonable, I can change it.

protocol/dubbo/config.go
joeyzhouy: fix

protocol/dubbo/config.go
joeyzhouy: fix

protocol/dubbo/config.go
joeyzhouy: Thank you for sharing your ideas

Issue #222, opened 2019-09-25
Ftr: when get java exception,do not triggered retry invoke


pantianying:
**What would you like to be added**:

**Why is this needed**:

PR #225, opened 2019-09-26
Ftr:issue 222

pantianying:
**What this PR does**:
https://github.com/apache/dubbo-go/issues/222

Fixes #

```release-note

```

### DIFF
--- a/cluster/cluster_impl/failover_cluster_invoker.go
+++ b/cluster/cluster_impl/failover_cluster_invoker.go
@@ -28,6 +28,7 @@ import (
 
 import (
 	"github.com/apache/dubbo-go/cluster"
+	"github.com/apache/dubbo-go/common"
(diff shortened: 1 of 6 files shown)


pantianying:
@fangyincheng pls review

hxmhlt:
I suggest that we should limit the condition  for not retry instead of limiting the condition for retry.
Because I think there should be more cases where retries are needed than where retries are not needed. And it's hard to enumerate them for us.

In dubbo, it limited that only user custom exceptions are not needed to retry. In addition, other exceptions should be caught and retried.

hxmhlt:
Maybe we need a better solution for this issue. @fangyincheng @pantianying

pantianying:
> Maybe we need a better solution for this issue. @fangyincheng @pantianying

I agreed , pls ignore this pr

PR #246, opened 2019-10-26
Feature: ExecuteLimit Support

### LINKED ISSUE #245: Ftr: ExecuteLimitFilter support
ExecuteLimitFilter: A filter to limit the concurrent requests.
**Why is this needed**:

Sometimes, we want to limit the concurrent requests on the server-side to protect our system. 

The difference between TpsLimitFilter and ExecuteLimitFilter is that the TpsLimitFilter focuses on the number of requests in the interval but the ExecuteLimitFilter will work for limiting the concurrent requests at a certain time.

flycash:
**What this PR does**:

Implement the ExecuteLimitFilter, which used to limit the concurrent requests.

Both the service-level and method-level are supports. But if the user doesn't configure the `execute.limit` in their service or the value of that is negative, the filter will do nothing.

Fixes #245 

The interface RejectedExecutionHandler is used to support this filter, so it has been removed to filter/common;

```release-note
The ExecuteLimitFilter was added to the default filters. 
```


README_CN.md
```diff
@@ -33,14 +33,14 @@ Apache License, Version 2.0
...
 - 负载均衡策略: Random/[RoundRobin](https://github.com/apache/dubbo-go/pull/66)/[LeastActive](https://github.com/apache/dubbo-go/pull/65)
-- 过滤器: Echo Health Check/[服务熔断&降级](https://github.com/apache/dubbo-go/pull/133)/[TokenFilter](https://github.com/apache/dubbo-go/pull/202)/[AccessLogFilter](https://github.com/apache/dubbo-go/pull/214)/[TpsLimitFilter](https://github.com/apache/dubbo-go/pull/237)
+- 过滤器: Echo Health Check/[服务熔断&降级](https://github.com/apache/dubbo-go/pull/133)/[TokenFilter](https://github.com/apache/dubbo-go/pull/202)/[AccessLogFilter](https://github.com/apache/dubbo-go/pull/214)/[TpsLimitFilter](https://github.com/apache/dubbo-go/pull/237)[ExecuteLimitFilter](https://github.com/apache/dubbo-go/pull/246)
 - 其他功能支持: [泛化调用](https://github.com/apache/dubbo-go/pull/122)/启动时检查/服务直连/多服务协议/多注册中心/多服务版本/服务分组
 
 开发中列表:
 
 - 集群策略: Forking
 - 负载均衡策略: ConsistentHash
-- 过滤器: CountFilter/ExecuteLimitFilter
```
AlexStocks: we do not need CountFilter now. pls delete it.

README_CN.md
```diff
@@ -33,14 +33,14 @@ Apache License, Version 2.0
...
+- 过滤器: CountFilter
```
AlexStocks: we do not need CountFilter now. pls delete it.

common/constant/default.go
```diff
@@ -46,7 +46,7 @@ const (
 const (
 	DEFAULT_KEY               = "default"
 	PREFIX_DEFAULT_KEY        = "default."
-	DEFAULT_SERVICE_FILTERS   = "echo,token,accesslog,tps"
+	DEFAULT_SERVICE_FILTERS   = "echo,token,accesslog,tps,execute"
```
fangyincheng: should `tps` and `execute` are default filter?  or should it be customed by users whether they work?  
I think it's a question that needs serious consideration.

flycash: In the implementation of those two filters, they won't work without relative configuration. 
Or if you think it would be better to removed them from default filters, I can do that and fix the examples.

fangyincheng: I am worried that if not used by most users, too many default filters will cause performance problems. Of cause, tps filter is an important filter. So you don't need to do others.

PR #262, opened 2019-11-24
Fix BUG: The SIGSYS and SIGSTOP are not supported in windows platform

flycash:
**What this PR does**:
The SIGSYS and SIGSTOP are not supported in the Windows platform, so we should define the signals we catch on different platforms.

Fixes #

```release-note

```


kimmking:
To be fixing bug should create a related issue first.

PR #275, opened 2019-12-07
feat(grpc): 支持client grpc 调用

xujianhai666:


remoting/zookeeper/listener.go
```diff
@@ -273,6 +276,7 @@ func (l *ZkEventListener) ListenServiceEvent(zkPath string, listener remoting.Da
 		children  []string
 	)
 
+	zkPath = strings.ReplaceAll(zkPath, "$", "%24")
```
xujianhai666: 这种特殊的编码 需要统一的处理

xujianhai666: 对齐  java 就好 @flycash

xujianhai666:
> 补UT。
> 我有个困惑的地方是，如果我作为一个用户，要使用grpc，那么是不是我就必须要实现DubboGrpcService这个接口？
> 能否在dubbogo-samples里面提供一些例子？

@flycash 这个后面 会自动生成，不需要用户实现

xujianhai666:
> It seems that all the stub should contain the method “GetDubboStub”. Should we define a interface?

@flycash the interface is difficult to define, the return is different. if we use reflect.Value, it is odd.

flycash:
> > It seems that all the stub should contain the method “GetDubboStub”. Should we define a interface?
> 
> @flycash the interface is difficult to define, the return is different. if we use reflect.Value, it is odd.

We don't define a complicated interface. How about defining the interface:
``` go
type DubboGrpcStub interface {
   GetDubboStub() 
}
```
It's a tag interface, which just use to tell the developers that they need to define the method "GetDubboStub" in their stub. 
Or how the developers know this limitation?

xujianhai666:
> > > It seems that all the stub should contain the method “GetDubboStub”. Should we define a interface?
> > 
> > 
> > @flycash the interface is difficult to define, the return is different. if we use reflect.Value, it is odd.
> 
> We don't define a complicated interface. How about defining the interface:
> 
> ```go
> type DubboGrpcStub interface {
>    GetDubboStub() 
> }
> ```
> 
> It's a tag interface, which just use to tell the developers that they need to define the method "GetDubboStub" in their stub.
> Or how the developers know this limitation?

@flycash it should be generated by protoc, not  implemented by user

go.mod
```diff
@@ -5,6 +5,7 @@ require (
 	github.com/afex/hystrix-go v0.0.0-20180502004556-fa1af6a1f4f5
 	github.com/aliyun/alibaba-cloud-sdk-go v0.0.0-20190802083043-4cd0c391755e // indirect
 	github.com/apache/dubbo-go-hessian2 v1.2.5-0.20191029001541-894e45c9aaaa
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869
```
AlexStocks: I do not think this so old package is qualified which is not been maintained. maybe u should use https://github.com/stretchr/testify/assert.

go.sum
```diff
@@ -449,6 +449,7 @@ github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81P
 github.com/tebeka/strftime v0.1.3 h1:5HQXOqWKYRFfNyBMNVc9z5+QzuBtIXy03psIhtdJYto=
 github.com/tebeka/strftime v0.1.3/go.mod h1:7wJm3dZlpr4l/oVK0t1HYIc4rMzQ2XJlOMIUJUJH6XQ=
 github.com/tent/http-link-go v0.0.0-20130702225549-ac974c61c2f9/go.mod h1:RHkNRtSLfOK7qBTHaeSX1D6BNpI3qw7NTxsmNr4RvN8=
+github.com/tevid/gohamcrest v1.1.1 h1:ou+xSqlIw1xfGTg1uq1nif/htZ2S3EzRqLm2BP+tYU0=
```
AlexStocks: u used so much unqualified packages.

xujianhai666: it is from agollo：
-> % go mod why github.com/tevid/gohamcrest
# github.com/tevid/gohamcrest
github.com/apache/dubbo-go/config_center/apollo
github.com/zouyx/agollo
github.com/zouyx/agollo.test
github.com/tevid/gohamcrest

go.mod
xujianhai666: it have been deleted

protocol/grpc/grpc_protocol.go
```diff
@@ -0,0 +1,104 @@
...
+func (gp *GrpcProtocol) openServer(url common.URL) {
+	_, ok := gp.serverMap[url.Location]
+	if !ok {
+		_, ok := gp.ExporterMap().Load(url.ServiceKey())
+		if !ok {
+			panic("[GrpcProtocol]" + url.Key() + "is not existing")
+		}
+
+		gp.serverLock.Lock()
+		_, ok = gp.serverMap[url.Location]
```
flycash: How about:
```go
gp.serverLock.Lock()
		_, ok = gp.serverMap[url.Location]
		if !ok {
			srv := NewServer()
			gp.serverMap[url.Location] = srv
			gp.serverLock.Unlock()
			srv.Start(url)
			return
		}
		gp.serverLock.Unlock()
```
I think we can release the lock before invoking srv.Start(url)

flycash:
LGTM

PR #278, opened 2019-12-10
ftr: Metric support Part1: metric filter, method count, method rt

flycash:
**What this PR does**:
I try to build the metric mechanism according to the java-dubbo's metric mechanism. But the metric module is complicated.
So I will implement the metric module in steps:
1. Build the metric filter to collect the method count and method rt data: in this step, the major interfaces like MetricManager, MetricRegistry, Metric are defined and the default implementations are provided;
2. Using `Prometheus ` to collect the data of method count and method rt;
3. Implements more data, like Counter, Meter, Gauge, Timer, Histogram... And use the Prometheus to collect them;
4. Try to support more instrumentation, like Zipkin, open tracing... 
5. After we finish step 3, the BBR rate limit also could be implemented.

Now this PR finished step1.

Fixes #

```release-note

```


flycash:
steps in Chinese:
计划：

依据java dubbo已有的metric机制，实现对方法调用次数和响应时间的统计。在这个过程中，会初步搭建出来一个类似dubbo metric的东西（好吧，就是抄了一份）；
@刘晓东 可以接入Prometheus，此时只能接入方法调用次数和响应时间；
持续实现剩余的metric，如Counter, Meter等，直至完全实现完；
尝试接入zipkin和open tracing等框架。实际上，这个部分和metric是相关但是又不完全一致的；
在3完成之后，就可以尝试BBR之类的rate limit等东西了。实际上，在3完成之后，所有依赖于metric特征的都可以提上议程；

metrics/impl/default_metric_manager.go
```diff
@@ -0,0 +1,76 @@
...
+	extension.SetMetricManager(constant.DEFAULT_KEY, newDefaultMetricManager())
+}
+
+/*
+ * this implementation dependent on  the MetricRegistry.
+ * com.alibaba.metrics.AliMetricManager
+ */
+type DefaultMetricManager struct {
+	// group name -> MetricRegistry
+	metricRegistryMap sync.Map
```
imxyb: I think it is better to maintain with normal map and lock

zouyx: actually i agree to use sync.Map,   this scene is fit to sync.Map use case,i think metricRegistryMap will be a lot of data

>>The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

Ref: [sync.Map godoc](https://golang.org/pkg/sync/#Map)

imxyb: > actually i agree to use sync.Map, this scene is fit to sync.Map use case,i think metricRegistryMap will be a lot of data
> 
> > > The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.
> 
> Ref: [sync.Map godoc](https://golang.org/pkg/sync/#Map)

get it, thx

metrics/impl/bucket_counter_impl.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (bci *BucketCounterImpl) GetBucketInterval() time.Duration {
+	return bci.interval
+}
+
+// it will align the timestmp to N*interval and then convert the timestmp to seconds
+// for example: if the timestmp is 11000ms, and the interval is 5s, the result will be 11000/1000 /5 *5 = 10
+func (bci *BucketCounterImpl) alignTimestamp(timestmp int64) int64 {
+	return timestmp / int64(1e3) / int64(bci.interval.Seconds()) * int64(bci.interval.Seconds())
```
zouyx: if change to second ,it will calculate one more time, right?but 1e3 is not good for readable

metrics/impl/default_registry_impl.go
```diff
@@ -0,0 +1,93 @@
...
+ * limitations under the License.
+ */
+
+package impl
+
+import (
+	"sync"
+	"sync/atomic"
+
+	"github.com/apache/dubbo-go/config"
```
zouyx: you know it

metrics/impl/manual_clock.go
```diff
@@ -0,0 +1,42 @@
...
+	tickInNanos int64
+	mutex       sync.Mutex
+}
+
+func (mc *ManualClock) GetTick() int64 {
+	return mc.tickInNanos
+}
+
+func (mc *ManualClock) GetTime() int64 {
+	return mc.tickInNanos / 1e6
```
zouyx: should it duration.second?

metrics/metric_name.go
```diff
@@ -0,0 +1,69 @@
...
+ */
+func (mn *MetricName) HashKey() string {
+	mn.hashKeyOnce.Do(func() {
+		mn.hashKey = mn.Key
+		if len(mn.Tags) <= 0 {
+			return
+		}
+		mn.hashKey += "-"
+		for key, value := range mn.Tags {
+			mn.hashKey += key + ":" + value
```
zouyx: are ":" and "-" should constant?

common/extension/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+
+// the manager should be a pointer of MetricManager.
+func SetMetricManager(name string, manager metrics.MetricManager) {
+	metricManagerMap[name] = manager
+}
+
+func GetMetricManager(name string) metrics.MetricManager {
+	manager, found := metricManagerMap[name]
+	if !found {
+		panic("Can not find the metric manager with name: " + name +
```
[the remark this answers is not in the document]
flycash: No, I think panic would be better. If someone wants to get the MetricManger but we can't find out, it means that he must forget to register the MetricManager or the name is wrong. No matter the cause is, it's a programming problem and is not a tolerable problem. Even we don't know how to handle the returned error.

metrics/clock.go
```diff
@@ -0,0 +1,52 @@
...
+	Clock
+	startNanoTime int64
+}
+
+func (uc *UserTimeClock) GetTick() int64 {
+	return time.Now().UnixNano() - uc.startNanoTime
+}
+
+func (uc *UserTimeClock) GetTime() int64 {
+	return time.Now().UnixNano() / 1e6
```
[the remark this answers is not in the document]
flycash: It should be time.Millisecond.

metrics/impl/bucket_counter_impl.go
```diff
@@ -0,0 +1,163 @@
...
+	return bci.GetBucketCountsSince(0)
+}
+
+func (bci *BucketCounterImpl) GetBucketCountsSince(startTime int64) map[int64]int64 {
+
+	currentTs := bci.alignTimestamp(bci.clock.GetTime())
+	bucketList := bci.buckets.getBucketList()
+	result := make(map[int64]int64, len(bucketList))
+	for _, value := range bucketList {
+		bkTimestampInMs := 1e3 * value.timestamp
```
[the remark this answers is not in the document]
flycash: the unit of timestamp is Second...

metrics/impl/bucket_counter_impl.go
flycash: The comments had pointed out that the `timestmp` is MS but we need the SECOND, so we should dive it.

flycash:
Be careful about the time unit. 1ms = 10^6ns.
And the timestamp of bucket struct is SECOND, not Millisecond. Because we do not have any bucket which interval is < 1s.
If the interval is less than 1s, the cost of collecting data will be too huge to accept.

flycash:
But if someone wants to use the `bucket` struct, he can define the `timestamp's` unit as Millisecond. But when I use the `bucket` struct in my implementation, the time unit of the `timestamp` is Second.
It's the same with Dubbo-Java's implementation.

zouyx:
i think ```feature/metrics``` should rebase or merge master or develop.

PR #283, opened 2019-12-15
feat(registry/zk):url encode zkpath

xujianhai666:
**What this PR does**:
be consistent with java impl, support url encoder for zkpath

### DIFF
--- a/registry/zookeeper/registry.go
+++ b/registry/zookeeper/registry.go
@@ -261,6 +261,10 @@ func (r *zkRegistry) Register(conf common.URL) error {
 	return nil
 }
 
+func (r *zkRegistry) service(c common.URL) string {
+	return url.QueryEscape(c.Service())
+}
+
 func (r *zkRegistry) register(c common.URL) error {
 	var (
 		err error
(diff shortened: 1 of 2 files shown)


xujianhai666:
> I had checked codes using the zkPath. It seems that we don't need to decode the zkPath when we get the zkPath. Right?

yes, just for zk registry

xujianhai666:
> I had checked codes using the zkPath. It seems that we don't need to decode the zkPath when we get the zkPath. Right?

@flycash yes, just for zk registry

zouyx:
I had checked codes using the zkPath too. we use zkPath to registry and get children node,so it seems that no need to decode.

hxmhlt:
I checked java version. There is method named 'encoded' in URL.java and it called by ZookeeperRegistry for zkPath encode. Though I have not received any issue about this problem...Maybe it is encoded by zk golang client.

xujianhai666:
> I had checked codes using the zkPath too. we use zkPath to registry and get children node,so it seems that no need to decode.

@zouyx for consistent with java impl. Besides, grpc need this decode method to resolve "$" for inner class

PR #286, opened 2019-12-18
change the position of the lock

pantianying:
**What this PR does**:

Fixes #
image

```release-note

```

### DIFF
--- a/registry/protocol/protocol.go
+++ b/registry/protocol/protocol.go
@@ -338,10 +338,10 @@ func setProviderUrl(regURL *common.URL, providerURL *common.URL) {
 }
 
 func GetProtocol() protocol.Protocol {
-	if regProtocol == nil {
-		regProtocol = newRegistryProtocol()
+	if regProtocol != nil {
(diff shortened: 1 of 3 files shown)


registry/protocol/protocol.go
```diff
@@ -338,10 +338,10 @@ func setProviderUrl(regURL *common.URL, providerURL *common.URL) {
 }
 
 func GetProtocol() protocol.Protocol {
```
flycash: This method I had reverted to its origin version. You can merge master branch and there is not necessary to revert it again.

pantianying: Yes, I know this, but I don't know how to merge your code into the development branch. This problem affects my test, so I added your recovered code to my branch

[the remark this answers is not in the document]
pantianying: After the program runs, there are only read operations and no write operations. So I don't think there's a concurrency problem

zouyx:
LGTM

PR #288, opened 2019-12-18
fix(core): add lock for attachment

xujianhai666:


imxyb:
why not use sync.Map?

zouyx:
i have the same question.

protocol/invocation/rpcinvocation.go
```diff
@@ -91,6 +93,8 @@ func (r *RPCInvocation) AttachmentsByKey(key string, defaultValue string) string
 }
 
 func (r *RPCInvocation) SetAttachments(key string, value string) {
+	r.lock.Lock()
```
AlexStocks: it seems that you just use the lock for writing. if u reading the attachment, u should use lock too.

xujianhai666: @AlexStocks okay.

protocol/invocation/rpcinvocation.go
```diff
@@ -37,6 +38,7 @@ type RPCInvocation struct {
 	callBack       interface{}
 	attachments    map[string]string
 	invoker        protocol.Invoker
+	lock           sync.Mutex
```
fangyincheng: should it be a rwmutex?

hxmhlt: agree

xujianhai666: yes, done @fangyincheng

xujianhai666:
> why not use sync.Map?

@imxyb @zouyx sync.map is too terrible for humable readable. and for official comment, if u can use map to deal with it, do not use sync.map

protocol/invocation/rpcinvocation.go
```diff
@@ -80,6 +82,8 @@ func (r *RPCInvocation) Attachments() map[string]string {
 }
 
 func (r *RPCInvocation) AttachmentsByKey(key string, defaultValue string) string {
+	r.lock.Lock()
+	defer r.lock.Unlock()
```
fangyincheng: func (rw *RWMutex) RLock()

func (rw *RWMutex) RUnlock()

xujianhai666: @fangyincheng done

PR #289, opened 2019-12-20
[WIP] kubernetes as registry

sxllwx:
**What this PR does**:
Ftr: kubernetes for registry

Fixes #

temporary lack of unit testing

```release-note

```

### DIFF
--- a/registry/etcdv3/listener_test.go
+++ b/registry/etcdv3/listener_test.go
@@ -19,7 +19,6 @@ package etcdv3
 
 import (
 	"context"
-	"github.com/apache/dubbo-go/config_center"
 	"testing"
 	"time"
 )
@@ -32,6 +31,7 @@ import (
 
 import (
 	"github.com/apache/dubbo-go/common"
+	"github.com/apache/dubbo-go/config_center"
 	"github.com/apache/dubbo-go/remoting"
 )
 
@@ -46,7 +46,8 @@ func (suite *RegistryTestSuite) SetupSuite() {
 	t := suite.T()
 
 	cfg := embed.NewConfig()
-	cfg.Dir = "/tmp/default.etcd"
+	// avoid conflict with default etcd work-dir
+	cfg.Dir = "/tmp/default-dubbo-go.etcd"
 	e, err := embed.StartEtcd(cfg)
 	if err != nil {
 		t.Fatal(err)
--- a/registry/etcdv3/listener_test.go
+++ b/registry/etcdv3/listener_test.go
@@ -41,6 +41,8 @@ type RegistryTestSuite struct {
 	etcd *embed.Etcd
 }
 
+const defaultEtcdV3WorkDir = "/tmp/default-dubbo-go-registry.etcd"
+
 // start etcd server
 func (suite *RegistryTestSuite) SetupSuite() {
 
@@ -48,7 +50,7 @@ func (suite *RegistryTestSuite) SetupSuite() {
 
 	cfg := embed.NewConfig()
 	// avoid conflict with default etcd work-dir
-	cfg.Dir = "/tmp/default-dubbo-go-registry.etcd"
+	cfg.Dir = defaultEtcdV3WorkDir
 	e, err := embed.StartEtcd(cfg)
 	if err != nil {
 		t.Fatal(err)
@@ -69,7 +71,7 @@ func (suite *RegistryTestSuite) SetupSuite() {
 func (suite *RegistryTestSuite) TearDownSuite() {
 	suite.etcd.Close()
 	// clean the etcd workdir
-	if err := exec.Command("rm", "-rf", "/tmp/default-dubbo-go-registry.etcd").Run(); err != nil {
+	if err := exec.Command("rm", "-rf", defaultEtcdV3WorkDir).Run(); err != nil {
 		suite.FailNow(err.Error())
 	}
 }
--- a/registry/etcdv3/listener_test.go
+++ b/registry/etcdv3/listener_test.go
@@ -19,6 +19,7 @@ package etcdv3
 
 import (
 	"context"
+	"os/exec"
 	"testing"
 	"time"
 )
@@ -67,6 +68,10 @@ func (suite *RegistryTestSuite) SetupSuite() {
 // stop etcd server
 func (suite *RegistryTestSuite) TearDownSuite() {
 	suite.etcd.Close()
+	// clean the etcd workdir
+	if err := exec.Command("rm", "-rf","/tmp/default-dubbo-go.etcd").Run(); err != nil{
+		suite.FailNow(err.Error())
+	}
 }
 
 func (suite *RegistryTestSuite) TestDataChange() {
--- a/registry/etcdv3/listener_test.go
+++ b/registry/etcdv3/listener_test.go
@@ -48,7 +48,7 @@ func (suite *RegistryTestSuite) SetupSuite() {
 
 	cfg := embed.NewConfig()
 	// avoid conflict with default etcd work-dir
-	cfg.Dir = "/tmp/default-dubbo-go.etcd"
+	cfg.Dir = "/tmp/default-dubbo-go-registry.etcd"
 	e, err := embed.StartEtcd(cfg)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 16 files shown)


remoting/kubernetes/client.go
```diff
@@ -0,0 +1,638 @@
...
+}
+
+// initCurrentPod
+// 1. get current pod
+// 2. give the dubbo-label for this pod
+func (c *Client) initCurrentPod() (*v1.Pod, error) {
+
+	// read the current pod status
+	currentPod, err := c.rawClient.CoreV1().Pods(c.ns).Get(c.currentPodName, metav1.GetOptions{})
+	if err != nil {
```
hxmhlt: Is there any solution to avoid watch all pods events in namespace?

registry/kubernetes/registry.go
```diff
@@ -0,0 +1,351 @@
...
+		services = append(services, confIf)
+	}
+
+	flag := true
+	for _, confIf := range services {
+		err := r.Register(confIf)
+		if err != nil {
+			logger.Errorf("(kubernetesProviderRegistry)register(conf{%#v}) = error{%#v}",
+				confIf, perrors.WithStack(err))
+			flag = false
```
flycash: It seems that you can `return false` directly and `return true` at the end of this method. So the variable `flag` could be removed

registry/kubernetes/registry.go
```diff
@@ -0,0 +1,351 @@
...
+		return true
+	})
+	params.Add("pid", processID)
+	params.Add("ip", localIP)
+	params.Add("anyhost", "true")
+	params.Add("category", (common.RoleType(common.PROVIDER)).String())
+	params.Add("dubbo", "dubbo-provider-golang-"+constant.Version)
+	params.Add("side", (common.RoleType(common.PROVIDER)).Role())
+
+	if len(svc.Methods) == 0 {
```
flycash: I found that you had return error when you find out that `len(svc.Methods) ==0 ` at the begining of this method. So why do we need to check `len(svc.Methods) == 0` again? Do I miss something?

registry/kubernetes/registry.go
```diff
@@ -0,0 +1,351 @@
...
+		client := r.client
+		r.cltLock.Unlock()
+		if client == nil {
+			return nil, perrors.New("kubernetes client broken")
+		}
+
+		// new client & listener
+		listener := kubernetes.NewEventListener(r.client)
+
+		r.listenerLock.Lock()
```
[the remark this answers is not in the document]
sxllwx: This is indeed a bug, and the same problem exists in etcdv3, the latest commit has been fixed。

registry/kubernetes/registry.go
sxllwx: Yes, it would be more straightforward to return the bool value directly。

registry/kubernetes/registry.go
sxllwx: This is indeed a bug, and the same problem exists in etcdv3, the latest commit has been fixed。

registry/kubernetes/registry.go
```diff
@@ -0,0 +1,367 @@
...
+
+func (r *kubernetesRegistry) registerConsumer(svc common.URL) error {
+
+	consumersNode := fmt.Sprintf("/dubbo/%s/%s", svc.Service(), common.DubboNodes[common.CONSUMER])
+	if err := r.createDirIfNotExist(consumersNode); err != nil {
+		logger.Errorf("kubernetes client create path %s: %v", consumersNode, err)
+		return perrors.WithMessage(err, "kubernetes create consumer nodes")
+	}
+	providersNode := fmt.Sprintf("/dubbo/%s/%s", svc.Service(), common.DubboNodes[common.PROVIDER])
+	if err := r.createDirIfNotExist(providersNode); err != nil {
```
flycash: Should we create the provider's node here?

remoting/kubernetes/facade.go
```diff
@@ -0,0 +1,94 @@
...
+				err = ValidateClient(r)
+				logger.Infof("Kubernetes ProviderRegistry.validateKubernetesClient = error{%#v}", perrors.WithStack(err))
+
+				if err == nil {
+					if r.RestartCallBack() {
+						break
+					}
+				}
+				failTimes++
+				if MaxFailTimes <= failTimes {
```
flycash: Why do we reset the failTimes to MaxFailTImes. Do we use the MaxFailTimes to avoid the client restarts too many times? But I don't see any validation about it. Do I miss something?

registry/kubernetes/listener.go
```diff
@@ -0,0 +1,117 @@
...
+				default:
+				}
+				continue
+			}
+			return &registry.ServiceEvent{Action: e.ConfigType, Service: e.Value.(common.URL)}, nil
+		}
+	}
+}
+func (l *configurationListener) Close() {
+	l.registry.wg.Done()
```
flycash: Please read the #222 to check whether this implementation has the same problem.

hxmhlt:
LGTM

remoting/kubernetes/client.go
```diff
@@ -0,0 +1,655 @@
...
+
+	c.wg.Add(1)
+
+	// try once
+	watcher, err := c.rawClient.CoreV1().Pods(c.ns).Watch(metav1.ListOptions{
+		LabelSelector: fields.OneTermEqualSelector(DubboIOLabelKey, DubboIOLabelValue).String(),
+		Watch:         true,
+	})
+	if err != nil {
+		return perrors.WithMessagef(err, "try to watch the namespace (%s) pods", c.ns)
```
flycash: If you return here, c.wg.Add(1) but you do not start a goroutine. Is this a problem?

registry/etcdv3/listener_test.go
```diff
@@ -66,6 +68,10 @@ func (suite *RegistryTestSuite) SetupSuite() {
 // stop etcd server
 func (suite *RegistryTestSuite) TearDownSuite() {
 	suite.etcd.Close()
+	// clean the etcd workdir
+	if err := exec.Command("rm", "-rf", "/tmp/default-dubbo-go-registry.etcd").Run(); err != nil {
```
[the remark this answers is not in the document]
sxllwx: Yes,I had fix in latest commit.

registry/kubernetes/registry.go
sxllwx: Yes,I had fix in latest commit.

sxllwx: Yes, there is no need to create a Provider directory here, and it is not required in etcdv3 && kubernetes。

remoting/kubernetes/facade.go
sxllwx: You have found the answer yourself～

registry/kubernetes/listener.go
sxllwx: Yes，I had fix the issue in latest commit

remoting/kubernetes/client.go
sxllwx: Yes, this is a bad logic. I had fix this in latest commit.

registry/kubernetes/registry.go
```diff
@@ -0,0 +1,369 @@
...
+
+	if r.configListener != nil {
+		r.configListener.Close()
+	}
+	r.stop()
+}
+
+func (r *kubernetesRegistry) stop() {
+
+	close(r.done)
```
AlexStocks: pls use a once.do to defeat using error. maybe it would be invoked more than once.

remoting/kubernetes/listener.go
```diff
@@ -0,0 +1,205 @@
...
+				// if event is delete
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+// return true mean the event type is DELETE
```
AlexStocks: this return value is not in its right way. u should return the create/update/delete event itself.

registry/kubernetes/registry.go
sxllwx: the latest commit has been fixed。

remoting/kubernetes/listener.go
sxllwx: It's just like zookeeper,  detail https://github.com/apache/dubbo-go/blob/8d344956260b2359637381cfab35b9470601cf1a/remoting/zookeeper/listener.go#L56。

PR #316, opened 2020-01-11
add nacos configcenter

amudong:
add nacos configcenter


AlexStocks:
@zouyx pls own this pr and check it.

zouyx:
@amudong code coverage will decrease , could you add more UT？

config_center/nacos/impl.go
```diff
@@ -0,0 +1,160 @@
...
+	client       *NacosClient
+	keyListeners sync.Map
+	parser       parser.ConfigurationParser
+}
+
+func newNacosDynamicConfiguration(url *common.URL) (*nacosDynamicConfiguration, error) {
+	c := &nacosDynamicConfiguration{
+		rootPath:     "/" + url.GetParam(constant.CONFIG_NAMESPACE_KEY, config_center.DEFAULT_GROUP) + "/config",
+		url:          url,
+		keyListeners: sync.Map{},
```
AlexStocks: u need not to init it in this way. delete this line and you can use it too.

config_center/nacos/client.go
```diff
@@ -0,0 +1,247 @@
...
+func (n *NacosClient) Done() <-chan struct{} {
+	return n.exit
+}
+
+func (n *NacosClient) stop() bool {
+	select {
+	case <-n.exit:
+		return true
+	default:
+		close(n.exit)
```
AlexStocks: u need to add a once.do here. because it is not thread safe.

config_center/nacos/client.go
```diff
@@ -0,0 +1,247 @@
...
+func (n *NacosClient) NacosClientValid() bool {
+	select {
+	case <-n.exit:
+		return false
+	default:
+	}
+
+	valid := true
+	n.Lock()
+	if n.Client == nil {
```
AlexStocks: pls delete the if cond. just one "n.Client = nil" is enough.

config_center/nacos/client.go
```diff
@@ -0,0 +1,247 @@
...
+	for _, opt := range opts {
+		opt(opions)
+	}
+
+	err = nil
+
+	lock := container.NacosClientLock()
+	url := container.GetUrl()
+
+	lock.Lock()
```
AlexStocks: the lock scope is so large. I think u should try your best to reduce its scope.

config_center/nacos/impl.go
AlexStocks: because keyListeners's default value is sync.Map{}

config_center/nacos/impl.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (n *nacosDynamicConfiguration) AddListener(key string, listener config_center.ConfigurationListener, opions ...config_center.Option) {
+	n.addListener(key, listener)
+}
+
+func (n *nacosDynamicConfiguration) RemoveListener(key string, listener config_center.ConfigurationListener, opions ...config_center.Option) {
+	n.removeListener(key, listener)
+}
+
+// 在nacos group是dubbo DataId是key configfile 或  appconfigfile
```
AlexStocks: u should not use Chinese remark.

zouyx:
@amudong i think you may add test case about ```config_center/nacos/client.go``` and ```config_center/nacos/listener.go```  Coverage will increase

config_center/nacos/client.go
```diff
@@ -0,0 +1,247 @@
...
+		svrConfList = append(svrConfList, svrconf)
+	}
+	client, err := clients.CreateConfigClient(map[string]interface{}{
+		"serverConfigs": svrConfList,
+		"clientConfig": nacosconst.ClientConfig{
+			TimeoutMs:           uint64(timeout.Nanoseconds() / 1e6),
+			ListenInterval:      20000,
+			NotLoadCacheAtStart: true,
+			LogDir:              "logs/nacos/log", //TODO unified log directory
+		},
```
hxmhlt: Can nacos log provided by dubbogo log?

config_center/nacos/facade.go
```diff
@@ -0,0 +1,96 @@
...
+
+			// Connect nacos until success.
+			failTimes = 0
+			for {
+				select {
+				case <-r.GetDone():
+					logger.Warnf("(NacosProviderRegistry)reconnectZkRegistry goroutine exit now...")
+					break LOOP
+				case <-getty.GetTimeWheel().After(timeSecondDuration(failTimes * ConnDelay)): // Prevent crazy reconnection nacos.
+				}
```
hxmhlt: use dubbogo/gost time wheel instead of time wheel in getty

config_center/nacos/client.go
zouyx: i have checked the source code of nacos sdk，no interface，no method。。

config_center/nacos/listener.go
```diff
@@ -0,0 +1,66 @@
...
+func callback(ctx context.Context, listener config_center.ConfigurationListener, namespace, group, dataId, data string) {
+	listener.Process(&config_center.ConfigChangeEvent{Key: dataId, Value: data, ConfigType: remoting.EventTypeUpdate})
+}
+
+func (l *nacosDynamicConfiguration) addListener(key string, listener config_center.ConfigurationListener) {
+	_, loaded := l.keyListeners.Load(key)
+	if !loaded {
+		_, cancel := context.WithCancel(context.Background())
+		(*l.client.Client).ListenConfig(vo.ConfigParam{ //TODO 这个listen接口应该要有个context的
+			//(*l.client.Client).ListenConfigWithContext(ctx, vo.ConfigParam{
```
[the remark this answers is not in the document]
amudong: ok.i delete it

config_center/nacos/client.go
```diff
@@ -0,0 +1,247 @@
...
+				IpAddr: split[0],
+				Port:   port,
+			}
+			svrConfList = append(svrConfList, svrconf)
+		}
+
+		client, err := clients.CreateConfigClient(map[string]interface{}{
+			"serverConfigs": svrConfList,
+			"clientConfig": nacosconst.ClientConfig{
+				TimeoutMs:           uint64(container.NacosClient().Timeout.Nanoseconds() / 1e6),
```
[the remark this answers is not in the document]
amudong: use time.Milliseconds() instead.

config_center/nacos/client.go
amudong: At first, i also want to code like zk that add 'remoting'. Gradually, I find that nacos_sdk
's listening model does not support this writing. I think we need to refactor the interface of nacos_sdk.

hxmhlt: emm...ok.

PR #318, opened 2020-01-12
Imp: reduce protocol/dubbo/pool.go lock scope

AlexStocks:
**What this PR does**:

reduce protocol/dubbo/pool.go lock scope

### DIFF
--- a/protocol/dubbo/pool.go
+++ b/protocol/dubbo/pool.go
@@ -154,11 +154,11 @@ func (c *gettyRPCClient) addSession(session getty.Session) {
 	}
 
 	c.lock.Lock()
+	defer c.lock.Unlock()
 	if c.sessions == nil {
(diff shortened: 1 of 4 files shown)


protocol/dubbo/pool.go
```diff
@@ -238,28 +251,42 @@ func (c *gettyRPCClient) isAvailable() bool {
...
-		c.sessions = c.sessions[:0]
+		var (
+			gettyClient getty.Client
+			sessions    []*rpcSession
+		)
+		func() {
+			c.lock.Lock()
+			defer c.lock.Unlock()
+
+			gettyClient = c.gettyClient
```
pantianying: What's the use of this 'gettyclient'

AlexStocks: close it out of lock

PR #322, opened 2020-01-14
[WIP] dubbogo compass metrics Prometheus support

flixgithub:
**What this PR does**:
Prometheus for dubbogo first step, more metrics collect please stay tuned
1. Compass metrics collect summary implement

Fixes #

```release-note

```

### DIFF
--- a/metrics/histogram.go
+++ b/metrics/histogram.go
@@ -26,14 +26,12 @@ type Histogram interface {
 	 *
 	 * @param value the length of the value
 	 */
-	 Update(value int)
+	Update(value int)
 
 	/**
 	 * Adds a recorded value.
 	 *
 	 * @param value the length of the value
 	 */
-	 UpdateN(value int64)
-
-
+	UpdateN(value int64)
 }
--- a/metrics/histogram.go
+++ b/metrics/histogram.go
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package metrics
+
+type Histogram interface {
+	Metric
+	Sampling
+	Counting
+	/**
+	 * Adds a recorded value.
+	 *
+	 * @param value the length of the value
+	 */
+	 Update(value int)
+
+	/**
+	 * Adds a recorded value.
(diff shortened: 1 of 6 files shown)


monitor/dubbogo_exporter.go
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monitor
+
+import (
```
AlexStocks: split

monitor/collector/collector.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (e *Exporter) Describe(ch chan<- *prometheus.Desc) {
+	for _, m := range e.metrics {
+		ch <- m
+	}
+}
+
+func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
+	e.mutex.Lock()
+	defer e.mutex.Unlock()
```
AlexStocks: the lock scope is so large, sir. in my opinion, u just need to use the read lock in the for-loop.

metrics/histogram.go
```diff
@@ -0,0 +1,37 @@
...
+type Histogram interface {
+	Metric
+	Sampling
+	Counting
+	/**
+	 * Adds a recorded value.
+	 *
+	 * @param value the length of the value
+	 */
+	Update(value int)
```
flycash: What's the difference between Update and UpdateN. I think that Update = UpdateN(1). Right?

flycash: What's the difference between Update and UpdateN. I think that Update = UpdateN(1). Right?

AlexStocks: @flycash  Agree with u. just one func is enough.

monitor/collector/collector.go
```diff
@@ -0,0 +1,83 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package collector
+
+import (
+	"github.com/apache/dubbo-go/metrics/impl"
+	"github.com/prometheus/client_golang/prometheus"
+	"sync"
```
zouyx: split?

metrics/impl/histogram_impl.go
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package impl
+
+import (
+	"github.com/apache/dubbo-go/metrics"
+	"time"
+)
+
+type HistorgramImpl struct {
```
zouyx: i am not agree to use HistorgramImpl ,it seems not very intuitive
my suggest:```PrometheusHistorgram``` or ```HistorgramManager```

monitor/dubbogo_exporter.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"net/http"
+)
+
+const (
+	namespace = "dubbogo" // For Prometheus metrics.
+)
+
+var (
+	listenAddr       = flag.String("web.listen-port", "9001", "An port to listen on for web interface and telemetry.")
```
hxmhlt: It is necessary to open the way to configure prometheus port and namespace.

fangyincheng:
don't use `impl` directory. refer to `filter` directory.

AlexStocks:
@flixgithub pls review those comments and improve them.

PR #326, opened 2020-01-16
Fix:zookeeper config center namespace default value invalid

zouyx:
**What this PR does**:

zookeeper config center can't use when added  item key which is configcenter.namepace in config file

reason:

zookeeper use namespace key but have not  add to config file

ref:
https://github.com/apache/dubbo-go/blob/cabd56a95cd775ebd976c2b29c5ff95260986d61/config_center/zookeeper/impl.go#L59

Fixes #

```release-note

```


AlexStocks:
LGTM

PR #328, opened 2020-01-19
supplement rpc stats

CodingSinger:


filter/filter_impl/active_filter.go
```diff
@@ -17,14 +17,22 @@
...
 	"github.com/apache/dubbo-go/common/logger"
 	"github.com/apache/dubbo-go/filter"
 	"github.com/apache/dubbo-go/protocol"
+	invocation2 "github.com/apache/dubbo-go/protocol/invocation"
 )
 
-const active = "active"
+const (
+	active                  = "active"
+	dubbo_invoke_start_time = "dubbo_invoke_start_time"
```
zouyx: i think you should use camel case for the var

filter/filter_impl/active_filter.go
```diff
@@ -36,13 +44,19 @@ type ActiveFilter struct {
...
 	protocol.BeginCount(invoker.GetUrl(), invocation.MethodName())
 	return invoker.Invoke(invocation)
 }
 
 func (ef *ActiveFilter) OnResponse(result protocol.Result, invoker protocol.Invoker, invocation protocol.Invocation) protocol.Result {
 
-	protocol.EndCount(invoker.GetUrl(), invocation.MethodName())
+	startTime, err := strconv.ParseInt(invocation.(*invocation2.RPCInvocation).AttachmentsByKey(dubbo_invoke_start_time, "0"), 10, 64)
+	if err != nil {
+		panic("parse dubbo_invoke_start_time to int64 failed")
```
zouyx: i dont think panic in filter is a good choice

protocol/rpc_status.go
```diff
@@ -28,17 +29,68 @@ import (
...
+	return atomic.LoadInt64(&rpc.lastRequestFailedTimestamp)
+}
+
+func (rpc *RpcStatus) GetSuccessiveRequestFailureCount() int32 {
+	return atomic.LoadInt32(&rpc.successiveRequestFailureCount)
+}
+
+func GetUrlStatus(url common.URL) *RpcStatus {
+	rpcStatus, _ := serviceStatistic.LoadOrStore(url.Key(), &RpcStatus{})
+	return rpcStatus.(*RpcStatus)
```
zouyx: should you add if rpcStatus==nil before return？

CodingSinger: will never return empty here.

filter/filter_impl/active_filter.go
CodingSinger: ok, I will return a RpcResult to wrap this err.

filter/filter_impl/active_filter.go
CodingSinger: good idea.

protocol/rpc_status.go
zouyx: ok，mobile phone screen too small to see the loadorstore method

protocol/rpc_status.go
```diff
@@ -58,18 +110,44 @@ func GetStatus(url common.URL, methodName string) *RpcStatus {
 }
 
 func BeginCount(url common.URL, methodName string) {
-	beginCount0(GetStatus(url, methodName))
+	beginCount0(GetUrlStatus(url))
```
zouyx: use GetURLStatus not Url

protocol/rpc_status.go
```diff
@@ -58,18 +110,44 @@ func GetStatus(url common.URL, methodName string) *RpcStatus {
...
+	beginCount0(GetMethodStatus(url, methodName))
 }
 
-func EndCount(url common.URL, methodName string) {
-	endCount0(GetStatus(url, methodName))
+func EndCount(url common.URL, methodName string, elapsed int64, succeeded bool) {
+	endCount0(GetUrlStatus(url), elapsed, succeeded)
```
zouyx: use GetURLStatus not Url

protocol/rpc_status.go
```diff
@@ -28,17 +29,68 @@ import (
 
 var (
 	methodStatistics sync.Map // url -> { methodName : RpcStatus}
+	serviceStatistic sync.Map // url -> RpcStatus
 )
 
 type RpcStatus struct {
```
zouyx: RPC not rpc

protocol/rpc_status.go
CodingSinger: done.

protocol/rpc_status.go
CodingSinger: done.

protocol/rpc_status.go
CodingSinger: done.

PR #343, opened 2020-01-26
Optimize zk ListenServiceEvent listener 

hxmhlt:
**What this PR does**:

Fixes #

```release-note

```


remoting/zookeeper/client.go
```diff
@@ -513,7 +514,7 @@ func (z *ZookeeperClient) GetChildrenW(path string) ([]string, <-chan zk.Event,
 		return nil, nil, perrors.Errorf("path{%s} has none children", path)
 	}
 	if len(children) == 0 {
-		return nil, nil, perrors.Errorf("path{%s} has none children", path)
+		return nil, nil, errNilChildren
```
zouyx: i think you should print the zkpath as before, otherwise users don't know what is going on.

remoting/zookeeper/client.go
```diff
@@ -544,7 +545,7 @@ func (z *ZookeeperClient) GetChildren(path string) ([]string, error) {
 		return nil, perrors.Errorf("path{%s} has none children", path)
 	}
 	if len(children) == 0 {
-		return nil, perrors.Errorf("path{%s} has none children", path)
+		return nil, errNilChildren
```
zouyx: as above

remoting/zookeeper/listener.go
```diff
@@ -272,50 +280,50 @@ func timeSecondDuration(sec int) time.Duration {
...
-		}
-		logger.Infof("listen dubbo service key{%s}", dubboPath)
-		go func(zkPath string, listener remoting.DataListener) {
-			if l.ListenServiceNodeEvent(zkPath) {
-				listener.DataChange(remoting.Event{Path: zkPath, Action: remoting.EventTypeDel})
-			}
-			logger.Warnf("listenSelf(zk path{%s}) goroutine exit now", zkPath)
-		}(dubboPath, listener)
-	}
+	//var (
```
zouyx: should be deleted rather than change to comment?

remoting/zookeeper/client.go
AlexStocks: @zouyx , @hxmhlt used the errNilChildren later. so there is no need to add the zkPath in the  error. maybe an error log is better.

zouyx: @AlexStocks ok.

AlexStocks:
LGTM

PR #344, opened 2020-01-30
Ftr/Transfer tracing context for dubbo protocol

flycash:
**What this PR does**:
We can transfer the tracing context end to end now

### DIFF
--- a/common/constant/key.go
+++ b/common/constant/key.go
@@ -78,7 +78,7 @@ const (
 
 const (
 	DUBBOGO_CTX_KEY = "dubbogo-ctx"
-	CONTEXT_KEY = "context"
+	CONTEXT_KEY     = "context"
 )
 
 const (
--- a/common/constant/key.go
+++ b/common/constant/key.go
@@ -78,6 +78,7 @@ const (
 
 const (
 	DUBBOGO_CTX_KEY = "dubbogo-ctx"
+	CONTEXT_KEY = "context"
 )
 
 const (
--- a/common/constant/key.go
+++ b/common/constant/key.go
@@ -78,7 +78,6 @@ const (
 
 const (
 	DUBBOGO_CTX_KEY = "dubbogo-ctx"
-	CONTEXT_KEY     = "context"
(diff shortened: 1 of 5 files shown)


common/constant/key.go
```diff
@@ -78,6 +78,7 @@ const (
 
 const (
 	DUBBOGO_CTX_KEY = "dubbogo-ctx"
+	CONTEXT_KEY     = "context"
```
zouyx: what is that key use for ? it should add some comment

flycash: It's not used, so I remove it.

protocol/dubbo/listener.go
```diff
@@ -327,6 +331,22 @@ func (h *RpcServerHandler) OnCron(session getty.Session) {
 	}
 }
 
+// rebuild the context by attachment.
```
AlexStocks: as go lint said, the func name should be the first word of its comment.

protocol/dubbo/listener.go
```diff
@@ -327,6 +331,22 @@ func (h *RpcServerHandler) OnCron(session getty.Session) {
...
+// Once we decided to transfer more context's key-value, we should change this.
+// now we only support rebuild the tracing context
+func rebuildCtx(inv *invocation.RPCInvocation) context.Context {
+	ctx := context.Background()
+
+	// actually, if user do not use any opentracing framework, it will always be error.
```
AlexStocks: it will always be error. --> the error will not be nil.

protocol/dubbo/listener.go
```diff
@@ -327,6 +331,22 @@ func (h *RpcServerHandler) OnCron(session getty.Session) {
...
+	spanCtx, err := opentracing.GlobalTracer().Extract(opentracing.TextMap,
+		opentracing.TextMapCarrier(inv.Attachments()))
+
```
AlexStocks: delete this blank line.

AlexStocks:
LGTM

PR #354, opened 2020-02-07
grpc client end and server end tracing function implementation

flixgithub:
**What this PR does**:
dubbo-go grpc client and server end tracing function

Fixes #

Use otgrpc ( github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc )  to implement grpc client end and server end tracing, but not sure the way i use this package is suitable for dubbo-go project internally, pls have a look

```release-note

```

### DIFF
--- a/protocol/grpc/client.go
+++ b/protocol/grpc/client.go
@@ -45,11 +45,11 @@ type Client struct {
 // NewClient ...
 func NewClient(url common.URL) *Client {
 	var (
-		conn *grpc.ClientConn
-		err error
+		conn   *grpc.ClientConn
+		err    error
 		tracer opentracing.Tracer
 	)
-   // if global trace instance was set in filter, it means trace function enabled
+	// if global trace instance was set in filter, it means trace function enabled
 	if opentracing.IsGlobalTracerRegistered() {
 		tracer = opentracing.GlobalTracer()
 		conn, err = grpc.Dial(url.Location, grpc.WithInsecure(), grpc.WithBlock(),
--- a/protocol/grpc/client.go
+++ b/protocol/grpc/client.go
@@ -18,7 +18,12 @@ limitations under the License.
 package grpc
 
 import (
+	"github.com/opentracing/opentracing-go"
 	"reflect"
+)
+
+import (
+	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
 )
 
 import (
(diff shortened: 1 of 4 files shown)


protocol/grpc/client.go
```diff
@@ -18,9 +18,14 @@ limitations under the License.
 package grpc
 
 import (
+	"github.com/opentracing/opentracing-go"
```
zouyx: split package.

protocol/grpc/server.go
```diff
@@ -19,10 +19,15 @@ package grpc
 
 import (
 	"fmt"
+	"github.com/opentracing/opentracing-go"
```
zouyx: split package.

flixgithub: already split

protocol/grpc/client.go
flixgithub: already split

protocol/grpc/client.go
```diff
@@ -39,9 +44,24 @@ type Client struct {
...
-	conn, err := grpc.Dial(url.Location, grpc.WithInsecure(), grpc.WithBlock())
-	if err != nil {
-		panic(err)
+	var (
+		conn   *grpc.ClientConn
+		err    error
+		tracer opentracing.Tracer
+	)
+	// if global trace instance was set in filter, it means trace function enabled
+	if opentracing.IsGlobalTracerRegistered() {
```
flycash: You don't need to check this because if the GlobalTracer is not set, the opentracing will do nothing and return a mockTracer.

protocol/grpc/server.go
```diff
@@ -55,15 +60,23 @@ type DubboGrpcService interface {
...
 	)
 	addr = url.Location
 	lis, err := net.Listen("tcp", addr)
 	if err != nil {
 		panic(err)
 	}
-	server := grpc.NewServer()
+
+	// if global trace instance was set, then server tracer instance can be get, and span context can also be get
+	if opentracing.IsGlobalTracerRegistered() {
```
flycash: As mentioned above

protocol/grpc/client.go
```diff
@@ -21,6 +21,11 @@ import (
 	"reflect"
 )
 
+import (
+	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
+	"github.com/opentracing/opentracing-go"
+)
+
 import (
 	"google.golang.org/grpc"
```
zouyx: should merge this block to above

protocol/grpc/server.go
```diff
@@ -23,6 +23,11 @@ import (
 	"reflect"
 )
 
+import (
+	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
+	"github.com/opentracing/opentracing-go"
+)
+
 import (
 	"google.golang.org/grpc"
```
zouyx: should merge this block to above

protocol/grpc/client.go
flixgithub: hi, if GlobalTracer not set, it means the tracer function not enabled in dubbo-go Grpc，so i think it is necessary to check this first

protocol/grpc/server.go
flixgithub: As mentioned above

AlexStocks:
@zouyx pls own this pr and merge it.

zouyx:
@AlexStocks ok

PR #357, opened 2020-02-10
Ftr : nacos config

zouyx:
**What this PR does**:

Fixes #

```release-note

```


AlexStocks:
zouyx,  u'd better add some remark for those open funcs.

zouyx:
i will fix it today , fixed test case yesterday

config_center/nacos/client.go
```diff
@@ -0,0 +1,218 @@
...
+
+// WithNacosName Set nacos name
+func WithNacosName(name string) option {
+	return func(opt *options) {
+		opt.nacosName = name
+	}
+}
+
+// ValidateNacosClient Validate nacos client , if null then create it
+func ValidateNacosClient(container nacosClientFacade, opts ...option) error {
```
flycash: Is possible that this method is invoked concurrently? If so, if-null and then create will cause some problem.

config_center/nacos/client.go
```diff
@@ -0,0 +1,218 @@
...
+			close(n.exit)
+		},
+	}
+
+	svrConfList := []nacosconst.ServerConfig{}
+	for _, nacosAddr := range n.NacosAddrs {
+		split := strings.Split(nacosAddr, ":")
+		port, err := strconv.ParseUint(split[1], 10, 64)
+		if err != nil {
+			continue
```
flycash: log the error

config_center/nacos/factory.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/apache/dubbo-go/common"
+	"github.com/apache/dubbo-go/common/extension"
+	"github.com/apache/dubbo-go/config_center"
+	"github.com/apache/dubbo-go/config_center/parser"
+)
+
+func init() {
+	extension.SetConfigCenterFactory("nacos", func() config_center.DynamicConfigurationFactory { return &nacosDynamicConfigurationFactory{} })
```
flycash: I think the nacosDynamicConfigurationFactory is singleton but the `func()` will always create new one.

config_center/nacos/impl.go
```diff
@@ -0,0 +1,167 @@
...
+		return false
+	default:
+		return true
+	}
+}
+
+func (r *nacosDynamicConfiguration) closeConfigs() {
+	r.cltLock.Lock()
+	defer r.cltLock.Unlock()
+	logger.Infof("begin to close provider nacos client")
```
flycash: How about change to:
``` go
r.cltLock.Lock()
client := r.client
r.client = nil
r.cltLock.Unlock()
client.Close()
```

config_center/nacos/listener.go
```diff
@@ -0,0 +1,61 @@
...
+		newListener[listener] = cancel
+		l.keyListeners.Store(key, newListener)
+	} else {
+		// TODO check goroutine alive, but this version of go_nacos_sdk is not support.
+		logger.Infof("profile:%s. this profile is already listening", key)
+	}
+}
+
+func (l *nacosDynamicConfiguration) removeListener(key string, listener config_center.ConfigurationListener) {
+	// TODO not supported in current go_nacos_sdk version
```
flycash: log some information...

config_center/nacos/factory.go
zouyx: you can check zookeeper implement, it is the same to nacos.
https://github.com/apache/dubbo-go/blob/33f82cd154a3ee5ef861c6465333c4945c9d4b1a/config_center/zookeeper/factory.go#L28

config_center/nacos/client.go
zouyx: may be invoked concurrently, so it has a lock in ``` container.SetNacosClient(newClient) ``` and ``` container.NacosClient().SetClient(&client) ```.  Finally added if-null logic .

AlexStocks:
LGTM

PR #358, opened 2020-02-14
Fix:deal the panic when invoker destroy

pantianying:
**What this PR does**:

Fixes #

```release-note

```

### DIFF
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -73,7 +73,8 @@ func (di *DubboInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
 		result protocol.RPCResult
 	)
 	if di.reqNum == -1 {
-		// Generally, this is not the case, because the invoker has been removed from the invoker list before destory
+		// Generally, the case will not happen, because the invoker has been removed
+		// from the invoker list before destroy,so no new request will enter the destroyed invoker
 		logger.Warnf("this dubboInvoker is destroyed")
 		result.Err = ErrDestroyedInvoker
 	}
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -72,7 +72,7 @@ func (di *DubboInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
 		err    error
 		result protocol.RPCResult
 	)
-	if di.reqNum == -1 {
+	if di.reqNum < 0 {
 		// Generally, the case will not happen, because the invoker has been removed
 		// from the invoker list before destroy,so no new request will enter the destroyed invoker
 		logger.Warnf("this dubboInvoker is destroyed")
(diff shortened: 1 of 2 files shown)


protocol/dubbo/dubbo_invoker.go
```diff
@@ -47,6 +47,7 @@ var (
 
 // DubboInvoker ...
```
AlexStocks: add comment for this file, pls.

protocol/dubbo/dubbo_invoker.go
```diff
@@ -47,6 +47,7 @@ var (
...
 type DubboInvoker struct {
+	sync.RWMutex
```
AlexStocks: u should add a lock name.

protocol/dubbo/dubbo_invoker.go
```diff
@@ -109,11 +112,15 @@ func (di *DubboInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
 
 // Destroy ...
 func (di *DubboInvoker) Destroy() {
+	di.Lock()
```
AlexStocks: the lock scope is so large.

protocol/dubbo/dubbo_invoker.go
```diff
@@ -62,6 +63,8 @@ func NewDubboInvoker(url common.URL, client *Client) *DubboInvoker {
 
 // Invoke ...
 func (di *DubboInvoker) Invoke(ctx context.Context, invocation protocol.Invocation) protocol.Result {
+	di.RLock()
```
AlexStocks: so large lock scope.

protocol/dubbo/dubbo_invoker.go
Patrick0308: May be using a `isDestroyed` status is good.

protocol/dubbo/dubbo_invoker.go
Patrick0308: Before `di.RLock()`, I suggest  to judge where this is be destroyed or not. If already being destroyed, func should return error soon.

protocol/dubbo/dubbo_invoker.go
```diff
@@ -110,11 +120,20 @@ func (di *DubboInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
 // Destroy ...
 func (di *DubboInvoker) Destroy() {
 	di.quitOnce.Do(func() {
```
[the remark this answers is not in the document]
hxmhlt: the reqNum is used for judging if dubbo invoker is available ? Why not use isAvailable() func instead?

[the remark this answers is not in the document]
pantianying: This is used to check whether there are any stock requests in the invoker during destroy

hxmhlt:
@flycash I think maybe graceful down should consider the issue.

protocol/dubbo/dubbo_invoker.go
```diff
@@ -110,11 +125,21 @@ func (di *DubboInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
...
+				if di.client != nil {
+					di.client.Close()
+					di.client = nil
+				}
+				break
+			}
+			logger.Warnf("DubboInvoker is to be destroyed, wait {%v} req end,url:{%s}", di.reqNum, di.GetUrl().Key())
+			time.Sleep(1 * time.Second)
 		}
+
```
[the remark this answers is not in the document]
pantianying: In extreme cases，there is always a request in the invoker that has not ended. If it is forcibly destroyed, it may cause a program panic. The severity of this result is greater than adding a loop. I changed this destroy to asynchronous. Please check it

PR #372, opened 2020-02-24
Fix/ignore apollo prop file

zjhmale:
after ut, apollo will generate a trivial config file.

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	config_center/apollo/mockDubbog.properties.json
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ remoting/zookeeper/zookeeper-4unittest/
 config_center/zookeeper/zookeeper-4unittest/
 registry/zookeeper/zookeeper-4unittest/
 registry/consul/agent*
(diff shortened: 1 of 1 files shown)


AlexStocks:
@zjhmale thx for your pr. the target branch should be develop. thank you.

zouyx:
@zjhmale thanks for your PR. In develop branch, UT will delete this file once it finish. you can try it.

https://github.com/apache/dubbo-go/blob/67ccfb23bfa27ac1ee178a2a704467b15316ac47/config_center/apollo/impl_test.go#L269

zjhmale:
ok, thanks for the hint, will checkout `develop` then

PR #375, opened 2020-02-26
Fix remoting/etcdv3/client test cases

zjhmale:
Add up more assertations for the etcdv3 client.

### DIFF
--- a/remoting/etcdv3/client_test.go
+++ b/remoting/etcdv3/client_test.go
@@ -31,6 +31,7 @@ import (
 import (
 	"github.com/coreos/etcd/mvcc/mvccpb"
 	perrors "github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 	"go.etcd.io/etcd/embed"
 	"google.golang.org/grpc/connectivity"
@@ -171,6 +172,10 @@ func (suite *ClientTestSuite) TestClientDone() {
 	}()
 
 	c.Wait.Wait()
+
+	if c.Valid() == true {
+		suite.T().Fatal("client should be invalid then")
+	}
 }
 
 func (suite *ClientTestSuite) TestClientCreateKV() {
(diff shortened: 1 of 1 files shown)


zouyx:
should make travis succes @zjhmale

zjhmale:
@zouyx pls review again :)

remoting/etcdv3/client_test.go
```diff
@@ -295,13 +299,26 @@ func (suite *ClientTestSuite) TestClientWatch() {
...
 			for _, event := range e.Events {
+				events = append(events, (*mvccpb.Event)(event))
 				t.Logf("type IsCreate %v k %s v %s", event.IsCreate(), event.Kv.Key, event.Kv.Value)
 			}
 		}
 
+		e1 := events[0]
+		e2 := events[1]
+
+		if e1.Type != mvccpb.PUT || string(e1.Kv.Key) != "name" || string(e1.Kv.Value) != "scott.wang" {
```
zouyx:
```suggestion
		assert.Contains()
```
if array is not always sorting, suggest use assert.Contains()

remoting/etcdv3/client_test.go
```diff
@@ -334,25 +351,38 @@ func (suite *ClientTestSuite) TestClientRegisterTemp() {
...
-					return
+					break
 				}
 				t.Logf("type IsCreate %v k %s v %s", event.IsCreate(), event.Kv.Key, event.Kv.Value)
 			}
 		}
+
+		e1 := events[0]
+		e2 := events[1]
+		if e1.Type != mvccpb.PUT || string(e1.Kv.Key) != "scott/wang" || e1.Kv.Value != nil {
```
zouyx: as above

remoting/etcdv3/client_test.go
zjhmale: thanks for the hint, but for the sake of event queue semantic, the watching channel should be a sequential generator which means the event created in the first place should always be the first one consumed outside the watching channel IIRC, maybe I'm wrong.

PR #378, opened 2020-02-27
Improve: add more ci environments

zjhmale:
Add more CI jobs to cover both `go1.12` and `go1.13` for `osx` and `linux`.

image


AlexStocks:
LGTM

Issue #380, opened 2020-02-29
Bug: HessianCodec failed to check package header length


witkeysa:
```Go
func (h *HessianCodec) ReadHeader(header *DubboHeader) error {
var err error
if h.reader.Size() < HEADER_LENGTH {
	return ErrHeaderNotEnough
}
buf, err := h.reader.Peek(HEADER_LENGTH)
if err != nil { // this is impossible
	return perrors.WithStack(err)
}
...
```

这一段会检查长度是否大于包头长度。由下面这段调用。

```Go
// Unmarshal ...
func (p *DubboPackage) Unmarshal(buf *bytes.Buffer, opts ...interface{}) error {
codec := hessian.NewHessianCodec(bufio.NewReaderSize(buf, buf.Len()))

// read header
err := codec.ReadHeader(&p.Header)
```

但是在bufio.NewReaderSize(buf, buf.Len())中

```Go
func NewReaderSize(rd io.Reader, size int) *Reader {
// Is it already a Reader?
b, ok := rd.(*Reader)
if ok && len(b.buf) >= size {
return b
}
if size < minReadBufferSize {
size = minReadBufferSize
}
r := new(Reader)
r.reset(make([]byte, size), rd)
return r
}
```
这里创建出的reader长度永远大于16.导致包头分包时，解包异常。

AlexStocks:
@witkeysa maybe u should submit this issue to [dubbo-go-hessian2](github.com/apache/dubbo-go-hessian2)?

witkeysa:
> @witkeysa maybe u should submit this issue to [dubbo-go-hessian2](github.com/apache/dubbo-go-hessian2)?

ok,thanks

AlexStocks:
@witkeysa pls reopen this issue.

AlexStocks:
@witkeysa I have fix this issue in https://github.com/apache/dubbo-go/pull/381  and this pr will be merged into develop branch later. pls check this branch and make sure this bug is fixed. If not, pls tell me again here.

PR #381, opened 2020-02-29
Fix: issue #380

AlexStocks:
**What this PR does**:

fix issue https://github.com/apache/dubbo-go/issues/380


protocol/dubbo/codec_test.go
```diff
@@ -72,3 +74,10 @@ func TestDubboPackage_MarshalAndUnmarshal(t *testing.T) {
 	assert.Equal(t, []interface{}{"a"}, pkgres.Body.([]interface{})[5])
 	assert.Equal(t, map[string]string{"dubbo": "2.0.2", "interface": "Service", "path": "path", "timeout": "1000", "version": "2.6"}, pkgres.Body.([]interface{})[6])
 }
+
+func TestIssue380(t *testing.T) {
```
fangyincheng: rename method name?

AlexStocks: No, this method is more clearly.

aliiohs:
LGTM

PR #398, opened 2020-03-12
user can add attachment

pantianying:
**What this PR does**:

Fixes #

```release-note

```

### DIFF
--- a/common/proxy/proxy.go
+++ b/common/proxy/proxy.go
@@ -140,6 +140,14 @@ func (p *Proxy) Implement(v common.RPCService) {
 				inv.SetAttachments(k, value)
 			}
 
+			// add user setAttachment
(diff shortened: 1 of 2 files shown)


common/proxy/proxy.go
```diff
@@ -140,6 +140,14 @@ func (p *Proxy) Implement(v common.RPCService) {
 				inv.SetAttachments(k, value)
 			}
 
+			// add user setAttachment
+			atm := invCtx.Value("attachment")
+			if m, ok := atm.(map[string]string); ok {
+				for k, value := range m {
```
hxmhlt: May null pointer error?

pantianying: when invCtx with no 'attachment',here ok is false

PR #417, opened 2020-03-17
Modify registryProtocol as a singleton to fix graceful shutdown

CodingSinger:
Fixed the bug that registry could not be deleted when  graceful shutdown.

### DIFF
--- a/registry/protocol/protocol.go
+++ b/registry/protocol/protocol.go
@@ -44,7 +44,7 @@ import (
 )
 
 var (
-	regProtocol *registryProtocol
+	regProtocol = newRegistryProtocol()
 )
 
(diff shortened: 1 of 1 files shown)


registry/protocol/protocol.go
```diff
@@ -44,7 +44,7 @@ import (
 )
 
 var (
-	regProtocol *registryProtocol
+	regProtocol = newRegistryProtocol()
```
AlexStocks: pls do not init a package global var in this way. u should init it in a init func.

CodingSinger: why?

AlexStocks: My personal programming practice.

PR #441, opened 2020-03-21
Publish config

### LINKED ISSUE #428: Align 2.7.5: Add publishConfig method in DynamicConfiguration.

publishConfig method in DynamicConfiguration.

**Why is this needed**:

If we want to implement the ServiceNameMapping, we should know how to store the mapping.

The default option(Dubbo option) is storing mappings in config center, which means that we should add publishConfig method in DynamicConfiguration interface.

The most difficult problem is that there are several implementations of DynamicConfiguration. So we should change all of those implementations. It will be a hard work.

flycash:
**What this PR does**:

Fixes #428 

```release-note

```

### DIFF
--- a/config_center/apollo/impl.go
+++ b/config_center/apollo/impl.go
@@ -25,7 +25,7 @@ import (
 )
 
 import (
-	"github.com/pkg/errors"
+	perrors "github.com/pkg/errors"
 	"github.com/zouyx/agollo"
 )
 
@@ -119,7 +119,7 @@ func getNamespaceName(namespace string, configFileFormat agollo.ConfigFileFormat
 func (c *apolloConfiguration) GetInternalProperty(key string, opts ...cc.Option) (string, error) {
 	config := agollo.GetConfig(c.appConf.NamespaceName)
 	if config == nil {
-		return "", errors.New(fmt.Sprintf("nothing in namespace:%s ", key))
+		return "", perrors.New(fmt.Sprintf("nothing in namespace:%s ", key))
 	}
 	return config.GetStringValue(key, ""), nil
 }
@@ -130,7 +130,7 @@ func (c *apolloConfiguration) GetRule(key string, opts ...cc.Option) (string, er
 
 // PublishConfig will publish the config with the (key, group, value) pair
 func (c *apolloConfiguration) PublishConfig(string, string, string) error {
-	return errors.New("unsupport operation")
+	return perrors.New("unsupport operation")
 }
 
 func (c *apolloConfiguration) GetProperties(key string, opts ...cc.Option) (string, error) {
@@ -140,7 +140,7 @@ func (c *apolloConfiguration) GetProperties(key string, opts ...cc.Option) (stri
 	 */
 	config := agollo.GetConfig(key)
 	if config == nil {
-		return "", errors.New(fmt.Sprintf("nothing in namespace:%s ", key))
+		return "", perrors.New(fmt.Sprintf("nothing in namespace:%s ", key))
 	}
 	return config.GetContent(agollo.Properties), nil
 }
--- a/config_center/apollo/impl.go
+++ b/config_center/apollo/impl.go
@@ -130,8 +130,7 @@ func (c *apolloConfiguration) GetRule(key string, opts ...cc.Option) (string, er
 
 // PublishConfig will publish the config with the (key, group, value) pair
 func (c *apolloConfiguration) PublishConfig(string, string, string) error {
-	// todo(@zouyx)
-	return nil
+	return errors.New("unsupport operation")
 }
 
 func (c *apolloConfiguration) GetProperties(key string, opts ...cc.Option) (string, error) {
--- a/config_center/apollo/impl.go
+++ b/config_center/apollo/impl.go
@@ -128,6 +128,12 @@ func (c *apolloConfiguration) GetRule(key string, opts ...cc.Option) (string, er
 	return c.GetInternalProperty(key, opts...)
 }
 
+// PublishConfig will publish the config with the (key, group, value) pair
+func (c *apolloConfiguration) PublishConfig(string, string, string) error {
+	// todo(@zouyx)
+	return nil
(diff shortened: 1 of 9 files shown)


config_center/nacos/impl_test.go
```diff
@@ -60,12 +60,7 @@ func runMockConfigServer(configHandler func(http.ResponseWriter, *http.Request),
 
 func mockCommonNacosServer() *httptest.Server {
 	return runMockConfigServer(func(writer http.ResponseWriter, request *http.Request) {
-		data := `
-	dubbo.service.com.ikurento.user.UserProvider.cluster=failback
-	dubbo.service.com.ikurento.user.UserProvider.protocol=myDubbo1
-	dubbo.protocols.myDubbo.port=20000
-	dubbo.protocols.myDubbo.name=dubbo
-`
+		data := "true"
```
zouyx: why  change this mock data?

config_center/nacos/impl_test.go
```diff
@@ -93,6 +88,16 @@ func Test_GetConfig(t *testing.T) {
...
 }
 
+func TestNacosDynamicConfiguration_PublishConfig(t *testing.T) {
+	nacos, err := initNacosData(t)
+	assert.Nil(t, err)
+	key := "myKey"
+	group := "/custom/a/b"
+	value := "MyValue"
+	err = nacos.PublishConfig(key, group, value)
+	assert.Nil(t, err)
```
zouyx: You verify the mock data after ```PublishConfig``` in zookeeper , but not in nacos , why.?

remoting/zookeeper/client.go
```diff
@@ -392,8 +392,16 @@ func (z *ZookeeperClient) Close() {
...
-// Create ...
+// Create will create the node recursively, which means that if the parent node is absent,
+// it will create parent node first.
+// And the value for the basePath is ""
 func (z *ZookeeperClient) Create(basePath string) error {
+	return z.CreateWithValue(basePath, []byte(""))
+}
+
+// Create will create the node recursively, which means that if the parent node is absent,
+// it will create parent node first.
```
zouyx:
```suggestion
// CreateWithValue will create the node recursively, which means that if the parent node is absent,
// it will create parent node first.
```

config_center/zookeeper/impl.go
```diff
@@ -214,3 +225,17 @@ func (c *zookeeperDynamicConfiguration) closeConfigs() {
...
+
+func (c *zookeeperDynamicConfiguration) getPath(key string, group string) string {
+	if len(key) <= 0 {
+		return c.buildPath(group)
+	}
+	return c.buildPath(group) + pathSeparator + key
+}
+
+func (c *zookeeperDynamicConfiguration) buildPath(group string) string {
+	if len(group) <= 0 {
```
AlexStocks: `len(group) == 0` should be enough.

config_center/nacos/impl.go
```diff
@@ -65,101 +66,130 @@ func newNacosDynamicConfiguration(url *common.URL) (*nacosDynamicConfiguration,
 }
 
 // AddListener Add listener
-func (n *nacosDynamicConfiguration) AddListener(key string, listener config_center.ConfigurationListener, opions ...config_center.Option) {
-	n.addListener(key, listener)
+func (nacos *nacosDynamicConfiguration) AddListener(key string, listener config_center.ConfigurationListener, opions ...config_center.Option) {
```
AlexStocks: pls do not change the receiver name. because it is used in Go way which should be a as simple as possible.

in my opinion, if u think there are so many different nacosDynamicConfiguration receiver name, what u should do is change them in a unified name like "n" or "c".

config_center/nacos/impl.go
```diff
@@ -65,101 +66,130 @@ func newNacosDynamicConfiguration(url *common.URL) (*nacosDynamicConfiguration,
...
+	group = nacos.resolvedGroup(group)
+
+	ok, err := (*nacos.client.Client()).PublishConfig(vo.ConfigParam{
+		DataId:  key,
+		Group:   group,
+		Content: value,
+	})
+
+	if err != nil {
+		return err
```
AlexStocks: why not using perrors.WithStack(err)?

config_center/nacos/impl_test.go
flycash: After I change the data to "true" it always is true...

config_center/nacos/impl_test.go
flycash: Actually, other test do not check this data. But in PublishConfig, the nacos client will check the data, and only if the data is "true", nacos client publish service successfully.

So I change this data.

config_center/apollo/impl.go
```diff
@@ -128,6 +128,11 @@ func (c *apolloConfiguration) GetRule(key string, opts ...cc.Option) (string, er
 	return c.GetInternalProperty(key, opts...)
 }
 
+// PublishConfig will publish the config with the (key, group, value) pair
+func (c *apolloConfiguration) PublishConfig(string, string, string) error {
+	return errors.New("unsupport operation")
```
fangyincheng: rename `error` to `perrors`

config_center/nacos/listener.go
```diff
@@ -35,28 +35,28 @@ func callback(listener config_center.ConfigurationListener, namespace, group, da
...
-		err := (*l.client.Client()).ListenConfig(vo.ConfigParam{
+		err := (*n.client.Client()).ListenConfig(vo.ConfigParam{
 			DataId: key,
 			Group:  "dubbo",
 			OnChange: func(namespace, group, dataId, data string) {
 				go callback(listener, namespace, group, dataId, data)
 			},
 		})
-		logger.Errorf("nacos : listen config fail, error:%v ", err)
+		logger.Errorf("n : listen config fail, error:%v ", err)
```
fangyincheng: when this error is throwed, user will not know what is the `n` . so I suggest `nacos`

PR #453, opened 2020-03-24
Service name mapping support

### LINKED ISSUE #419: Align 2.7.5: Service name mapping between interface-level and application-level

ServiceNameMapping, see org.apache.dubbo.metadata.ServiceNameMapping

**Why is this needed**:

Once we support application-level registry, we should think about one question:

How to connect those two different model?

(interface, group, version) <=> (application)

So we should build a ServiceNameMapping to resolve this problem.

flycash:
**What this PR does**:

Fixes #419 

```release-note

```

### DIFF
--- a/config_center/apollo/impl.go
+++ b/config_center/apollo/impl.go
@@ -25,6 +25,7 @@ import (
 )
 
 import (
+	gxset "github.com/dubbogo/gost/container/set"
 	perrors "github.com/pkg/errors"
 	"github.com/zouyx/agollo"
 )
@@ -131,6 +132,11 @@ func (c *apolloConfiguration) GetRule(key string, opts ...cc.Option) (string, er
 // PublishConfig will publish the config with the (key, group, value) pair
 func (c *apolloConfiguration) PublishConfig(string, string, string) error {
 	return perrors.New("unsupport operation")
+}
+
+// GetConfigKeysByGroup will return all keys with the group
+func (c *apolloConfiguration) GetConfigKeysByGroup(group string) (*gxset.HashSet, error) {
+	return nil, perrors.New("unsupport operation")
 }
 
 func (c *apolloConfiguration) GetProperties(key string, opts ...cc.Option) (string, error) {
--- a/config_center/dynamic_configuration.go
+++ b/config_center/dynamic_configuration.go
@@ -23,7 +23,9 @@ import (
 
 import (
 	gxset "github.com/dubbogo/gost/container/set"
+)
 
+import (
 	"github.com/apache/dubbo-go/common"
 	"github.com/apache/dubbo-go/config_center/parser"
 )
--- a/config_center/dynamic_configuration.go
+++ b/config_center/dynamic_configuration.go
@@ -22,13 +22,15 @@ import (
 )
 
 import (
+	gxset "github.com/dubbogo/gost/container/set"
+
 	"github.com/apache/dubbo-go/common"
 	"github.com/apache/dubbo-go/config_center/parser"
 )
 
-//////////////////////////////////////////
+// ////////////////////////////////////////
 // DynamicConfiguration
-//////////////////////////////////////////
+// ////////////////////////////////////////
 const (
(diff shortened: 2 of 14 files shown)


config_center/dynamic_configuration.go
```diff
@@ -22,13 +22,15 @@ import (
 )
 
 import (
+	gxset "github.com/dubbogo/gost/container/set"
```
zouyx: split package?

config_center/apollo/impl.go
```diff
@@ -25,6 +25,7 @@ import (
 )
 
 import (
+	gxset "github.com/dubbogo/gost/container/set"
```
hxmhlt: I used the `set` interface defined in `"github.com/emirpasic/gods/sets"` as func's return parameter. Should we have a talk about the set data structure? I think we can either rewrite our set container structure or use `"github.com/emirpasic/gods/sets"`

config_center/zookeeper/impl.go
```diff
@@ -159,6 +160,24 @@ func (c *zookeeperDynamicConfiguration) PublishConfig(key string, group string,
...
+	if err != nil {
+		return nil, perrors.WithStack(err)
+	}
+
+	if len(result) == 0 {
+		return nil, perrors.New("could not find keys with group: " + group)
+	}
+	set := gxset.NewSet()
+	for _, ele := range result {
+		set.Add(ele)
```
hxmhlt: ele -> e or v ？

config_center/apollo/impl.go
flycash: Using `github.com/emirpasic/gods/sets` and then remove `github.com/dubbogo/gost/container/set` is better.

AlexStocks:
hi, guy, is it a good idea naming a package like " dynamic_mapping" or "in_memory"? I do not think so.

hxmhlt:
> hi, guy, is it a good idea naming a package like " dynamic_mapping" or "in_memory"? I do not think so.

What's the matter?

flycash:
I don't know what is the common practice in Golang. Following the common practice is a good idea.

AlexStocks:
> I don't know what is the common practice in Golang. Following the common practice is a good idea.



> I don't know what is the common practice in Golang. Following the common practice is a good idea.

refers to the package name in github.com/golang/go.

AlexStocks:
> I don't know what is the common practice in Golang. Following the common practice is a good idea.

refers to the package name in github.com/golang/go.

zouyx:
what about ```mapping/dynamic``` and ```mapping/memory```

metadata/namemapping/dynamic/service_name_mapping.go
```diff
@@ -0,0 +1,87 @@
...
+func (d *DynamicConfigurationServiceNameMapping) Get(serviceInterface string, group string, version string, protocol string) (*gxset.HashSet, error) {
+	return d.dc.GetConfigKeysByGroup(d.buildGroup(serviceInterface, group, version, protocol))
+}
+
+// buildGroup will return group, now it looks like defaultGroup/serviceInterface
+func (d *DynamicConfigurationServiceNameMapping) buildGroup(
+	serviceInterface string,
+	group string,
+	version string,
+	protocol string) string {
```
pantianying: Since don't need the version group field, can you consider removing the input parameter

metadata/namemapping/dynamic/service_name_mapping.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/apache/dubbo-go/metadata"
+)
+
+const (
+	defaultGroup = config_center.DEFAULT_GROUP
+	slash        = "/"
+
+	// metadata service is the admin service, should not be mapped
+	metadataService = "org.apache.dubbo.metadata.MetadataService"
+)
```
hxmhlt: Define a constant in constant dir. It will be referenced by metadata module too.

PR #454, opened 2020-03-24
Add: GetInterface for rpc_service.go

fangyincheng:
**What this PR does**:

Fixes #

```release-note

```

### DIFF
--- a/common/rpc_service.go
+++ b/common/rpc_service.go
@@ -153,6 +153,7 @@ type serviceMap struct {
 	interfaceMap map[string][]*Service          // interface -> service
 }
 
+// GetService get a service defination by protocol and name
 func (sm *serviceMap) GetService(protocol, name string) *Service {
 	sm.mutex.RLock()
 	defer sm.mutex.RUnlock()
@@ -175,6 +176,7 @@ func (sm *serviceMap) GetInterface(interfaceName string) []*Service {
 	return nil
 }
 
+// Register register a service by @interfaceName and @protocol
 func (sm *serviceMap) Register(interfaceName, protocol string, rcvr RPCService) (string, error) {
 	if sm.serviceMap[protocol] == nil {
 		sm.serviceMap[protocol] = make(map[string]*Service)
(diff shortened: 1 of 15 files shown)


AlexStocks:
serviceMap   map[string]map[string]*Service // protocol -> service name -> servicefirst of all,

fangyincheng:
> serviceMap map[string]map[string]*Service // protocol -> service name -> service
> 
> first of all, I do not think this type of structure is a good idea. maybe `protocol + servcie name` as the map key is better.

We will get easyly service number of a protocol through this way, and I think `protocol + service`  is inappropriate.

common/rpc_service.go
```diff
@@ -163,10 +165,23 @@ func (sm *serviceMap) GetService(protocol, name string) *Service {
...
+func (sm *serviceMap) GetInterface(interfaceName string) []*Service {
+	sm.mutex.RLock()
+	defer sm.mutex.RUnlock()
+	if s, ok := sm.interfaceMap[interfaceName]; ok {
+		return s
+	}
+	return nil
+}
+
+func (sm *serviceMap) Register(interfaceName, protocol string, rcvr RPCService) (string, error) {
```
zouyx: Add some comment?

common/rpc_service.go
```diff
@@ -201,12 +216,13 @@ func (sm *serviceMap) Register(protocol string, rcvr RPCService) (string, error)
...
 	sm.mutex.Lock()
 	sm.serviceMap[protocol][s.name] = s
+	sm.interfaceMap[interfaceName] = append(sm.interfaceMap[interfaceName], s)
 	sm.mutex.Unlock()
 
 	return strings.TrimSuffix(methods, ","), nil
 }
 
-func (sm *serviceMap) UnRegister(protocol, serviceId string) error {
+func (sm *serviceMap) UnRegister(interfaceName, protocol, serviceId string) error {
```
zouyx: as above

common/rpc_service.go
```diff
@@ -216,15 +232,32 @@ func (sm *serviceMap) UnRegister(protocol, serviceId string) error {
...
+		sm.mutex.RUnlock()
+		return perrors.New("no service for " + interfaceName)
+	}
+	index := -1
+	for i, svr := range svrs {
+		if svr == s {
+			index = i
+		}
+	}
 	sm.mutex.RUnlock()
```
AlexStocks: this code block should be in a lambda func as follows
func() {
lock()
defer unlock()

}()

AlexStocks:
LGTM

PR #474, opened 2020-04-11
add simplified attribute in registry config

gaoxinge:
**What this PR does**:

- add simplified attribute in registry config

Adding a simplified attribute in registry config is for:

- Firstly, there is a simplified attribute in [java version](http://dubbo.apache.org/en-us/docs/user/references/xml/dubbo-registry.html).

- Secondly, simplified provider urls can release the pressure of the registry center. And we can move other parameters in provider ursl to metedata center and config center.

The code follows the [java version](https://github.com/apache/dubbo/blob/317e62f73fd0a9bac4b34eb8b2c31bbae591ef91/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java#L204), but not strictly.

```release-note
NONE
```

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -195,7 +195,6 @@ func NewURLWithOptions(opts ...option) *URL {
 // NewURL will create a new url
 // the urlString should not be empty
 func NewURL(urlString string, opts ...option) (URL, error) {
-
 	var (
 		err          error
 		rawUrlString string
@@ -471,8 +470,6 @@ func (c URL) GetParamInt(s string, d int64) int64 {
 
 // GetMethodParamInt ...
 func (c URL) GetMethodParamInt(method string, key string, d int64) int64 {
-	c.paramsLock.RLock()
-	defer c.paramsLock.RUnlock()
 	r, err := strconv.Atoi(c.GetParam("methods."+method+"."+key, ""))
 	if r == 0 || err != nil {
 		return d
--- a/common/url.go
+++ b/common/url.go
@@ -609,7 +609,7 @@ func (c *URL) CloneWithParams(reserveParams []string) *URL {
 	params := url.Values{}
 	for _, reserveParam := range reserveParams {
 		v := c.GetParam(reserveParam, "")
-		if v != "" {
+		if len(v) != 0 {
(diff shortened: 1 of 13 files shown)


common/url.go
```diff
@@ -454,33 +455,28 @@ func (c URL) GetRawParam(key string) string {
...
 	}
 	return int64(r)
 }
 
 // GetMethodParamInt ...
 func (c URL) GetMethodParamInt(method string, key string, d int64) int64 {
-	var r int
-	var err error
 	c.paramsLock.RLock()
 	defer c.paramsLock.RUnlock()
```
zouyx, line 477: why do you add a read lock here?

``` func (c URL) GetParam(s string, d string) string ```  has lock too

File: common/url.go [hunk 2]
```diff
@@ -615,8 +609,29 @@ func (c *URL) Clone() *URL {
 	return newUrl
 }
 
+func (c *URL) CloneWithParams(reserveParams []string, methods []string) *URL {
```
zouyx, line 612: Should add a  comment for  exported method

gaoxinge: Add comment.

common/url.go
gaoxinge: You are right. I restore the code.

gaoxinge: You are right. I remove the lock.

common/url.go
```diff
@@ -615,8 +604,30 @@ func (c *URL) Clone() *URL {
 	return newUrl
 }
 
+// Copy url based on the reserved parameters' keys.
+func (c *URL) CloneWithParams(reserveParams []string) *URL {
+	params := url.Values{}
+	for _, reserveParam := range reserveParams {
+		v := c.GetParam(reserveParam, "")
+		if v != "" {
```
fangyincheng, line 612: len(v)==0

fangyincheng, line 612: len(v)==0

gaoxinge: Nice. I fixed this.

gaoxinge: This is same as above.

PR #476, opened 2020-04-13
Ftr: start provider with random port

zouyx:
**What this PR does**:

* if want to use random port in provider then don't provide port in server config file , like 
```
protocols:
  "dubbo":
    name: "dubbo"
#    port: 20000
```

Fixes #

```release-note

```


config/service_config.go
```diff
@@ -105,6 +106,20 @@ func NewServiceConfig(id string, context context.Context) *ServiceConfig {
 	}
 }
 
+// Get Random Ports with size
+func getRandomPorts(size int) []int {
```
flycash, line 110: Do we need to check whether the port is used by other applications?

zouyx: maybe.........

zouyx: @flycash recheck pls

config/service_config.go
```diff
@@ -105,6 +106,45 @@ func NewServiceConfig(id string, context context.Context) *ServiceConfig {
...
+		addr = nil
+		go func() {
+			listener, err := net.Listen("tcp", ":0")
+			if err != nil {
+				return
+			}
+
+			flag = true
+			addr = listener.Addr().(*net.TCPAddr)
+		}()
```
AlexStocks, line 137: localAddr := net.TCPAddr{
			IP:   net.IPv4zero,
			Port: 0,
		}
		//listener, err := net.Listen("tcp", ":0") 这种方式在有些虚机环境下，可能拿不到随机端口
		listener, err := net.ListenTCP("tcp4", &localAddr)
		if err != nil {
			return
		}

		flag = true
		addr = listener.Addr().(*net.TCPAddr)

config/service_config.go
```diff
@@ -105,6 +106,45 @@ func NewServiceConfig(id string, context context.Context) *ServiceConfig {
...
+	var (
+		flag bool
+		addr *net.TCPAddr
+	)
+	i := 0
+	for i < size {
+		flag = false
+		addr = nil
+		go func() {
+			listener, err := net.Listen("tcp", ":0")
```
watermelo, line 130: Whether it is necessary to use the goroutine here, because one goroutine block and wait for 50 milliseconds, then 10 addrs will take at least 500 milliseconds. Personally, I think it can be dealt with in the for loop.

config/service_config.go
AlexStocks: 注意，如果用户在配置文件中指定了 IP 地址，则

       localAddr := net.TCPAddr{
		IP:   user-defined-ip,
		Port: 0,
	}

config/service_config.go
AlexStocks: agree with u.

config/service_config_test.go
```diff
@@ -18,6 +18,7 @@
 package config
 
 import (
+	"github.com/stretchr/testify/assert"
```
fangyincheng, line 21: format

config/service_config.go
zouyx: agree

config/service_config_test.go
zouyx: agree

config/service_config.go
```diff
@@ -105,6 +106,16 @@ func NewServiceConfig(id string, context context.Context) *ServiceConfig {
 	}
 }
 
+// Get Random Port
+func getRandomPort(ip string) string {
+	tcp, err := gxnet.ListenOnTCPRandomPort(ip)
+	if err != nil {
+		panic(perrors.New(fmt.Sprintf("Get tcp port error,err is {%v}", err)))
+	}
+	defer tcp.Close()
```
AlexStocks, line 115: why close the listener?

config/service_config.go
```diff
@@ -20,6 +20,7 @@ package config
 import (
 	"context"
 	"fmt"
+	gxnet "github.com/dubbogo/gost/net"
```
AlexStocks, line 23: ^_^

config/service_config.go
```diff
@@ -123,10 +134,12 @@ func (c *ServiceConfig) Export() error {
 	regUrls := loadRegistries(c.Registry, providerConfig.Registries, common.PROVIDER)
 	urlMap := c.getUrlMap()
 	protocolConfigs := loadProtocol(c.Protocol, providerConfig.Protocols)
-	if len(protocolConfigs) == 0 {
+	protocolSize := len(protocolConfigs)
+	if protocolSize == 0 {
 		logger.Warnf("The service %v's '%v' protocols don't has right protocolConfigs ", c.InterfaceName, c.Protocol)
```
hxmhlt, line 139: what is the difference?

zouyx: ```protocolSize``` will use in following code before - -

AlexStocks:
there are three things left to do. Number 1,  wait for the latest dubbogo/getty pr. Number 2, @zouyx pls change your example in ur first comment to let followers know how to use this feature clearly. 

for examples:

```text
# server.yml
protocols:
  "dubbo":
    name: "dubbo"
#    port: 20000
```

or 


```text
# server.yml
protocols:
  "dubbo":
    name: "dubbo"
    ip: "192.168.0.1"
```

or other examples.

Number 3, if the server or the server examples of dubbo-samples do not support set ip, pls add or improve it if you think the binding-ip is a important feature.

zouyx:
I think this is the best implement in random port , because it compatibility with  `jsonrpc` ,`grpc` and `rest`, that three protocols are not use getty to start server .

rest:
https://github.com/apache/dubbo-go/blob/f1de84807d4b47d8b9af5634e667998b32c9a3f4/protocol/rest/server/server_impl/go_restful_server.go#L76

jsonrpc:
https://github.com/apache/dubbo-go/blob/f1de84807d4b47d8b9af5634e667998b32c9a3f4/protocol/jsonrpc/server.go#L232

grpc:
https://github.com/apache/dubbo-go/blob/f1de84807d4b47d8b9af5634e667998b32c9a3f4/protocol/grpc/server.go#L103

config/service_config.go
```diff
@@ -105,6 +107,24 @@ func NewServiceConfig(id string, context context.Context) *ServiceConfig {
...
+		if len(proto.Port) > 0 {
+			continue
+		}
+
+		tcp, err := gxnet.ListenOnTCPRandomPort(proto.Ip)
+		if err != nil {
+			panic(perrors.New(fmt.Sprintf("Get tcp port error,err is {%v}", err)))
+		}
+		ports.PushBack(strings.Split(tcp.Addr().String(), ":")[1])
+		defer tcp.Close()
```
hxmhlt, line 123: move defer tcp.Close after line 118.

config/service_config.go
```diff
@@ -123,10 +143,14 @@ func (c *ServiceConfig) Export() error {
 	regUrls := loadRegistries(c.Registry, providerConfig.Registries, common.PROVIDER)
 	urlMap := c.getUrlMap()
 	protocolConfigs := loadProtocol(c.Protocol, providerConfig.Protocols)
-	if len(protocolConfigs) == 0 {
+	protocolSize := len(protocolConfigs)
```
hxmhlt, line 146: Why do this change?

config/service_config_test.go
```diff
@@ -23,6 +23,8 @@ import (
 
 import (
 	"github.com/apache/dubbo-go/common/extension"
+	gxnet "github.com/dubbogo/gost/net"
+	"github.com/stretchr/testify/assert"
```
hxmhlt, line 27: format

config/service_config.go
```diff
@@ -116,11 +116,13 @@ func getRandomPort(protocolConfigs []*ProtocolConfig) *list.List {
 		}
 
 		tcp, err := gxnet.ListenOnTCPRandomPort(proto.Ip)
+		if tcp != nil {
```
AlexStocks, line 119: 老何坑你，你不能 argue 下啊。这代码也太 ugly 了。

zouyx: 我没理解到何老板的用意，我的错。

AlexStocks:
LGTM

PR #492, opened 2020-04-25
Mod: add comment and modify log level

fangyincheng:


common/rpc_service.go
```diff
@@ -59,7 +59,6 @@ type AsyncCallback func(response CallbackResponse)
 //     return map[string][string]{}
 // }
 const (
-	// METHOD_MAPPER ...
 	METHOD_MAPPER = "MethodMapper"
```
zouyx, line 62: add comment for this variable?

PR #497, opened 2020-04-28
Implement event dispatch method

flycash:
**What this PR does**:
Last time when I implement the NacosServiceDiscovery, the DispatchEvent method is marked as TODO.

Now I do implement this.

### DIFF
--- a/registry/nacos/service_discovery.go
+++ b/registry/nacos/service_discovery.go
@@ -237,7 +237,7 @@ func (n *nacosServiceDiscovery) DispatchEventForInstances(serviceName string, in
 
 // DispatchEvent will dispatch the event
 func (n *nacosServiceDiscovery) DispatchEvent(event *registry.ServiceInstancesChangedEvent) error {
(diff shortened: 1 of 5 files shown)


registry/nacos/service_discovery.go
```diff
@@ -237,7 +237,7 @@ func (n *nacosServiceDiscovery) DispatchEventForInstances(serviceName string, in
 
 // DispatchEvent will dispatch the event
 func (n *nacosServiceDiscovery) DispatchEvent(event *registry.ServiceInstancesChangedEvent) error {
-	// TODO(waiting for event dispatcher, another task)
+	extension.GetGlobalDispatcher().Dispatch(event)
```
zouyx, line 240: In the other implements.Should they call ```extension.GetGlobalDispatcher().Dispatch(event)``` in `DispatchEvent`  mehod as well?

flycash: Yes, but I think we could allow users to configure the dispatcher.

PR #506, opened 2020-05-07
add the comment of ConsistentHash

aliiohs:
**What this PR does**:

Fixes #

```release-note

```


cluster/loadbalance/consistent_hash.go
```diff
@@ -85,22 +85,25 @@ func (lb *ConsistentHashLoadBalance) Select(invokers []protocol.Invoker, invocat
 	return selector.Select(invocation)
 }
 
-// Uint32Slice ...
+// Uint32Slice Slice of uint32
 type Uint32Slice []uint32
```
zouyx, line 89: I think make this struct to private would be better.

cluster/loadbalance/consistent_hash.go
```diff
@@ -53,16 +53,16 @@ func init() {
 	extension.SetLoadbalance(ConsistentHash, NewConsistentHashLoadBalance)
 }
 
-// ConsistentHashLoadBalance ...
+// ConsistentHashLoadBalance Implementation of load balancing: using consistent hashing
```
Patrick0308, line 56: See https://golang.org/doc/effective_go.html#commentary, after function name, the first character of word don't need to be capitalized.

cluster/loadbalance/consistent_hash.go
flycash: please move this to gost. We already have similar structure...

hxmhlt: > please move this to gost. We already have similar structure...

+1

aliiohs: already replaced Uint32Slice  by gxsort.Uint32Slice

cluster/loadbalance/consistent_hash.go
```diff
@@ -27,7 +27,9 @@ import (
 	"strconv"
 	"strings"
 )
-
+import (
+	gxsort "github.com/dubbogo/gost/sort"
+)
```
zouyx, line 32:
```suggestion

import (
	gxsort "github.com/dubbogo/gost/sort"
)

```
i think should add a blank line before and after this block

PR #509, opened 2020-05-09
filter registry url params starting with dot

CodingSinger:
filter registry url params starting with dot.

### DIFF
--- a/common/constant/key.go
+++ b/common/constant/key.go
@@ -210,9 +210,9 @@ const (
 	// consumer
 	CONSUMER = "consumer"
 	// key of access key id
-	ACCESS_KEY_ID_KEY = "accessKeyId"
+	ACCESS_KEY_ID_KEY = ".accessKeyId"
 	// key of secret access key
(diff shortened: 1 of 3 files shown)


CodingSinger:
> Will it cause compatibility issues between new version and old version?
Usually not, unless you use a key that starts with dot in the old version.

zouyx:
> Why do we need to add "."

maybe he has added in `dubbo` ? [doge]

CodingSinger:
> Why do we need to add "."

Maybe we will have some sensitive fields that we don’t want to pass to the consumer side via url. For example, secret or password...

shalk:
This is good way to  hiden ak/ak from  dubbo url.

do this feature merge into java version ?

zouyx:
> This is good way to hiden ak/ak from dubbo url.
> 
> do this feature merge into java version ?

https://github.com/apache/dubbo/pull/5878, do you want this feature? 

if you have a problem ,you can create a new issue rather than reply in closed issue.

PR #510, opened 2020-05-10
Ftr : add UnRegister and UnSubscribe for zookeeper Registry

### LINKED ISSUE #489: Support Unregister && Unsubscribe

support unregister && unsubscribe

**Why is this needed**:

mosn use dubbo-go to sub && pub, when app exit or restart, there operations are needed to maintain the service states

zouyx:
**What this PR does**:

* add UnRegister and UnSubscribe for Registry

Fixes #489

```release-note

```

### DIFF
--- a/registry/base_registry.go
+++ b/registry/base_registry.go
@@ -29,7 +29,6 @@ import (
 )
 
 import (
-	gxset "github.com/dubbogo/gost/container/set"
 	gxnet "github.com/dubbogo/gost/net"
 	perrors "github.com/pkg/errors"
 )
@@ -80,7 +79,7 @@ type FacadeBasedRegistry interface {
 	// DoSubscribe actually subscribe the URL
 	DoSubscribe(conf *common.URL) (Listener, error)
 	// DoUnsubscribe does unsubscribe the URL
-	DoUnsubscribe(svc *common.URL) error
+	DoUnsubscribe(conf *common.URL) (Listener, error)
 	// CloseAndNilClient close the client and then reset the client in registry to nil
 	// you should notice that this method will be invoked inside a lock.
 	// So you should implement this method as light weighted as you can.
@@ -96,12 +95,11 @@ type BaseRegistry struct {
 	context             context.Context
 	facadeBasedRegistry FacadeBasedRegistry
 	*common.URL
-	birth           int64          // time of file birth, seconds since Epoch; 0 if unknown
-	wg              sync.WaitGroup // wg+done for zk restart
-	done            chan struct{}
-	cltLock         sync.Mutex                //ctl lock is a lock for services map
-	services        map[string]common.URL     // service name + protocol -> service config, for store the service registered
-	serviceListener map[string]*gxset.HashSet // service name + protocol -> service listener, for store the service listener registered
+	birth    int64          // time of file birth, seconds since Epoch; 0 if unknown
+	wg       sync.WaitGroup // wg+done for zk restart
+	done     chan struct{}
+	cltLock  sync.Mutex            //ctl lock is a lock for services map
+	services map[string]common.URL // service name + protocol -> service config, for store the service registered
 }
 
 // InitBaseRegistry for init some local variables and set BaseRegistry's subclass to it
@@ -368,14 +366,17 @@ func sleepWait(n int) {
 
 // Subscribe :subscribe from registry, event will notify by notifyListener
 func (r *BaseRegistry) Subscribe(url *common.URL, notifyListener NotifyListener) {
-	n := 0
+	r.processSubscribe(url, notifyListener, r.facadeBasedRegistry.DoSubscribe)
+}
 
-	r.cltLock.Lock()
-	set := r.serviceListener[url.Key()]
-	if set == nil {
-		r.serviceListener[url.Key()] = gxset.NewSet()
-	}
-	r.cltLock.Unlock()
+// UnSubscribe :
+func (r *BaseRegistry) UnSubscribe(url *common.URL, notifyListener NotifyListener) {
+	r.processSubscribe(url, notifyListener, r.facadeBasedRegistry.DoUnsubscribe)
+}
+
+// process Subscribe or UnSubscribe
+func (r *BaseRegistry) processSubscribe(url *common.URL, notifyListener NotifyListener, f func(conf *common.URL) (Listener, error)) {
+	n := 0
 	for {
 		n++
 		if !r.IsAvailable() {
@@ -383,7 +384,7 @@ func (r *BaseRegistry) Subscribe(url *common.URL, notifyListener NotifyListener)
 			return
 		}
 
-		listener, err := r.facadeBasedRegistry.DoSubscribe(url)
+		listener, err := f(url)
 		if err != nil {
 			if !r.IsAvailable() {
 				logger.Warnf("event listener game over.")
@@ -401,27 +402,12 @@ func (r *BaseRegistry) Subscribe(url *common.URL, notifyListener NotifyListener)
 				break
 			} else {
 				logger.Infof("update begin, service event: %v", serviceEvent.String())
-				listener := notifyListener
-				listener.Notify(serviceEvent)
-				r.cltLock.Lock()
-				set := r.serviceListener[url.Key()]
-				set.Add(listener)
-				r.cltLock.Unlock()
+				notifyListener.Notify(serviceEvent)
 			}
 
 		}
 		sleepWait(n)
 	}
-}
-
-// UnSubscribe :
-func (r *BaseRegistry) UnSubscribe(url *common.URL, notifyListener NotifyListener) {
-	////r.serviceListener[url.Key()] = notifyListener
-	//for index, configuration := range configurations {
-	//	if configuration == a2 {
-	//		configurations = append(configurations[:index], configurations[index+1:]...)
-	//	}
-	//}
 }
 
 // closeRegisters close and remove registry client and reset services map
--- a/registry/base_registry.go
+++ b/registry/base_registry.go
@@ -411,7 +411,7 @@ func (r *BaseRegistry) Subscribe(url *common.URL, notifyListener NotifyListener)
 	}
 }
 
-// UnSubscribe :UnSubscribeURL
+// UnSubscribe URL
 func (r *BaseRegistry) UnSubscribe(url *common.URL, notifyListener NotifyListener) error {
 	if !r.IsAvailable() {
 		logger.Warnf("event listener game over.")
--- a/registry/base_registry.go
+++ b/registry/base_registry.go
@@ -56,6 +56,8 @@ func init() {
 	localIP, _ = gxnet.GetLocalIP()
 }
 
+type createPathFunc func(dubboPath string) error
+
 /*
  * -----------------------------------NOTICE---------------------------------------------
  * If there is no special case, you'd better inherit BaseRegistry and implement the
@@ -332,7 +334,7 @@ func (r *BaseRegistry) providerRegistry(c common.URL, params url.Values, createP
 }
 
 // consumerRegistry for consumer role do
-func (r *BaseRegistry) consumerRegistry(c common.URL, params url.Values, createPathFunc func(dubboPath string) error) (string, string, error) {
+func (r *BaseRegistry) consumerRegistry(c common.URL, params url.Values, f createPathFunc) (string, string, error) {
 	var (
 		dubboPath string
 		rawURL    string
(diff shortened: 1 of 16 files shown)


registry/zookeeper/registry.go
```diff
@@ -149,10 +150,18 @@ func (r *zkRegistry) DoRegister(root string, node string) error {
 	return r.registerTempZookeeperNode(root, node)
 }
 
+func (r *zkRegistry) DoUnregister(root string, node string) error {
+	return r.ZkClient().Delete(path.Join(root, node))
```
CodingSinger, line 154: Need to check the status of zkClient.

registry/zookeeper/registry.go
```diff
@@ -255,3 +264,40 @@ func (r *zkRegistry) getListener(conf *common.URL) (*RegistryConfigurationListen
 
 	return zkListener, nil
 }
+
+func (r *zkRegistry) getCloseListener(conf *common.URL) (*RegistryConfigurationListener, error) {
+
+	var zkListener *RegistryConfigurationListener
+	dataListener := r.dataListener
+	dataListener.mutex.Lock()
+	defer dataListener.mutex.Unlock()
```
hxmhlt, line 273: The datalistener's lock is so long.May be change to rwlock and make the lock time shorter.

registry/base_registry.go
```diff
@@ -332,7 +384,7 @@ func (r *BaseRegistry) Subscribe(url *common.URL, notifyListener NotifyListener)
 			return
 		}
 
-		listener, err := r.facadeBasedRegistry.DoSubscribe(url)
+		listener, err := f(url)
```
pantianying, line 387: when do unSubsrcibe,this listener is closed,But this cycle can't exit.  Is this the right logic?

registry/base_registry.go
```diff
@@ -154,6 +158,32 @@ func (r *BaseRegistry) Register(conf common.URL) error {
 	return nil
 }
 
+// UnRegister
+func (r *BaseRegistry) UnRegister(conf common.URL) error {
+	var (
+		ok  bool
+		err error
+	)
+	r.cltLock.Lock()
```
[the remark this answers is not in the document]
flycash: > pls change to RWlock

+1

registry/base_registry.go
zouyx: i will check it soon

registry/base_registry.go
```diff
@@ -154,6 +158,32 @@ func (r *BaseRegistry) Register(conf common.URL) error {
...
+	if !ok {
+		return perrors.Errorf("Path{%s} has not registered", conf.Key())
+	}
+
+	err = r.unregister(conf)
+	if err != nil {
+		return perrors.WithMessagef(err, "register(conf:%+v)", conf)
+	}
+
+	r.cltLock.Lock()
```
flycash, line 180: There is still a time gap between unregister and delete(map)..
I think we should think about:
```go
lock()
service := services[key]
delete(services, key)
unlock
err = service.unregister()
if err != nil 
   lock()
   services[key] = service
   unlock()
```
We should be more careful here.

registry/nacos/registry.go
```diff
@@ -136,6 +136,12 @@ func (nr *nacosRegistry) Register(url common.URL) error {
 	return nil
 }
 
+// UnRegister
+func (nr *nacosRegistry) UnRegister(conf common.URL) error {
+	panic(" UnRegister not support in nacosRegistry ")
```
flycash, line 141: return error

registry/nacos/registry.go
```diff
@@ -174,6 +180,11 @@ func (nr *nacosRegistry) Subscribe(url *common.URL, notifyListener registry.Noti
 	}
 }
 
+// UnSubscribe :
+func (nr *nacosRegistry) UnSubscribe(url *common.URL, notifyListener registry.NotifyListener) {
+	panic(" UnSubscribe not support in nacosRegistry ")
```
flycash, line 185: return error

registry/etcdv3/registry.go
```diff
@@ -168,3 +172,8 @@ func (r *etcdV3Registry) DoSubscribe(svc *common.URL) (registry.Listener, error)
 
 	return configListener, nil
 }
+
+func (r *etcdV3Registry) DoUnsubscribe(conf *common.URL) (registry.Listener, error) {
+	panic("DoUnsubscribe is not support in etcdV3Registry")
+	return nil, nil
```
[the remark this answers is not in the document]
zouyx: all not support method return error? but boss cao said panic would better.....

registry/base_registry.go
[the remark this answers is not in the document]
zouyx: em..............

registry/etcdv3/registry.go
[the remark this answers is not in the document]
fangyincheng: if this method of registry interface  is called in `dubbo-go`,  exist will happen with `panic`.

registry/base_registry.go
zouyx: check it again

registry/zookeeper/registry.go
zouyx: check it again

registry/base_registry.go
```diff
@@ -165,20 +165,29 @@ func (r *BaseRegistry) UnRegister(conf common.URL) error {
...
-	oldURL, ok = r.services[conf.Key()]
-	delete(r.services, conf.Key())
-	r.cltLock.Unlock()
+	func() {
+		r.cltLock.RLock()
+		defer r.cltLock.RUnlock()
+		oldURL, ok = r.services[conf.Key()]
+	}()
+
+	func() {
```
AlexStocks, line 174: look, guy, Number 1, I do not agree with ur modification of the above four lines code block. Number 2, u should not use read lock for a write action(delete). Number 3, u should use RUnlock for RLock.

remoting/zookeeper/listener.go
```diff
@@ -316,5 +316,6 @@ func (l *ZkEventListener) valid() bool {
 
 // Close ...
```
AlexStocks, line 317: pls add comment for this func. thx.

registry/base_registry.go
```diff
@@ -323,39 +376,54 @@ func sleepWait(n int) {
...
-				return
-			}
-			logger.Warnf("getListener() = err:%v", perrors.WithStack(err))
-			time.Sleep(time.Duration(RegistryConnDelay) * time.Second)
-			continue
+	for {
+		if serviceEvent, err := listener.Next(); err != nil {
+			logger.Warnf("Selector.watch() = error{%v}", perrors.WithStack(err))
+			listener.Close()
+			break
```
pantianying, line 419: 这里break，return nil会造成整个Subscribe退出 ，这样会造成zk重连后丢失Subscribe

pantianying:
suggested to run off the scene of ZK disconnection at last

registry/zookeeper/registry_test.go
```diff
@@ -74,6 +99,39 @@ func Test_Subscribe(t *testing.T) {
...
+func Test_NoSubscribe(t *testing.T) {
+	regurl, _ := common.NewURL("registry://127.0.0.1:1111", common.WithParamsValue(constant.ROLE_KEY, strconv.Itoa(common.PROVIDER)))
+	url, _ := common.NewURL("dubbo://127.0.0.1:20000/com.ikurento.user.UserProvider", common.WithParamsValue(constant.CLUSTER_KEY, "mock"), common.WithMethods([]string{"GetUser", "AddUser"}))
+	ts, reg, _ := newMockZkRegistry(&regurl)
+
+	//provider register
+	err := reg.Register(url)
+	assert.NoError(t, err)
+
+	if err != nil {
```
fangyincheng, line 111: why add err != nil? by `assert.NoError(t, err)`, when err != nil, the ut will fail.

cch123:
LGTM

AlexStocks:
> LGTM

have u tested this pr?

registry/zookeeper/registry_test.go
zouyx: This block is the same to Test_subsribe ,or you mean Test_subscribe is incorrect too ?

https://github.com/apache/dubbo-go/blob/8a3c02c9b7737a12712451b16a3cdc8ce46a5c7d/registry/zookeeper/registry_test.go#L57

registry/base_registry.go
zouyx: subscribe 不修改了。。

registry/base_registry.go
```diff
@@ -274,31 +332,26 @@ func (r *BaseRegistry) providerRegistry(c common.URL, params url.Values) (string
 }
 
 // consumerRegistry for consumer role do
-func (r *BaseRegistry) consumerRegistry(c common.URL, params url.Values) (string, string, error) {
+func (r *BaseRegistry) consumerRegistry(c common.URL, params url.Values, createPathFunc func(dubboPath string) error) (string, string, error) {
```
AlexStocks, line 335: pls define a func type for "createPathFunc"

registry/zookeeper/registry.go
```diff
@@ -255,3 +269,39 @@ func (r *zkRegistry) getListener(conf *common.URL) (*RegistryConfigurationListen
...
+func (r *zkRegistry) getCloseListener(conf *common.URL) (*RegistryConfigurationListener, error) {
+
+	var zkListener *RegistryConfigurationListener
+	r.dataListener.mutex.Lock()
+	configurationListener := r.dataListener.subscribed[conf]
+	if configurationListener != nil {
+
+		zkListener, _ := configurationListener.(*RegistryConfigurationListener)
+		if zkListener != nil {
+			r.listenerLock.Lock()
```
flycash, line 282: It looks like there is not necessary to add lock here because you do not change any system status inside the if-block. It's not the check-do something scenario.

registry/zookeeper/registry_test.go
fangyincheng: it is not necessary. i think it is better to delete them. or other resons?

registry/zookeeper/registry.go
zouyx: this block reference `getListener` , then i not sure why add lock here at this moment. just  stay the same with it.

registry/zookeeper/registry_test.go
zouyx: @hxmhlt can you remember the reason? If not , can i delete it?

registry/base_registry.go
```diff
@@ -189,6 +232,18 @@ func (r *BaseRegistry) RestartCallBack() bool {
...
 func (r *BaseRegistry) register(c common.URL) error {
+	return r.processURL(c, r.facadeBasedRegistry.DoRegister, r.createPath)
+}
+
+// unregister for unregister url to registry, include init params
+func (r *BaseRegistry) unregister(c common.URL) error {
+	return r.processURL(c, r.facadeBasedRegistry.DoUnregister, nil)
+}
+
+func (r *BaseRegistry) processURL(c common.URL, f func(string, string) error, createPathFunc func(dubboPath string) error) error {
```
AlexStocks, line 243: omg. this func definition should be :

func (r *BaseRegistry) processURL(c common.URL, f func(string, string) error, createPath createPathFunc) error

registry/base_registry.go
```diff
@@ -213,24 +268,31 @@ func (r *BaseRegistry) register(c common.URL) error {
...
 		return perrors.WithMessagef(err, "register Node(path:%s, url:%s)", dubboPath, rawURL)
 	}
 	return nil
 }
 
+// createPath will create dubbo path in register
+func (r *BaseRegistry) createPath(dubboPath string) error {
+	r.cltLock.Lock()
+	defer r.cltLock.Unlock()
+	return r.facadeBasedRegistry.CreatePath(dubboPath)
```
AlexStocks, line 291: have u got the reason of lock facadeBasedRegistry.CreatePath by r.cltLock.Lock()?

registry/base_registry.go
```diff
@@ -358,6 +413,38 @@ func (r *BaseRegistry) Subscribe(url *common.URL, notifyListener NotifyListener)
 	}
 }
 
+// UnSubscribe URL
+func (r *BaseRegistry) UnSubscribe(url *common.URL, notifyListener NotifyListener) error {
+	if !r.IsAvailable() {
+		logger.Warnf("event listener game over.")
+		return nil
```
AlexStocks, line 420: the error is nil?

registry/base_registry.go
```diff
@@ -358,6 +413,38 @@ func (r *BaseRegistry) Subscribe(url *common.URL, notifyListener NotifyListener)
...
+	}
+
+	listener, err := r.facadeBasedRegistry.DoUnsubscribe(url)
+	if err != nil {
+		if !r.IsAvailable() {
+			logger.Warnf("event listener game over.")
+			return nil
```
AlexStocks, line 427: the error is nil?

registry/base_registry.go
zouyx: the source code as below, i just extract it.
https://github.com/apache/dubbo-go/blob/8a3c02c9b7737a12712451b16a3cdc8ce46a5c7d/registry/base_registry.go#L240

https://github.com/apache/dubbo-go/blob/8a3c02c9b7737a12712451b16a3cdc8ce46a5c7d/registry/base_registry.go#L288

https://github.com/apache/dubbo-go/blob/8a3c02c9b7737a12712451b16a3cdc8ce46a5c7d/registry/base_registry.go#L300

PR #519, opened 2020-05-18
remove lock from url

### LINKED ISSUE #511: API design problem causes lock copy
Currently,  dubbo-go NewURL returns a non-ptr instance of URL，NewURLWithOptions returns a ptr of URL, 

```go
func NewURL(urlString string, opts ...option) (URL, error) {

func NewURLWithOptions(opts ...option) *URL {
```

and Register API uses a non-ptr instance of URL:

```go
func (r *BaseRegistry) Register(conf common.URL) error {
```

This will cause a lock copy, because URL contains a mutex field:

```go
type baseUrl struct {
	Protocol string
	Location string // ip+port
	Ip       string
	Port     string

	paramsLock   sync.RWMutex   <------ here
	params       url.Values
	PrimitiveURL string
}
```

The Register API should accept a pointer to avoid this

flycash:
**What this PR does**:

Remove the RWLock from baseURL

As my understanding, the URL struct should be immutable. But there are some Set* and Add* methods which can modify the URL's instance.

However, we need those methods to create a URL's instance. But the instance should not be modified anymore after finish initialization.

Another concern is that when we try to clone a URL instance, the lock will be cloned too. It will cause some concurrent problems.

Thus remove this lock will be more reasonable. 

As the expense of that, the developer should be more careful to use the URL, especially when they decide to invoke Set* or Add* method.

Fixes #511 

```release-note

```


common/url.go
```diff
@@ -425,8 +424,8 @@ func (c URL) GetParams() url.Values {
 
 // GetParamAndDecoded ...
 func (c URL) GetParamAndDecoded(key string) (string, error) {
-	c.paramsLock.RLock()
-	defer c.paramsLock.RUnlock()
+	// c.paramsLock.RLock()
+	// defer c.paramsLock.RUnlock()
```
zouyx, line 428: should you delete it?

common/url.go
```diff
@@ -409,8 +408,8 @@ func (c *URL) RangeParams(f func(key, value string) bool) {
 
 // GetParam ...
 func (c URL) GetParam(s string, d string) string {
-	c.paramsLock.RLock()
-	defer c.paramsLock.RUnlock()
+	// c.paramsLock.RLock()
+	// defer c.paramsLock.RUnlock()
```
zouyx, line 412: should you delete it?

zouyx:
resolve conflicts

common/url.go
flycash: done

common/url.go
flycash: done

PR #536, opened 2020-05-22
Opt: opt err handling

watermelo:
**What this PR does**:
optimize error handling

Fixes #

```release-note

```

### DIFF
--- a/config/metadata_report_config.go
+++ b/config/metadata_report_config.go
@@ -103,7 +103,7 @@ func startMetadataReport(metadataType string, metadataReportConfig *MetadataRepo
 	if url, err := metadataReportConfig.ToUrl(); err == nil {
 		instance.GetMetadataReportInstance(url)
 	} else {
-		return perrors.New("MetadataConfig is invalid!")
+		return perrors.Wrap(err, "Start MetadataReport failed.")
 	}
 
 	return nil
--- a/registry/directory/directory.go
+++ b/registry/directory/directory.go
@@ -152,9 +152,11 @@ func (dir *RegistryDirectory) refreshInvokers(res *registry.ServiceEvent) {
 }
 
 func (dir *RegistryDirectory) toGroupInvokers() []protocol.Invoker {
-	newInvokersList := []protocol.Invoker{}
+	var (
+		err             error
+		newInvokersList []protocol.Invoker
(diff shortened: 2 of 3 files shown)


config/metadata_report_config.go
```diff
@@ -103,7 +103,7 @@ func startMetadataReport(metadataType string, metadataReportConfig *MetadataRepo
 	if url, err := metadataReportConfig.ToUrl(); err == nil {
 		instance.GetMetadataReportInstance(url)
 	} else {
-		return perrors.New("MetadataConfig is invalid!")
+		return perrors.Wrap(err, "Start MetadataReport failed.")
```
AlexStocks, line 106: why use Wrap instead of New?

watermelo: fuction metadataReportConfig.ToUrl() returns the err, but original codes ignore it and return a new error, actual err is the same as new error. so wrap it to get more infomation.

watermelo: > why use Wrap instead of New?

fuction metadataReportConfig.ToUrl() returns the err, but original codes ignore it and return a new error, actual err is the same as new error. so wrap it to get more infomation.

registry/directory/directory.go
```diff
@@ -178,9 +181,13 @@ func (dir *RegistryDirectory) toGroupInvokers() []protocol.Invoker {
...
 			staticDir := directory.NewStaticDirectory(invokers)
-			cluster := extension.GetCluster(dir.GetUrl().SubURL.GetParam(constant.CLUSTER_KEY, constant.DEFAULT_CLUSTER))
-			staticDir.BuildRouterChain(invokers)
-			groupInvokersList = append(groupInvokersList, cluster.Join(staticDir))
+			cst := extension.GetCluster(dir.GetUrl().SubURL.GetParam(constant.CLUSTER_KEY, constant.DEFAULT_CLUSTER))
+			err = staticDir.BuildRouterChain(invokers)
+			if err != nil {
+				logger.Error(err)
+				continue
+			}
```
Patrick0308, line 189: I like this
```suggestion
			if err = staticDir.BuildRouterChain(invokers); err != nil {
				logger.Error(err)
				continue
			}
```

PR #539, opened 2020-05-24
Fix: defer in loop will cause resource leak

watermelo:
**What this PR does**:

Fixes #

```release-note

```

### DIFF
--- a/config/service_config.go
+++ b/config/service_config.go
@@ -119,8 +119,8 @@ func getRandomPort(protocolConfigs []*ProtocolConfig) *list.List {
 		if err != nil {
 			panic(perrors.New(fmt.Sprintf("Get tcp port error,err is {%v}", err)))
 		}
-		defer tcp.Close()
 		ports.PushBack(strings.Split(tcp.Addr().String(), ":")[1])
+		tcp.Close()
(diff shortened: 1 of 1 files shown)


config/service_config.go
```diff
@@ -119,8 +119,8 @@ func getRandomPort(protocolConfigs []*ProtocolConfig) *list.List {
 		if err != nil {
 			panic(perrors.New(fmt.Sprintf("Get tcp port error,err is {%v}", err)))
 		}
-		defer tcp.Close()
 		ports.PushBack(strings.Split(tcp.Addr().String(), ":")[1])
+		tcp.Close()
```
zouyx, line 123: I thought may panic in line 122 ,so use defer in loop .
In this case ,  resoure usage will under control . 
what do you think?

flycash, line 123: I am not very sure whether `tcp.Addr().String()` will return something like "127.0.0.1" which will cause `ports.PushBack(strings.Split(tcp.Addr().String(), ":")[1])` panic.

watermelo: updated.

watermelo: updated.

PR #547, opened 2020-05-31
Mod: update the comments in metrics

williamfeng323:
**What this PR does**:
Mod: Updated the comments in metrics, corrected the grammar.

### DIFF
--- a/metrics/prometheus/reporter.go
+++ b/metrics/prometheus/reporter.go
@@ -23,7 +23,12 @@ import (
 	"strings"
 	"sync"
 	"time"
+)
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
 
+import (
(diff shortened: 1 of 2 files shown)


metrics/prometheus/reporter.go
```diff
@@ -23,19 +23,15 @@ import (
 	"strings"
 	"sync"
 	"time"
-)
-import (
```
AlexStocks, line 27: pls reserve this import block. And do not merge it into the next block.

metrics/reporter.go
```diff
@@ -20,16 +20,15 @@ package metrics
 import (
 	"context"
 	"time"
-)
-import (
+
```
zouyx, line 23: revert this change pls

PR #578, opened 2020-06-01
Rft : Optimize lock for zookeeper registry

zouyx:
**What this PR does**:

1. Optimize lock for zookeeper registry , change to RWLock.

In zookeeper client , we can see too many write lock in it. Even read operation use write lock also.
So i think read operation use read lock will better in performance.

```release-note

```

### DIFF
--- a/registry/zookeeper/registry.go
+++ b/registry/zookeeper/registry.go
@@ -156,15 +156,13 @@ func (r *zkRegistry) DoRegister(root string, node string) error {
 }
 
 func (r *zkRegistry) DoUnregister(root string, node string) error {
-	client := r.client
-	if client == nil {
-		return perrors.New("zk Client is null, can not process registerTempZookeeperNode ")
-	}
+	r.cltLock.Lock()
+	defer r.cltLock.Unlock()
 
-	if !client.ZkConnValid() {
+	if !r.ZkClient().ZkConnValid() {
 		return perrors.Errorf("zk client is not valid.")
 	}
-	return client.Delete(path.Join(root, node))
+	return r.ZkClient().Delete(path.Join(root, node))
 }
 
 func (r *zkRegistry) DoSubscribe(conf *common.URL) (registry.Listener, error) {
@@ -255,7 +253,9 @@ func (r *zkRegistry) getListener(conf *common.URL) (*RegistryConfigurationListen
 
 	zkListener = NewRegistryConfigurationListener(r.client, r, conf)
 	if r.listener == nil {
+		r.cltLock.Lock()
 		client := r.client
+		r.cltLock.Unlock()
 		if client == nil {
 			return nil, perrors.New("zk connection broken")
 		}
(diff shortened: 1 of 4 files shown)


registry/zookeeper/registry.go
```diff
@@ -156,12 +156,15 @@ func (r *zkRegistry) DoRegister(root string, node string) error {
 }
 
 func (r *zkRegistry) DoUnregister(root string, node string) error {
-	r.cltLock.Lock()
-	defer r.cltLock.Unlock()
-	if !r.ZkClient().ZkConnValid() {
+	client := r.client
```
zouyx, line 159: copy new var for operation

registry/zookeeper/registry.go
```diff
@@ -252,9 +255,7 @@ func (r *zkRegistry) getListener(conf *common.URL) (*RegistryConfigurationListen
 
 	zkListener = NewRegistryConfigurationListener(r.client, r, conf)
 	if r.listener == nil {
-		r.cltLock.Lock()
```
zouyx, line 255: `client := r.client` is atomic

remoting/zookeeper/client.go
```diff
@@ -446,9 +444,7 @@ func (z *ZookeeperClient) Delete(basePath string) error {
 	)
 
 	err = errNilZkClientConn
-	z.Lock()
 	conn := z.Conn
```
zouyx, line 447: as above

remoting/zookeeper/client.go
```diff
@@ -468,9 +464,7 @@ func (z *ZookeeperClient) RegisterTemp(basePath string, node string) (string, er
 	err = errNilZkClientConn
 	data = []byte("")
 	zkPath = path.Join(basePath) + "/" + node
-	z.Lock()
 	conn := z.Conn
```
zouyx, line 467: as above

remoting/zookeeper/client.go
```diff
@@ -419,9 +419,7 @@ func (z *ZookeeperClient) CreateWithValue(basePath string, value []byte) error {
 	for _, str := range strings.Split(basePath, "/")[1:] {
 		tmpPath = path.Join(tmpPath, "/", str)
 		err = errNilZkClientConn
-		z.Lock()
 		conn := z.Conn
```
zouyx, line 422: `conn := z.Conn` is atomic

remoting/zookeeper/client.go
```diff
@@ -487,7 +493,9 @@ func (z *ZookeeperClient) RegisterTempSeq(basePath string, data []byte) (string,
 	)
 
 	err = errNilZkClientConn
+	z.RLock()
```
AlexStocks, line 496: big old, maybe u can define a func as follows

```Go
func (z *zookeeperClient) getConn() net.Conn {
        z.RLock()
        defer z.RUnlock()

	return z.Conn
}

Issue #616, opened 2020-06-20
Enhancement: gitee quality analyses -- common


georgehao:
**What would you like to be added**:

**Why is this needed**:

PR #627, opened 2020-06-23
fix:  fix gitee sonar code check(issue #616)

relaxedCat:


common/rpc_service.go
```diff
@@ -68,7 +68,7 @@ var (
 	typeOfError = reflect.TypeOf((*error)(nil)).Elem()
 
 	// ServiceMap store description of service.
-	// todo: lowerecas?
+	// TODO
```
fangyincheng, line 71: todo miss `lowercas`,  you can delete it.

PR #629, opened 2020-06-24
Add: anyhost key judgment

willson-chen:
**What this PR does**:

Add anyhost judgment in ToConfigurators in registry/base_configuration_listener.go, which is used to improve code quality by removing 'TODO' comment

Fixes #

```release-note
NONE
```

### DIFF
--- a/registry/base_configuration_listener.go
+++ b/registry/base_configuration_listener.go
@@ -100,7 +100,13 @@ func ToConfigurators(urls []*common.URL, f func(url *common.URL) config_center.C
 			configurators = []config_center.Configurator{}
 			break
 		}
-		//TODO:anyhost_key judage
+
+		var override map[string][]string = url.GetParams()
+		delete(override, constant.ANYHOST_KEY)
+		if len(override) == 0 {
+			configurators = configurators[:0]
+			continue
+		}
 		configurators = append(configurators, f(url))
 	}
 	return configurators
--- a/registry/base_configuration_listener.go
+++ b/registry/base_configuration_listener.go
@@ -104,7 +104,7 @@ func ToConfigurators(urls []*common.URL, f func(url *common.URL) config_center.C
 		var override map[string][]string = url.GetParams()
 		delete(override, constant.ANYHOST_KEY)
 		if len(override) == 0 {
-			configurators = configurators[:0]
+			configurators = nil
 			continue
 		}
 		configurators = append(configurators, f(url))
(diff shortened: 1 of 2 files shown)


willson-chen:
I removed the `TODO` comment and add the anyhost judgment, which follows the [java implement ](https://github.com/apache/dubbo/blob/b8b42b16182b8fde245dd6c5edeb1115955d1004/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/Configurator.java#L86).

registry/base_configuration_listener.go
```diff
@@ -100,7 +100,13 @@ func ToConfigurators(urls []*common.URL, f func(url *common.URL) config_center.C
 			configurators = []config_center.Configurator{}
 			break
 		}
-		//TODO:anyhost_key judage
+
+		var override map[string][]string = url.GetParams()
```
AlexStocks, line 104: override := url.GetParams()

willson-chen:
> Based on https://github.com/apache/dubbo-go/blob/master/contributing.md to change the title

Thanks. I will pay attention to the rule in next PR.

registry/base_configuration_listener.go
willson-chen: Done. Thanks. :)

registry/base_configuration_listener.go
```diff
@@ -100,7 +100,13 @@ func ToConfigurators(urls []*common.URL, f func(url *common.URL) config_center.C
 			configurators = []config_center.Configurator{}
 			break
 		}
-		//TODO:anyhost_key judage
+
+		override := url.GetParams()
+		delete(override, constant.ANYHOST_KEY)
+		if len(override) == 0 {
+			configurators = nil
```
zouyx, line 107: It seems will clear `configurators` when url.GetParams()  is nil. Is it right logic?
Like:

Fisrt loop: url.GetParams() is not null, add a element to `configurators`.
Second loop :url.GetParams() is null ,will clear `configurators`?

willson-chen: I am also puzzled by the java implememt that I follow. Could some one help to check it?

chickenlj: > It seems will clear configurators when url.GetParams() is nil. Is it right logic?

I think `configurators  = nil` should be removed.

zouyx: > > It seems will clear configurators when url.GetParams() is nil. Is it right logic?
> 
> I think `configurators = nil` should be removed.

Wilson will fix it.

willson-chen:
Sorry for my lateness. Change request done.

PR #630, opened 2020-06-24
Rft : Extract priority router

### LINKED ISSUE #477: Optimize the design of Router
Now the Router has two key interfaces:
image

image

However, we use `Chain` interface in our codes as the entrance, like:
image

We should notice that, the Router is the top abstraction. Using `Chain` in codes will reduce the expandability.

So I proposal to refactor the design.

Firstly, we keep the `Router` interface, but remove the `Priority` method;
Secondly,  we keep `Chain` interface, but it will implement `Router` interface, which means that `Chain` is a `Router` too but it's more complicated.
Thirdly, we define `PriorityRouter` which is the `Router` interface + `Priority` method;
Finally, we will replace `Chain` with `Router` in our codes. The `Router` instance should be fetched from `extension`;

The class diagram is:
 
image

The benefit is that:
1. User can schedule many routers by their own way, not just Chain;
2. Only `Router` interface are exposed outside the Router module;

zouyx:
**What this PR does**:

* Extract priority router 

Fixes #477 

```release-note

```

### DIFF
--- a/cluster/router/chan.go
+++ b/cluster/router/chan.go
@@ -15,29 +15,11 @@
  * limitations under the License.
  */
 
-package healthcheck
+package router
 
-import (
-	"github.com/apache/dubbo-go/cluster/router"
-	"github.com/apache/dubbo-go/common"
-	"github.com/apache/dubbo-go/common/constant"
-	"github.com/apache/dubbo-go/common/extension"
-)
-
-func init() {
-	extension.SetRouterFactory(constant.HealthCheckRouterName, newHealthCheckRouteFactory)
-}
-
-// HealthCheckRouteFactory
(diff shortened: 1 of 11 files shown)


cluster/router/chan.go
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package router
+
+// Chain
```
pantianying, line 20: more  annotation please. help user to use it

PR #640, opened 2020-07-02
add namespaceId config for nacos

coffeehc:
add namespaceId config  for nacos


AlexStocks:
@coffeehc pls move the target branch from master to develop.

PR #644, opened 2020-07-04
Fix: waitGroup use

pantianying:
**What this PR does**:
fix waitGroup use and err log

### DIFF
--- a/registry/zookeeper/registry.go
+++ b/registry/zookeeper/registry.go
@@ -134,6 +134,7 @@ func (r *zkRegistry) InitListeners() {
 					regConfigListener.Close()
 				}
 				newDataListener.SubscribeURL(conf, NewRegistryConfigurationListener(r.client, r))
+				r.WaitGroup().Add(1)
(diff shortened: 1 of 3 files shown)


registry/zookeeper/registry.go
```diff
@@ -134,6 +134,7 @@ func (r *zkRegistry) InitListeners() {
 					regConfigListener.Close()
 				}
 				newDataListener.SubscribeURL(conf, NewRegistryConfigurationListener(r.client, r))
+				r.WaitGroup().Add(1)
 				go r.listener.ListenServiceEvent(conf, fmt.Sprintf("/dubbo/%s/"+constant.DEFAULT_CATEGORY, url.QueryEscape(conf.Service())), newDataListener)
```
zouyx, line 138: I think the operation which is called add wait group should relate to listen service event.
So they can extract a new method.

pantianying: Waitgroup is only used to record the number of goruntines. It is not intended to be bound to the listenserviceevent method

pantianying: Waitgroup is only used to record the number of goruntines. It is not intended to be bound to the listenserviceevent method

PR #662, opened 2020-07-18
Ftr: Feature/application and service level router

### LINKED ISSUE #596: Aligh with Dubbo: Service route configuration

Users can define the global configuration for all services.
If the service does not have its own route configuration, the global configuration will be used.

**Why is this needed**:

williamfeng323:
**What this PR does**:
1. Support multiple routers configuration in router_config.yml
2. Apply prioritization to routers
3. Enable service level routers

Fixes #596

After this change, the local router configuration file needs to follow below pattern to support multiple routing rules in local file configuration:
```release-note
routerRules: 
  - scope: application
    key: mock-app1
    priority: 1
    force: true
    conditions :
      - "a => b"
      - "c => d"
  - scope: application
    key: mock-app2
    priority: 1
    force: true
    conditions :
      - "a => b"
      - "c => d"
```
the scope defines the rule's applicable range, 'application' | 'service'
the key defines the rule's applicable application name(for application scope) or the service key(for service scope, [{group}/]{service}[:{version}])

### DIFF
--- a/cluster/directory/base_directory.go
+++ b/cluster/directory/base_directory.go
@@ -85,15 +85,21 @@ func (dir *BaseDirectory) SetRouters(urls []*common.URL) {
 	for _, url := range urls {
 		routerKey := url.GetParam(constant.ROUTER_KEY, "")
 
-		if len(routerKey) > 0 {
-			factory := extension.GetRouterFactory(url.Protocol)
-			r, err := factory.NewPriorityRouter(url)
-			if err != nil {
-				logger.Errorf("Create router fail. router key: %s, url:%s, error: %+v", routerKey, url.Service(), err)
-				return
-			}
-			routers = append(routers, r)
+		if len(routerKey) == 0 {
+			continue
 		}
+		if url.Protocol == constant.CONDITION_ROUTE_PROTOCOL {
+			if !dir.isProperRouter(url) {
+				continue
+			}
+		}
+		factory := extension.GetRouterFactory(url.Protocol)
+		r, err := factory.NewPriorityRouter(url)
+		if err != nil {
+			logger.Errorf("Create router fail. router key: %s, url:%s, error: %+v", routerKey, url.Service(), err)
+			return
+		}
+		routers = append(routers, r)
 	}
 
 	logger.Infof("Init file condition router success, size: %v", len(routers))
@@ -102,6 +108,24 @@ func (dir *BaseDirectory) SetRouters(urls []*common.URL) {
 	dir.mutex.Unlock()
 
 	rc.AddRouters(routers)
+}
+
+func (dir *BaseDirectory) isProperRouter(url *common.URL) bool {
+	app := url.GetParam(constant.APPLICATION_KEY, "")
+	serviceKey := dir.GetUrl().ServiceKey()
+	if serviceKey == "" {
+		serviceKey = dir.GetUrl().SubURL.ServiceKey()
+	}
+	if len(app) > 0 {
+		if app != dir.GetUrl().GetParam(constant.APPLICATION_KEY, "") {
+			return false
+		}
+	} else {
+		if url.ServiceKey() != serviceKey {
+			return false
+		}
+	}
+	return true
 }
 
 // Destroy Destroy
--- a/cluster/directory/base_directory.go
+++ b/cluster/directory/base_directory.go
@@ -116,14 +116,11 @@ func (dir *BaseDirectory) isProperRouter(url *common.URL) bool {
 	if serviceKey == "" {
 		serviceKey = dir.GetUrl().SubURL.ServiceKey()
 	}
-	if len(app) > 0 {
-		if app != dir.GetUrl().GetParam(constant.APPLICATION_KEY, "") {
-			return false
-		}
-	} else {
-		if url.ServiceKey() != serviceKey {
-			return false
-		}
+	if len(app) > 0 && app != dir.GetUrl().GetParam(constant.APPLICATION_KEY, "") {
+		return false
+	}
+	if url.ServiceKey() != serviceKey {
+		return false
 	}
 	return true
 }
--- a/cluster/directory/base_directory.go
+++ b/cluster/directory/base_directory.go
@@ -116,13 +116,13 @@ func (dir *BaseDirectory) isProperRouter(url *common.URL) bool {
 	if serviceKey == "" {
 		serviceKey = dir.GetUrl().SubURL.ServiceKey()
 	}
-	if len(app) > 0 && app != dir.GetUrl().GetParam(constant.APPLICATION_KEY, "") {
-		return false
+	if len(app) > 0 && app == dir.GetUrl().GetParam(constant.APPLICATION_KEY, "") {
+		return true
 	}
-	if url.ServiceKey() != serviceKey {
-		return false
+	if url.ServiceKey() == serviceKey {
+		return true
 	}
-	return true
+	return false
 }
 
 // Destroy Destroy
--- a/cluster/router/condition/file.go
+++ b/cluster/router/condition/file.go
@@ -71,7 +71,10 @@ func (f *FileConditionRouter) URL() common.URL {
 			common.WithParamsValue(constant.RouterPriority, strconv.Itoa(routerRule.Priority)),
 			common.WithParamsValue(constant.RULE_KEY, base64.URLEncoding.EncodeToString([]byte(rule))),
 			common.WithParamsValue(constant.ROUTER_KEY, constant.CONDITION_ROUTE_PROTOCOL),
-			common.WithParamsValue(constant.CATEGORY_KEY, constant.ROUTERS_CATEGORY))
+			common.WithParamsValue(constant.CATEGORY_KEY, constant.ROUTERS_CATEGORY),
+			common.WithParamsValue(constant.RouterRuleKey, routerRule.Key),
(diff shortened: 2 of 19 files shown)


cluster/router/condition/router_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"testing"
+)
+
+import (
+	"github.com/stretchr/testify/assert"
+)
+
+import (
+	"github.com/dubbogo/gost/container/set"
```
AlexStocks, line 29: this package should be placed in the 2rd import block.

config/router_config_test.go
```diff
@@ -28,9 +28,11 @@ import (
 
 import (
 	_ "github.com/apache/dubbo-go/cluster/router/condition"
+	"github.com/dubbogo/gost/container/set"
```
AlexStocks, line 31: it should be placed in the 2rd import block.

williamfeng323: resolved

cluster/router/condition/router_test.go
williamfeng323: resolved

cluster/router/condition/file.go
```diff
@@ -71,11 +72,41 @@ func (f *FileConditionRouter) URL() common.URL {
...
+			if len(grp) > 0 {
+				f.url.AddParam(constant.GROUP_KEY, grp)
+			}
+			if len(ver) > 0 {
+				f.url.AddParam(constant.VERSION_KEY, ver)
+			}
+			if len(srv) > 0 {
+				f.url.AddParam(constant.INTERFACE_KEY, srv)
+			}
+		}
```
zouyx, line 90:
```suggestion
		if routerRule.Scope == constant.RouterApplicationScope {
		f.url.AddParam(constant.APPLICATION_KEY, routerRule.Key)
		return
	}
	grp, srv, ver, e := parseServiceRouterKey(routerRule.Key)
	if e != nil {
		return
	}
	if len(grp) > 0 {
		f.url.AddParam(constant.GROUP_KEY, grp)
	}
	if len(ver) > 0 {
		f.url.AddParam(constant.VERSION_KEY, ver)
	}
	if len(srv) > 0 {
		f.url.AddParam(constant.INTERFACE_KEY, srv)
	}
```

cluster/router/condition/file.go
```diff
@@ -71,11 +72,41 @@ func (f *FileConditionRouter) URL() common.URL {
...
 	return f.url
 }
 
+func parseServiceRouterKey(key string) (string, string, string, error) {
+	if len(strings.TrimSpace(key)) == 0 {
+		return "", "", "", nil
+	}
+	reg := regexp.MustCompile(`(.*/{1})?([^:/]+)(:{1}[^:]*)?`)
+	strs := reg.FindAllStringSubmatch(key, -1)
+	if strs == nil || len(strs) != 1 {
```
zouyx, line 101:
```suggestion
	if strs == nil || len(strs) > 1 {
```

you mean greater than 1 or greater than 0?

cluster/router/condition/file.go
```diff
@@ -71,11 +72,41 @@ func (f *FileConditionRouter) URL() common.URL {
...
+		return "", "", "", perrors.Errorf("Invalid key, service key must follow [{group}/]{service}[:{version}] pattern")
+	}
+	grp := strings.TrimSpace(strings.TrimRight(strs[0][1], "/"))
+	srv := strings.TrimSpace(strs[0][2])
+	ver := strings.TrimSpace(strings.TrimLeft(strs[0][3], ":"))
```
zouyx, line 106: Should you check `len(strs[0]) ==3 ` or `len(strs[0]) >2 `

cluster/router/condition/router.go
```diff
@@ -117,7 +117,13 @@ func NewConditionRouter(url *common.URL) (*ConditionRouter, error) {
 	}
 
 	router.url = url
-	router.priority = url.GetParamInt(constant.RouterPriority, 0)
+	var defaultPriority int64
+	if url.GetParam(constant.APPLICATION_KEY, "") != "" {
+		defaultPriority = 150
+	} else {
+		defaultPriority = 140
```
zouyx, line 124: why change default priotiry ?

cluster/router/condition/router_rule.go
```diff
@@ -57,7 +58,7 @@ func getRule(rawRule string) (*RouterRule, error) {
 		return r, err
 	}
 	r.RawRule = rawRule
-	if len(r.Conditions) != 0 {
+	if len(r.Conditions) != 0 && r.Key != "" && (r.Scope == constant.RouterApplicationScope || r.Scope == constant.RouterServiceScope) {
```
zouyx, line 61:
```suggestion
	if len(r.Conditions) > 0 && len(r.Key)>0 && (r.Scope == constant.RouterApplicationScope || r.Scope == constant.RouterServiceScope) {
```

config/router_config.go
```diff
@@ -32,16 +33,37 @@ var (
...
+	}
+	if len(routerRules.RouterRules) == 0 {
+		return perrors.Errorf("No router configurations in file %s", confRouterFile)
+	}
+	fileRouterFactories := extension.GetFileRouterFactories()
+	for _, v := range routerRules.RouterRules {
+		content, _ := yaml.MarshalYML(v)
+		err = initRouterConfig(content, fileRouterFactories)
+	}
+	return err
```
zouyx, line 60: i think should add a interface here, for load the router config file.

cluster/router/condition/router.go
williamfeng323: This is the default priority for app router and service router. to ensure the service router has higher priority compare to app router if no priority is set in configuration.

cluster/router/condition/file.go
williamfeng323: i am gonna check len(strs[0]) != 4 since the regexp expected to return size == 4 slice to represent 1st slice: origin string; 2nd/3rd/4th slice: the first/second/third captured value even it is empty;

config/router_config.go
williamfeng323: Do you mean loading different types of config files to support the earlier version of router config files?

cluster/router/condition/router.go
zouyx: No, bro.....

You have got my point.

Look at the left block.....default priority is `0`, you change to `140`.

Your logic is if application router set priority to `150 , if not, set priority to `140`. But original logic is if not, set priority to `0`.

config/router_config.go
zouyx: I just mean loading different types of config files to support all router config files in `dubbo-go`.

After all, there are three  types of router in dubbo-go  at least .

* condition
* tag
* health instance first

cluster/directory/base_directory.go
```diff
@@ -104,6 +110,24 @@ func (dir *BaseDirectory) SetRouters(urls []*common.URL) {
 	rc.AddRouters(routers)
 }
 
+func (dir *BaseDirectory) isProperRouter(url *common.URL) bool {
+	app := url.GetParam(constant.APPLICATION_KEY, "")
+	serviceKey := dir.GetUrl().ServiceKey()
+	if serviceKey == "" {
+		serviceKey = dir.GetUrl().SubURL.ServiceKey()
+	}
+	if len(app) > 0 {
```
AlexStocks, line 119: the more clear codes is as follows

	if len(app) > 0 && app != dir.GetUrl().GetParam(constant.APPLICATION_KEY, "") {
		return false
	}

	if url.ServiceKey() != serviceKey {
		return false
	}

config/router_config.go
williamfeng323: as discussed, this will be checked in the factory, so do not create the interface for now. We may refactor this part in the future version.

cluster/directory/base_directory.go
williamfeng323: done.

cluster/directory/base_directory.go
```diff
@@ -104,6 +110,21 @@ func (dir *BaseDirectory) SetRouters(urls []*common.URL) {
...
+	}
+	if len(app) > 0 && app != dir.GetUrl().GetParam(constant.APPLICATION_KEY, "") {
+		return false
+	}
+	if url.ServiceKey() != serviceKey {
+		return false
+	}
+	return true
+}
+
```
pantianying, line 127: This means that when app = = = ', it must return true

cluster/router/condition/file.go
```diff
@@ -71,11 +72,47 @@ func (f *FileConditionRouter) URL() common.URL {
...
+			f.url.AddParam(constant.VERSION_KEY, ver)
+		}
+		if len(srv) > 0 {
+			f.url.AddParam(constant.INTERFACE_KEY, srv)
+		}
 	})
 	return f.url
 }
 
+func parseServiceRouterKey(key string) (string, string, string, error) {
```
pantianying, line 98: please add some Examples as comments

cluster/router/condition/router_test.go
pantianying: still have problem？

cluster/directory/base_directory.go
williamfeng323: updated.

cluster/router/condition/file.go
williamfeng323: added.

Issue #661, opened 2020-07-18
Imp: some improvements need to do in v1.5.


AlexStocks:
**What would you like to be added**:

1 delete the github.com/aliyun/aliyun-oss-go-sdk / github.com/baiyubin/aliyun-sts-go-sdk in the go.mod, which are useless.
2 "Apache Dubbo Go" should be  "Apache Dubbo-go" in NOTICE.
3 pls add "nolint" for EventType.String() and Event.String() remoting/listener.go.
4 Pls add comment for registry/servicediscovery/synthesizer/subscribed_urls_ synthesizer.go:SubscribedURLsSynthesizer.
5 Pls add comment or "nolint" for SequenceType and DubboPackage in protocol/dubbo/codec.go.
6 change the wrong word "dncode" in comment of protocol/dubbo/codec.go:DubboPackage.Unmarshal.
7 Pls add comment or "nolint" for GrpcExporter in protocol/grpc/grpc_exporter.go.
8 Pls add comment or "nolint" for all structs/funcs in protocol/grpc/grpc_invoker.go.
9 Pls add comment or "nolint" for struct Requst in protocol/jsonrpc/http.go.
10 Pls add comment or "nolint" for var Version in protocol/jsonrpc/json.go.
11 Pls add comments for func GetProtocol and FilterInvoker and its funcs in protocol/protocolwrapper/protocol_filter_wrapper.go.
12 Pls add comment or "nolint" for all structs/funcs in protocol/rest/config/rest_config.go.
13 Pls add comment or "nolint" for all structs/funcs in protocol/rest/rest_exporter.go.
14 Pls add comment or "nolint" for all structs/funcs in protocol/rest/rest_invoker.go.
15 Pls add comment or "nolint" for all structs/funcs in protocol/rest/rest_protocol.go.
16 Pls add comment or "nolint" for const FixedWindowKey in filter/filter_impl/tps/tps_limit_fix_window_stategy.go
17 Pls split the import block in filter/filter_impl/tps/tps_limit_strategy_mock.go.
18 Pls add comments  or "nolint" for consts in filter/filter_impl/access_log_filter.go.
19 Pls add comments  or "nolint" for func GetFilter() in filter/filter_impl/echo_filter.go.
20 Pls add comments  or "nolint" for struct ExecuteState in filter/filter_impl/execute_limit_filter.go.
21 Pls add comments  or "nolint" for struct GenericFilter in filter/filter_impl/generic_filter.go.
[...]

PR #664, opened 2020-07-19
Imp: Some improvements related to annotation

Howie66:
**What this PR does**:
Some improvements related to annotation,issue https://github.com/apache/dubbo-go/issues/661

if the annotation i added have any problems, please email haoweili35@gmail.com


cluster/cluster_impl/broadcast_cluster.go
```diff
@@ -39,6 +39,7 @@ func NewBroadcastCluster() cluster.Cluster {
 	return &broadcastCluster{}
 }
 
+// Join would return baseClusterInvoker instance
```
AlexStocks, line 42: the comment should be `Join returns a baseClusterInvoker instance`.

And the following comment should change as the same.

PR #687, opened 2020-08-02
Rft: Optimize remotes configuration

### LINKED ISSUE #501: Enhance: Optimize remotes configuration and constructor

Remotes Configuration.

**Why is this needed**:

Now, there are many implementations based on different middleware, including zookeeper, nacos...
However, we do not have a common way to configure those middleware.
For example, If I want to use the Nacos as registry and configure center, I must configure the Nacos in two different places:

Nacos as configure center

Nacos as registry

It's not convenient for users. If the user wants to change the address, he must change all places. 

So I propose that we should define the RemoteConfig, it looks like:
```yml
remotes:
  nacos1:
     address: "1.1.1.1:8080"

registries:
  nacos1:
    protocol: "nacos",
    remoteRef: "nacos1",

config_center:
  protocol: "nacos"
  remoteRef: "nacos1",
```

zouyx:
**What this PR does**:

* optimize remotes configuration
* refactor config center implement.
* will compatible with `baseConfig.ConfigCenterConfig.Address` and `baseConfig.ConfigCenterConfig.RemoteRef` before 1.6.0. After 1.6.0 will not compatible, only `baseConfig.ConfigCenterConfig.RemoteRef`.

Fixes #501

```release-note

```

### DIFF
--- a/config/base_config.go
+++ b/config/base_config.go
@@ -29,7 +29,6 @@ import (
 )
 
 import (
-	"github.com/apache/dubbo-go/common"
 	"github.com/apache/dubbo-go/common/config"
 	"github.com/apache/dubbo-go/common/logger"
 )
@@ -67,21 +66,6 @@ func (c *BaseConfig) GetServiceDiscoveries(name string) (config *ServiceDiscover
 func (c *BaseConfig) GetRemoteConfig(name string) (config *RemoteConfig, ok bool) {
 	config, ok = c.Remotes[name]
 	return
-}
-
-func (c *BaseConfig) toConfigCenterURL() (common.URL, error) {
-	rc, ok := GetBaseConfig().GetRemoteConfig(baseConfig.ConfigCenterConfig.RemoteRef)
-
-	if !ok {
-		return common.URL{}, perrors.New("Could not find out the remote ref config, name: " + name)
-	}
-
-	return common.NewURL(rc.Address,
-		common.WithUsername(rc.Username),
-		common.WithPassword(rc.Password),
-		common.WithLocation(rc.Address),
-		common.WithProtocol(baseConfig.ConfigCenterConfig.Protocol),
-	)
 }
 
 func getKeyPrefix(val reflect.Value) []string {
--- a/config/base_config.go
+++ b/config/base_config.go
@@ -69,8 +69,8 @@ func (c *BaseConfig) GetRemoteConfig(name string) (config *RemoteConfig, ok bool
 	return
 }
 
-func (c *BaseConfig) toURL(name string, protocol string) (common.URL, error) {
-	rc, ok := GetBaseConfig().GetRemoteConfig(name)
+func (c *BaseConfig) toConfigCenterURL() (common.URL, error) {
+	rc, ok := GetBaseConfig().GetRemoteConfig(baseConfig.ConfigCenterConfig.RemoteRef)
 
 	if !ok {
 		return common.URL{}, perrors.New("Could not find out the remote ref config, name: " + name)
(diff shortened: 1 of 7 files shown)


config/base_config.go
```diff
@@ -72,72 +69,19 @@ func (c *BaseConfig) GetRemoteConfig(name string) (config *RemoteConfig, ok bool
...
-	if len(appContent) != 0 {
-		appMapConent, err := dynamicConfig.Parser().Parse(appContent)
-		if err != nil {
-			return perrors.WithStack(err)
-		}
-		config.GetEnvInstance().UpdateAppExternalConfigMap(appMapConent)
-	}
-
-	return nil
+	return common.NewURL(rc.Address,
```
flycash, line 79: I prefer to define method inside RemoteRef:
```go
func (rc *RemoteRef) ToUrl() {
}
```
The RemoteRef is used in many modules, in this way we make sure that if we add more fileds to RemoteRef, all users is updated.

zouyx: ok

config/remote_config.go
```diff
@@ -18,6 +18,7 @@
 package config
 
 import (
+	"github.com/apache/dubbo-go/common"
```
hxmhlt, line 21: split

config/config_center_config.go
```diff
@@ -77,3 +85,94 @@ func (c *ConfigCenterConfig) GetUrlMap() url.Values {
...
+}
+
+// startConfigCenter will start the config center.
+// it will prepare the environment
+func (b *configCenter) startConfigCenter(baseConfig BaseConfig) error {
+	url, err := b.toURL(baseConfig)
+	if err != nil {
+		return err
+	}
+	if b.prepareEnvironment(baseConfig, &url) != nil {
```
gaoxinge, line 121: Why not `err = b.prepareEnvironment(baseConfig, &url)` and return `err` when it's not `nil`？

config/config_center_config.go
```diff
@@ -77,3 +85,94 @@ func (c *ConfigCenterConfig) GetUrlMap() url.Values {
...
+		return perrors.WithMessagef(err, "start config center error!")
+	}
+	// c.fresh()
+	return err
+}
+
+func (b *configCenter) prepareEnvironment(baseConfig BaseConfig, configCenterUrl *common.URL) error {
+	factory := extension.GetConfigCenterFactory(configCenterUrl.Protocol)
+	dynamicConfig, err := factory.GetDynamicConfiguration(configCenterUrl)
+	config.GetEnvInstance().SetDynamicConfiguration(dynamicConfig)
```
gaoxinge, line 131: Need this statement be placed after `if err != nil {}`?

config/config_center_config_test.go
```diff
@@ -0,0 +1,95 @@
...
+		}}
+
+	c := &configCenter{}
+	err := c.startConfigCenter(*baseConfig)
+	assert.NoError(t, err)
+	b, v := config.GetEnvInstance().Configuration().Back().Value.(*config.InmemoryConfiguration).GetProperty("dubbo.application.organization")
+	assert.True(t, b)
+	assert.Equal(t, "ikurento.com", v)
+
+	baseConfig = nil
```
gaoxinge, line 74: This statement seems useless.

config/config_center_config.go
zouyx: agree.

I just copy it.....
It may have some problem before....

config/config_center_config.go
zouyx: agree.

I just copy it.....
It may have some problem before....

config/config_center_config_test.go
zouyx: maybe ....  This `baseconfig` scope is in this test file, i just think clear it before next test.

PR #689, opened 2020-08-04
Fix: can not inovke nacos destroy when graceful shutdown

lzp0412:
1、fix not inovke nacos destroy when graceful shutdown
2、upgrade nacos client version


registry/nacos/registry.go
```diff
@@ -193,6 +224,13 @@ func (nr *nacosRegistry) IsAvailable() bool {
 
 // nolint
 func (nr *nacosRegistry) Destroy() {
+	for _, url := range nr.registryUrls {
+		err := nr.DeRegister(url)
+		logger.Infof("DeRegister Nacos url:%+v", url)
+		if err != nil {
+			logger.Errorf("Deregister url:%+v err:%v", url, err.Error())
```
AlexStocks, line 231: Deregister Nacos URL

registry/nacos/registry.go
```diff
@@ -193,6 +224,13 @@ func (nr *nacosRegistry) IsAvailable() bool {
 
 // nolint
 func (nr *nacosRegistry) Destroy() {
+	for _, url := range nr.registryUrls {
+		err := nr.DeRegister(url)
+		logger.Infof("DeRegister Nacos url:%+v", url)
```
AlexStocks, line 229: Deregister Nacos URL

registry/nacos/registry.go
```diff
@@ -128,6 +129,36 @@ func (nr *nacosRegistry) Register(url common.URL) error {
 	if !isRegistry {
 		return perrors.New("registry [" + serviceName + "] to  nacos failed")
 	}
+	nr.registryUrls = append(nr.registryUrls, url)
+	return nil
+}
+
+func createDeregisterParam(url common.URL, serviceName string) vo.DeregisterInstanceParam {
+	if len(url.Ip) == 0 {
```
hxmhlt, line 137: Pls add func comment

zouyx: i think public function only

zouyx: > Pls add func comment

i think public function only

PR #704, opened 2020-08-09
Imp: Correct words and format codes

Howie66:


protocol/dubbo/client.go
```diff
@@ -193,7 +193,7 @@ type Response struct {
 	atta  map[string]string
 }
 
-// NewResponse  create a new Response.
+// NewResponse create a new Response.
```
zouyx, line 196:
```suggestion
// NewResponse creates a new Response.
```

or // no lint

protocol/dubbo/codec.go
```diff
@@ -82,7 +82,7 @@ func (p *DubboPackage) Marshal() (*bytes.Buffer, error) {
 	return bytes.NewBuffer(pkg), nil
 }
 
-// Unmarshal dncode hessian package.
+// Unmarshal decode hessian package.
```
zouyx, line 85:
```suggestion
// Unmarshal decodes hessian package.
```

PR #708, opened 2020-08-10
Ftr: bitmap in router chain

beiwei30:
Introduce roaring bitmap in router chain.  By doing this, we can avoid iterating every invoker in Route as much as possible so that we could reduce consuming both CPU and memory. At the same time, by passing the pre-built address cache along the router chain, we can keep router's implementation lock-free at the same time.


cluster/router/chain/chain.go
```diff
@@ -79,6 +103,99 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
 	c.routers = newRouters
 }
 
+// SetInvokers receives updated invokers from registry center. If the times of notification exceeds countThreshold and
+// time interval exceeds timeThreshold since last cache update, then notify to update the cache.
```
AlexStocks, line 107: // If the times of notification exceeds countThreshold and time interval exceeds timeThreshold since last cache update

why the condition type is 'and' which is not 'OR'?

AlexStocks, line 107: // If the times of notification exceeds countThreshold and time interval exceeds timeThreshold since last cache update

why the condition type is 'and' which is not 'OR'?

AlexStocks, line 107: // If the times of notification exceeds countThreshold and time interval exceeds timeThreshold since last cache update

why the condition type is 'and' which is not 'OR'?

AlexStocks, line 107: // If the times of notification exceeds countThreshold and time interval exceeds timeThreshold since last cache update

why the condition type is 'and' which is not 'OR'?

AlexStocks, line 107: // If the times of notification exceeds countThreshold and time interval exceeds timeThreshold since last cache update

why the condition type is 'and' which is not 'OR'?

cluster/router/chain/chain.go
```diff
@@ -79,6 +103,99 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+
+	var mutex sync.Mutex
+	var wg sync.WaitGroup
+
+	for _, r := range c.copyRouters() {
+		if p, ok := r.(router.Poolable); ok {
+			wg.Add(1)
+			go func(p router.Poolable) {
+				pool, info := poolRouter(p, origin, invokers)
+				mutex.Lock()
```
AlexStocks, line 186: You'd better use `defer mutex.Unlock()` to unlock safely.

cluster/router/router.go
```diff
@@ -50,3 +55,39 @@ type PriorityRouter interface {
 	// 0 to ^int(0) is better
 	Priority() int64
 }
+
+// Poolable caches address pool and address metadata for a router instance which will be used later in Router's Route.
+type Poolable interface {
```
AlexStocks, line 60: this interface name is not a noun. So it should be 'Pool' or "RouterMetaPool'.

cluster/router/healthcheck/health_check_route.go
```diff
@@ -51,25 +58,48 @@ func NewHealthCheckRouter(url *common.URL) (router.PriorityRouter, error) {
...
 	return healthyInvokers
 }
 
+// Pool separates healthy invokers from others.
+func (r *HealthCheckRouter) Pool(invokers []protocol.Invoker) (router.AddrPool, router.AddrMetadata) {
+	if !r.enabled {
+		return nil, nil
+	}
+
+	rb := make(router.AddrPool)
```
AlexStocks, line 83: pls do not use make to create a struct obj. pls init it in this way `rb := &router.AddrPool{}`.

cluster/router/chain/chain.go
```diff
@@ -109,14 +226,62 @@ func NewRouterChain(url *common.URL) (*RouterChain, error) {
...
 
+// poolRouter calls poolable router's Pool() to create new address pool and address metadata if necessary.
+// If the corresponding cache entry exists, and the poolable router answers no need to re-pool (possibly because its
+// rule doesn't change), and the address list doesn't change, then the existing data will be re-used.
+func poolRouter(p router.Poolable, origin *InvokerCache, invokers []protocol.Invoker) (router.AddrPool, router.AddrMetadata) {
+	name := p.Name()
+	if isCacheMiss(origin, name) || p.ShouldPool() || isInvokersChanged(origin.invokers, invokers) {
+		logger.Debugf("build address cache for router %q", name)
+		return p.Pool(invokers)
+	} else {
```
AlexStocks, line 248: pls delete the else switch branch.

cluster/router/chain/chain.go
```diff
@@ -79,6 +103,99 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+	if v == nil {
+		return nil
+	}
+
+	return v.(*InvokerCache)
+}
+
+// buildCache builds address cache with the new invokers for all poolable routers.
+func (c *RouterChain) buildCache() {
+	if c.invokers == nil || len(c.invokers) == 0 {
```
AlexStocks, line 170: maybe lack of lock guard,

cluster/router/router.go
```diff
@@ -50,3 +55,39 @@ type PriorityRouter interface {
...
+	// Pool created address pool and address metadata from the invokers.
+	Pool([]protocol.Invoker) (AddrPool, AddrMetadata)
+
+	// ShouldPool returns if it should pool. One typical scenario is a router rule changes, in this case, a pooling
+	// is necessary, even if the addresses not changed at all.
+	ShouldPool() bool
```
AlexStocks, line 66: maybe u should use "Enable" as this func's name. If u add this interface or func as the Dubbo Interface, u can omit my suggestion.

cluster/router/utils/bitmap_util.go
```diff
@@ -0,0 +1,51 @@
...
+		left.And(right)
+	}
+	return left
+}
+
+func FallbackIfJoinToEmpty(left *roaring.Bitmap, right *roaring.Bitmap) *roaring.Bitmap {
+	ret := JoinIfNotEqual(left, right)
+	if ret == nil || ret.IsEmpty() {
+		return right
+	} else {
```
AlexStocks, line 42: pls delete this else switch branch.

beiwei30:
Thanks for Alex's review comment. I will take a look tomorrow.

cluster/router/chain/chain.go
beiwei30: This is because I want to make sure cache rebuilding happens as less as possible, that is, I need to make sure cache to rebuild only once when at least `countThreshold` times of address notifications happen within the given time window (specified by `timeThreshold`

beiwei30: dup.

beiwei30: dup.

cluster/router/chain/chain.go
beiwei30: done.

cluster/router/router.go
beiwei30: I still believe `Poolable` is a better interface name :). It is a common practice in Java, and even in dubbo go there's another interface named `Listenable`

cluster/router/healthcheck/health_check_route.go
beiwei30: `router.AddrPool` is not a struct, but a self-defined type backed by `map`.

cluster/router/chain/chain.go
beiwei30: I think we cannot remove `else` branch. This is because we want to reuse the previous cache result for the current router if no need to rebuild the cache. Besides address list changes, the cache also needs to be rebuilt when the router's rule (if it has) changes.

AlexStocks: What I mean is ur codes should be written as follows.

```Go
if {
  logger.xx
  return p.Pool(invokers)
}

return xxx
```

More clearly, I think.

cluster/router/utils/bitmap_util.go
AlexStocks: the same as above

```Go
if {
  logger.xx
  return p.Pool(invokers)
}

return xxx
```

cluster/router/chain/chain.go
beiwei30: I see. Done.

cluster/router/chain/chain.go
beiwei30: done.

cluster/router/router.go
beiwei30: ShouldPool is more accurate in my opinion. If a router receives a changed rule, it should re-pool the whole address list into bitmaps. That the reason why the router chain needs to ask each "Poolable" router should pool the address.

cluster/router/utils/bitmap_util.go
beiwei30: done.

cluster/router/chain/chain.go
```diff
@@ -79,6 +105,104 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+			c.buildCache()
+		case <-c.ch:
+			c.buildCache()
+		}
+	}
+}
+
+// copyRouters make a snapshot copy from RouterChain's router list.
+func (c *RouterChain) copyRouters() []router.PriorityRouter {
+	c.mutex.RLock()
```
AlexStocks, line 141: hey, guy, u should using unlock in this way `defer c.mutex.RUnlock()` to release the lock safely.

cluster/router/chain/chain.go
```diff
@@ -79,6 +105,104 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+	ret := copySlice(c.routers)
+	c.mutex.RUnlock()
+	return ret.([]router.PriorityRouter)
+}
+
+// copyInvokers copies a snapshot of the received invokers.
+func (c *RouterChain) copyInvokers() []protocol.Invoker {
+	c.mutex.RLock()
+	if c.invokers == nil || len(c.invokers) == 0 {
```
AlexStocks, line 150: OMG, u return the func value without release the lock!

cluster/router/chain/chain.go
```diff
@@ -79,6 +105,104 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+func (c *RouterChain) copyRouters() []router.PriorityRouter {
+	c.mutex.RLock()
+	ret := copySlice(c.routers)
+	c.mutex.RUnlock()
+	return ret.([]router.PriorityRouter)
+}
+
+// copyInvokers copies a snapshot of the received invokers.
+func (c *RouterChain) copyInvokers() []protocol.Invoker {
+	c.mutex.RLock()
```
AlexStocks, line 149: defer c.mutex.RUnlock()

cluster/router/chain/chain.go
```diff
@@ -48,20 +57,37 @@ type RouterChain struct {
 	mutex sync.RWMutex
 
 	url common.URL
+
+	// The times of address notification since last update for address cache
+	count int64
+	// The timestamp of last update for address cache
+	last time.Time
+	// Channel for notify to update the address cache
+	ch chan struct{}
```
AlexStocks, line 66: pls change this var name to 'notify'.

cluster/router/chain/invoker_cache.go
```diff
@@ -0,0 +1,80 @@
...
+	// Address metadata from routers which implement Poolable
+	metadatas map[string]router.AddrMetadata
+}
+
+// BuildCache builds address cache from the given invokers.
+func BuildCache(invokers []protocol.Invoker) *InvokerCache {
+	return &InvokerCache{
+		invokers:  invokers,
+		bitmap:    utils.ToBitmap(invokers),
+		pools:     make(map[string]router.AddrPool),
```
AlexStocks, line 52: pls remember that u should set a init length for map when u init a map var.

```Go
pools:     make(map[string]router.AddrPool, 8),
metadatas: make(map[string]router.AddrMetadata, 8),
```

cluster/router/router.go
AlexStocks: got it.

cluster/router/healthcheck/health_check_route.go
AlexStocks: u should init it as follows

```Go
rb := make(router.AddrPool, 8)
```

beiwei30: done.

cluster/router/chain/chain.go
beiwei30: done.

cluster/router/chain/chain.go
beiwei30: my mistake. done.

cluster/router/chain/chain.go
beiwei30: done.

cluster/router/chain/chain.go
beiwei30: done.

cluster/router/chain/invoker_cache.go
beiwei30: done.

cluster/router/chain/chain.go
```diff
@@ -109,14 +233,62 @@ func NewRouterChain(url *common.URL) (*RouterChain, error) {
...
+
+	logger.Debugf("reuse existing address cache for router %q", name)
+	return origin.pools[name], origin.metadatas[name]
+}
+
+// isCacheMiss checks if the corresponding cache entry for a poolable router has already existed.
+// False returns when the cache is nil, or cache's pool is nil, or cache's invokers snapshot is nil, or the entry
+// doesn't exist.
+func isCacheMiss(cache *InvokerCache, key string) bool {
+	if cache == nil || cache.pools == nil || cache.invokers == nil || cache.pools[key] == nil {
```
AlexStocks, line 265: u should lock cache.pools/cache.invokers/cache.pools

cluster/router/chain/chain.go
```diff
@@ -79,6 +105,104 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+		}()
+	}
+}
+
+// loop listens on events to update the address cache when it's necessary, either when it receives notification
+// from address update, or when timeInterval exceeds.
+func (c *RouterChain) loop() {
+	for {
+		select {
+		case <-time.Tick(timeInterval):
```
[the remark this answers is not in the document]
beiwei30: didn't realize this, and good to learn. Fixed.

cluster/router/chain/chain.go
```diff
@@ -79,6 +105,104 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+}
+
+// copyInvokers copies a snapshot of the received invokers.
+func (c *RouterChain) copyInvokers() []protocol.Invoker {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	if c.invokers == nil || len(c.invokers) == 0 {
+		return nil
+	}
+	ret := copySlice(c.invokers)
```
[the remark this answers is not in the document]
beiwei30: I introduce copySlice in order to avoid dup code, but I can live with it if reflection is a concern.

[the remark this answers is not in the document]
beiwei30: done.

cluster/router/chain/chain.go
```diff
@@ -79,6 +105,104 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+	return ret.([]protocol.Invoker)
+}
+
+func copySlice(s interface{}) interface{} {
```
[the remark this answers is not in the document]
beiwei30: done.

cluster/router/chain/chain.go
```diff
@@ -79,6 +105,104 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+		return nil
+	}
+
+	return v.(*InvokerCache)
+}
+
+// buildCache builds address cache with the new invokers for all poolable routers.
+func (c *RouterChain) buildCache() {
+	invokers := c.copyInvokers()
+	if invokers == nil || len(c.invokers) == 0 {
```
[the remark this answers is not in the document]
beiwei30: this is still necessary, since the copied invokers will be null if the original invoker to be copied is null.

cluster/router/healthcheck/health_check_route.go
```diff
@@ -51,25 +58,48 @@ func NewHealthCheckRouter(url *common.URL) (router.PriorityRouter, error) {
...
 	// Add healthy invoker to the list
-	for _, invoker := range invokers {
-		if r.checker.IsHealthy(invoker) {
-			healthyInvokers = append(healthyInvokers, invoker)
-		}
-	}
-	// If all Invoke are considered unhealthy, downgrade to all inovker
-	if len(healthyInvokers) == 0 {
+	healthyInvokers := utils.JoinIfNotEqual(addrPool[healthy], invokers)
+	// If all Invoke are considered unhealthy, downgrade to all invoker
```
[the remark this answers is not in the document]
beiwei30: done

cluster/router/chain/chain.go
```diff
@@ -79,6 +104,100 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+		return
+	}
+
+	cache := BuildCache(invokers)
+	origin := c.loadCache()
+
+	var mutex sync.Mutex
+	var wg sync.WaitGroup
```
hxmhlt, line 181: Pls combine the definition of variable.

cluster/router/tag/tag_router.go
```diff
@@ -76,21 +95,33 @@ func (c *tagRouter) Priority() int64 {
...
-			if v.GetUrl().GetParam(constant.Tagkey, "") == tag {
-				result = append(result, v)
+// Pool divided invokers into different address pool by tag.
+func (c *tagRouter) Pool(invokers []protocol.Invoker) (router.AddrPool, router.AddrMetadata) {
+	rb := make(router.AddrPool, 8)
+	for i, invoker := range invokers {
+		url := invoker.GetUrl()
+		tag := url.GetParam(constant.Tagkey, "")
+		if tag != "" {
+			if _, ok := rb[tag]; !ok {
```
hxmhlt, line 105: len(tag)>0

cluster/router/chain/chain.go
beiwei30: done.

cluster/router/tag/tag_router.go
beiwei30: done.

cluster/router/chain/chain.go
beiwei30: I believe it is unnecessary, since cache is fetched from atomic.Value.

common/url.go
```diff
@@ -643,6 +643,34 @@ func (c *URL) CloneWithParams(reserveParams []string) *URL {
...
+		return false
+	}
+
+	for lk, lv := range leftMap {
+		if rv, ok := rightMap[lk]; !ok {
+			return false
+		} else if lv != rv {
+			return false
+		}
+	}
```
cvictory, line 669: Can you use reflect.DeepEqual to compare map ?

cluster/router/chain/chain.go
```diff
@@ -88,6 +104,116 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+	for _, r := range c.copyRouters() {
+		if p, ok := r.(router.Poolable); ok {
+			wg.Add(1)
+			go func(p router.Poolable) {
+				pool, info := poolRouter(p, origin, invokers)
+				mutex.Lock()
+				defer mutex.Unlock()
+				cache.pools[p.Name()] = pool
+				cache.metadatas[p.Name()] = info
+				wg.Done()
```
AlexStocks, line 208: pls move it to the first line of this func as "defer wg.Done()".

common/url.go
AlexStocks: the performance of reflection package is very low in go.

AlexStocks (approved):

cluster/router/chain/chain.go
```diff
@@ -88,6 +104,116 @@ func (c *RouterChain) AddRouters(routers []router.PriorityRouter) {
...
+// SetInvokers receives updated invokers from registry center. If the times of notification exceeds countThreshold and
+// time interval exceeds timeThreshold since last cache update, then notify to update the cache.
+func (c *RouterChain) SetInvokers(invokers []protocol.Invoker) {
+	c.mutex.Lock()
+	c.invokers = invokers
+	c.mutex.Unlock()
+
+	c.count++
+	now := time.Now()
+	if c.count >= countThreshold && now.Sub(c.last) >= timeThreshold {
```
cvictory, line 116: If the data from registry is changed once， it will not notify to c.notify.  

And it will build cache when it is timeout or receive notify channel.  The worst situation is that it cannot build cache after 5 seconds.

```
//cluster/router/chain/chain.go:127
	for {
		ticker := time.NewTicker(timeInterval)
		select {
		case <-ticker.C:
			c.buildCache()
		case <-c.notify:
			c.buildCache()
		}
	}
```

cluster/router/chain/chain.go
beiwei30: done.

cluster/router/chain/chain.go
beiwei30: this is expected.

hxmhlt (approved):

fangyincheng (approved):

PR #739, opened 2020-09-07
Imp: Improve map access concurrency

skyao:
**What this PR does**:

This PR is to improve the accessing of map in concurrency environment.

1. improve the performace of GetURLStatus() method
2. fix the problem of concurrency risk in  GetMethodStatus() method

Fixes #

```release-note
NONE
```

### DIFF
--- a/protocol/rpc_status.go
+++ b/protocol/rpc_status.go
@@ -98,7 +98,10 @@ func (rpc *RPCStatus) GetSuccessiveRequestFailureCount() int32 {
 
 // GetURLStatus get URL RPC status.
 func GetURLStatus(url common.URL) *RPCStatus {
-	rpcStatus, _ := serviceStatistic.LoadOrStore(url.Key(), &RPCStatus{})
+	rpcStatus, found := serviceStatistic.Load(url.Key())
+	if !found {
+		rpcStatus, _ = serviceStatistic.LoadOrStore(url.Key(), &RPCStatus{})
+	}
 	return rpcStatus.(*RPCStatus)
 }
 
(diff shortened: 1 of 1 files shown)


zouyx:
Should change target branch to `deveop`

zouyx (commented):
Merge or rebase develop pls

protocol/rpc_status.go
```diff
@@ -98,7 +98,10 @@ func (rpc *RPCStatus) GetSuccessiveRequestFailureCount() int32 {
 
 // GetURLStatus get URL RPC status.
 func GetURLStatus(url common.URL) *RPCStatus {
-	rpcStatus, _ := serviceStatistic.LoadOrStore(url.Key(), &RPCStatus{})
+	rpcStatus, found := serviceStatistic.Load(url.Key())
```
zouyx, line 101: Why don't you use `LoadOrStore` directly? For avoid lock competition?

watermelo: It seems to have higher performance, go version is 1.13.
```go
func BenchmarkStore(b *testing.B) {
    for i := 0; i < b.N; i++ {
        _, found := methodStatistics.Load("test")
        if !found {
            _, _ = methodStatistics.LoadOrStore("test", &RPCStatus{})
        }
    }
}

func BenchmarkLoadOrStore(b *testing.B) {
    for i := 0; i < b.N; i++ {
        _, _ = methodStatistics.LoadOrStore("test", &RPCStatus{})
    }
}

BenchmarkStore-4                23008020                48.4 ns/op             0 B/op          0 allocs/op
BenchmarkLoadOrStore-4          10381435               102 ns/op              80 B/op          1 allocs/op
PASS
```

watermelo (approved):

AlexStocks (approved):

protocol/rpc_status.go
skyao: yes, the new code will save the cost of "&RPCStatus{}", which only be used when the key is not set and do "store", and late access  to this key, it always "load". But each invoke will still cost for the "&RPCStatus{}".

skyao (commented):
The standard operation to access similar map:

1. get with key, in most case it will return the stored one, make it AS FAST AS POSIBLE
2. special case, if the key is not set, we should do initial and store it to map. We will do this only once, if no concurreny.
3. More Special case, there are several thread access this NOT stored key, and all of them found the key is not set and should do initial and store, so we need to use LoadOrStore to avoid override the saved value.

zouyx (approved):

PR #775, opened 2020-09-23
Fix: url.params fatal error: concurrent map read and map write

jack15083:
**What this PR does**:
fix url.params fatal error: concurrent map read and map write
https://github.com/apache/dubbo-go/issues/772

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -78,10 +78,10 @@ func (t RoleType) Role() string {
 }
 
 type baseUrl struct {
-	Protocol     string
-	Location     string // ip+port
-	Ip           string
-	Port         string
+	Protocol string
+	Location string // ip+port
+	Ip       string
+	Port     string
 	//url.Values is not safe map, add to avoid concurrent map read and map write error
 	paramsLock   sync.RWMutex
 	params       url.Values
--- a/common/url.go
+++ b/common/url.go
@@ -397,8 +397,8 @@ func (c URL) Service() string {
 
 // AddParam will add the key-value pair
 func (c *URL) AddParam(key string, value string) {
-	c.paramsLock.RLock()
-	defer c.paramsLock.RUnlock()
+	c.paramsLock.Lock()
+	defer c.paramsLock.Unlock()
 	c.params.Add(key, value)
 }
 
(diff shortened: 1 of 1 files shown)


common/url.go
```diff
@@ -393,34 +396,33 @@ func (c URL) Service() string {
...
-	c.params.Add(key, value)
+	c.AddParam(key, value)
 }
 
 // SetParam will put the key-value pair into url
-// it's not thread safe.
-// think twice before you want to use this method
 // usually it should only be invoked when you want to initialized an url
 func (c *URL) SetParam(key string, value string) {
+	c.paramsLock.RLock()
```
AlexStocks, line 417: Hey, guy, pls using write lock instead.

common/url.go
```diff
@@ -393,34 +396,33 @@ func (c URL) Service() string {
...
+	defer c.paramsLock.RUnlock()
 	c.params.Set(key, value)
 }
 
 // RangeParams will iterate the params
-// it's not thread-safe
 func (c *URL) RangeParams(f func(key, value string) bool) {
```
AlexStocks, line 423: u should exec the @f outside the lock to limit the lock scope.

```Go
        var (
            flag bool
            key string
            value string
        )
        func() {
        c.paramsLock.RLock()
	defer c.paramsLock.RUnlock()
	   for k, v := range c.params {
		if !f(k, v[0]) {
                         key = k
                         value = v[0]
                         flag = true
			break
		}
	  }
        }()
        if flag {
          f(key, value) 
        }
```

AlexStocks (approved):

common/url.go
zouyx: Agree

common/url.go
jack15083: test this code, u should remove if flag {f(key, value)}

common/url.go
```diff
@@ -393,43 +396,59 @@ func (c URL) Service() string {
...
 // AddParamAvoidNil will add key-value pair
-// Not thread-safe
-// think twice before using it.
 func (c *URL) AddParamAvoidNil(key string, value string) {
 	if c.params == nil {
 		c.params = url.Values{}
 	}
 
-	c.params.Add(key, value)
+	c.AddParam(key, value)
```
zouyx, line 411: I think the write lock should cover all line of this method

georgehao:
pls see #511 , because many url param of function is copy, just add rwmutex may appear warnings.

watermelo (approved):

fangyincheng (approved):

PR #797, opened 2020-10-17
Imp: try to fix too many files open error

wenxuwan:
When the amount of concurrency is high, a large number of tcp links will be created, resulting in a large number of time_awaits, resulting in too many files error, as shown in the figure below.

A maximum of 1024 files are allowed to open
image

image

image


wenxuwan:
Add more information for review.

The qps and tcp connection number  as below:

image
企业微信截图_8c117092-ddf4-4726-bc10-938d7c81c040

I set poolSize to 2 and below is the test result:

企业微信截图_ae697d2d-b924-4e5d-9924-ba24c886f969
企业微信截图_0ca65779-a356-4f44-9a26-8f5a7f6e1b74


Btw, the qps value may not be accurate because it is restricted by the dubbo provider.

protocol/dubbo/pool.go
```diff
@@ -288,117 +287,113 @@ func (c *gettyRPCClient) close() error {
...
+		value, ok := p.poolQueue.popTail()
+		if ok {
+			conn := value.(*gettyRPCClient)
+			if d := now - conn.getActive(); d > p.ttl {
+				go conn.close()
+				continue
 			}
+			conn.updateActive(now)
+			return conn, nil
+		} else {
```
zouyx, line 394: `else` is unnessecary

protocol/dubbo/pool.go
```diff
@@ -288,117 +287,113 @@ func (c *gettyRPCClient) close() error {
...
 		}
 	}
-}
-
-func (p *gettyRPCClientPool) safeRemove(conn *gettyRPCClient) {
-	p.Lock()
-	defer p.Unlock()
-
-	p.remove(conn)
+	return nil, nil
```
zouyx, line 398: Why can not get connection but not return error?

wenxuwan: Can't get connection means pool is empty or all connections have expired , so just return nil  and will create new connection

protocol/dubbo/pool.go
```diff
@@ -288,117 +287,113 @@ func (c *gettyRPCClient) close() error {
...
 	}
+	return conn, perrors.WithStack(err)
+}
 
-	if len(p.conns) > 0 {
-		for idx, c := range p.conns {
-			if conn == c {
-				p.conns = append(p.conns[:idx], p.conns[idx+1:]...)
-				break
+func (p *gettyRPCClientPool) getConnFromPoll() (*gettyRPCClient, error) {
```
watermelo, line 379: `getConnFromPoll -> Pool`

protocol/dubbo/poolqueue.go
```diff
@@ -0,0 +1,181 @@
+package dubbo
```
watermelo, line 1: This is a inner package in golang, Maybe we can move it to dubbogo/gost.

protocol/dubbo/pool.go
```diff
@@ -288,117 +287,113 @@ func (c *gettyRPCClient) close() error {
...
+		if ok {
+			conn := value.(*gettyRPCClient)
+			if d := now - conn.getActive(); d > p.ttl {
+				go conn.close()
+				continue
 			}
+			conn.updateActive(now)
+			return conn, nil
+		} else {
+			break
```
watermelo, line 395: directly return.

protocol/dubbo/poolqueue.go
wenxuwan: yes, here just for easy review. I will move it to dubbogo/gost after review.

protocol/dubbo/pool.go
wenxuwan: done

protocol/dubbo/pool.go
wenxuwan: done

AlexStocks (approved):

protocol/dubbo/pool.go
```diff
@@ -288,117 +287,112 @@ func (c *gettyRPCClient) close() error {
...
-			if conn == c {
-				p.conns = append(p.conns[:idx], p.conns[idx+1:]...)
-				break
+func (p *gettyRPCClientPool) getConnFromPool() (*gettyRPCClient, error) {
+	now := time.Now().Unix()
+	if p.poolQueue == nil {
+		return nil, errClientPoolClosed
+	}
+	for {
+		value, ok := p.poolQueue.popTail()
```
zouyx, line 385: I have a question, if i have many requests in processing , but call gettyRPCClientPool.close().
What will it happen?

protocol/dubbo/pool.go
```diff
@@ -288,117 +287,112 @@ func (c *gettyRPCClient) close() error {
...
-		if p.conns[i] == conn {
-			return
+	if p.chInitialized == 0 {
+		p.ch = make(chan struct{}, p.maxSize)
+		for i := 0; i < p.maxSize; i++ {
+			p.ch <- struct{}{}
 		}
+		atomic.StoreUint32(&p.chInitialized, 1)
 	}
+	p.Unlock()
```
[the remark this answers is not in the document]
wenxuwan: yes, unlock with defer more safe

zouyx (approved):

watermelo (approved):

protocol/dubbo/client.go
```diff
@@ -267,11 +268,25 @@ func (c *Client) call(ct CallType, request *Request, response *Response, callbac
...
 		if err == nil {
-			c.pool.put(conn)
-			return
+			for {
+				ok := atomic.CompareAndSwapUint32(&c.pool.pushing, 0, 1)
+				if ok {
+					c.pool.poolQueue.pushHead(conn)
+					c.pool.pushing = 0
+					c.pool.ch <- struct{}{}
+					return
```
gaoxinge, line 279: Can these codes be encapsulated in one method, such as `func (p *gettyRPCClientPool) putGettyRpcClient(conn *gettyRPCClient)`?

protocol/dubbo/pool.go
```diff
@@ -288,117 +287,112 @@ func (c *gettyRPCClient) close() error {
...
-	// check whether @conn has existed in p.conns or not.
-	for i := range p.conns {
-		if p.conns[i] == conn {
-			return
+	if p.chInitialized == 0 {
+		p.ch = make(chan struct{}, p.maxSize)
+		for i := 0; i < p.maxSize; i++ {
+			p.ch <- struct{}{}
 		}
+		atomic.StoreUint32(&p.chInitialized, 1)
```
gaoxinge, line 344: Because the update of `p.chInitialized` is protected by a lock, we can remove `atomic.LoadUint32` and `atomic.StoreUint32`.

gaoxinge, line 344: Because the update of `c.chInitialized` is protected by a lock, I think we can remove `atomic.LoadUint32` and `atomic.StoreUint32`.

wenxuwan: atomic operation performance is better than lock.As stated in the comments, lock is slow path, atomic is fast path.

protocol/dubbo/client.go
wenxuwan: This is the only place to push conn, so i think the code logic can be placed directly here

protocol/dubbo/pool.go
gaoxinge: I mean that we can replace  

1 At line 333, `atomic.LoadUint32(&p.chInitialized) == 1` => `p.chInitialized == 1`
2 At line 344, `atomic.StoreUint32(&p.chInitialized, 1)` => `p.chInitialized = 1`

Because the update of `p.chInitialized` is protected by a lock, so `atomic` here is overhead.

protocol/dubbo/client.go
```diff
@@ -267,11 +268,25 @@ func (c *Client) call(ct CallType, request *Request, response *Response, callbac
...
 	defer func() {
+		failNumber := 0
 		if err == nil {
-			c.pool.put(conn)
-			return
+			for {
+				ok := atomic.CompareAndSwapUint32(&c.pool.pushing, 0, 1)
+				if ok {
+					c.pool.poolQueue.pushHead(conn)
+					c.pool.pushing = 0
```
gaoxinge, line 277: Need we use `atomic` here?

protocol/dubbo/pool.go
wenxuwan: I know your meaning, the lock can protect the value can't be changed at the same time, but read operation is under no lock protection, so atomic is needed. You can assume that there are two threads A and B. Thread A acquires the lock and performs the operation p.chInitialized = 1. At this time, B performs the if(p.chInitialized == 1) operation. This place is not protected by a lock, so Atomic operation is necessary, although it will not cause errors without atomic operation, but it will reduce performance

protocol/dubbo/client.go
wenxuwan: yes, pushHead adds val at the head of the queue. It must only be called by a single producer.

protocol/dubbo/pool.go
gaoxinge: Ok. Whatever `atomive` or `lock`, it is better that these code are placed in `sync.Once` ?

wenxuwan: The implementation is same with sync.Once.Do function

gaoxinge (approved):

gaoxinge:
@wenxuwan Nice job, and I approve your pull request. But I still have three confusions:

1. Why you use `cas` + busy waiting instead of a lock to protect the euque of a SPMC queue?

- `cas` + busy waiting:

https://github.com/apache/dubbo-go/blob/e320f3f7411bd50c519be161a241e6a6f7387031/protocol/dubbo/client.go#L273-L285

- lock:

```go
 	lock()
 	c.pool.poolQueue.pushHead(conn) 
 	c.pool.ch <- struct{}{}
 	unlock()
```

---

2. Why you choose the SPMC queue instead of a MPMC queue? 

- You can find a lock-free implementation of MPMC in [smallnest/queue](https://github.com/smallnest/queue/blob/master/lockfree_queue.go).

---

3. Why you use channel + queue instead of only one channel to implement connection pool?

- I think you can put connection into a channel directly, but I'm not sure. Codes may like:

```go
func (p *gettyRPCClientPool) get(protocol, addr string) {
    once.Do(
        func() {
               p.ch = make(chan *gettyRPCClient, p.maxSize)
		for i := 0; i < p.maxSize; i++ {
			p.ch <- nil
		}
        }
    )
    
    conn := <- p.ch
    if conn == nil {
        return newGettyRPCClientConn(p, protocol, addr)
    } 
    return conn
}

func (p *gettyRPCClientPool) put(conn *gettyRPCClient) {
    p.ch <- conn
}
```

wenxuwan:
> @wenxuwan Nice job, and I approve your pull request. But I still have three confusions:
> 
> 1. Why you use `cas` + busy waiting instead of a lock to protect the enque of a SPMC queue?
> 
> * `cas` + busy waiting:
> 
> https://github.com/apache/dubbo-go/blob/e320f3f7411bd50c519be161a241e6a6f7387031/protocol/dubbo/client.go#L273-L285
> 
> * lock:
> 
> ```go
>  	lock()
>  	c.pool.poolQueue.pushHead(conn) 
>  	c.pool.ch <- struct{}{}
>  	unlock()
> ```
> 
> 1. Why you choose the SPMC queue instead of MPMC queue?
> 
> * You can find a lock-free implementation of MPMC in [smallnest/queue](https://github.com/smallnest/queue/blob/master/lockfree_queue.go).
> 
> 1. Why you use channel + queue instead of only one channel to implement connection pool?
> 
> * I think you can put connection into a channel directly, but I'm not sure. Codes may like:
> 
> ```go
> func (p *gettyRPCClientPool) get(protocol, addr string) {
>     once.Do(
>         func() {
>                p.ch = make(chan *gettyRPCClient, p.maxSize)
> 		for i := 0; i < p.maxSize; i++ {
> 			p.ch <- nil
> 		}
>         }
>     )
>     
>     conn := <- p.ch
>     if conn == nil {
>         return newGettyRPCClientConn(p, protocol, addr)
>     } 
>     return conn
> }
> 
> func (p *gettyRPCClientPool) put(conn *gettyRPCClient) {
>     p.ch <- conn
> }
> ```

@gaoxinge ,Thank you for your careful review。

Why you use cas + busy waiting instead of a lock to protect the enque of a SPMC queue?

I just want to use atomic instead lock, maybe there is no big difference in their performance with sleep.


Why you choose the SPMC queue instead of MPMC queue?

I'am not familiar with MPMC queue， i will read the code you provided in the comments. 


Why you use channel + queue instead of only one channel to implement connection pool?

This is a good question. In my opinion, use signal channel it is ok, but there is such a scene，for example, maxSize is three, When the three tasks come， will create three new conns. Then task1 and task2 failed, task3  success. Now the value in chan should be nil, nil, conn. when new task come, it will create new connections although there have available conn. So in my code ,the queue used for store available conn, channel used to control the number of concurrent tasks.

gaoxinge:
@wenxuwan Thank you for your answers. You have designd the connetion pool carefully and throughly, and I have no other questions.

zouyx (commented):
What about now? If no more questions, i will merge it to 1.4 and release it.

fangyincheng (approved):

PR #800, opened 2020-10-22
enhance client's connectivity

beiwei30:
**What this PR does**:

when the server is down by accident, we need a mechanism to detect this situation quickly, so that the health check router will not route the request to the problematic provider.
 
Fixes #

```release-note

```

### DIFF
--- a/remoting/getty/getty_client.go
+++ b/remoting/getty/getty_client.go
@@ -204,12 +204,10 @@ func (c *Client) Request(request *remoting.Request, timeout time.Duration, respo
 	return perrors.WithStack(err)
 }
 
+// isAvailable returns true if the connection is available, or it can be re-established.
 func (c *Client) IsAvailable() bool {
-	client, err := c.pool.get()
-	if client == nil || !client.isAvailable() || err != nil {
-		return false
-	}
-	return true
+	_, _, err := c.selectSession(c.addr)
+	return err == nil
 }
 
 func (c *Client) selectSession(addr string) (*gettyRPCClient, getty.Session, error) {
--- a/remoting/getty/getty_client.go
+++ b/remoting/getty/getty_client.go
@@ -204,6 +204,14 @@ func (c *Client) Request(request *remoting.Request, timeout time.Duration, respo
 	return perrors.WithStack(err)
 }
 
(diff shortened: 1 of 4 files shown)


remoting/getty/getty_client.go
```diff
@@ -204,6 +204,14 @@ func (c *Client) Request(request *remoting.Request, timeout time.Duration, respo
 	return perrors.WithStack(err)
 }
 
+func (c *Client) IsAvailable() bool {
+	client, err := c.pool.get()
+	if client == nil || !client.isAvailable() || err != nil {
+		return false
+	}
+	return true
```
zouyx, line 212:
```suggestion
	return  client != nil && client.isAvailable() && err == nil 
```

beiwei30: I have changed the implementation to 

```go
// isAvailable returns true if the connection is available, or it can be re-established.
func (c *Client) IsAvailable() bool {
	_, _, err := c.selectSession(c.addr)
	return err == nil
}
```

since we need to re-establish the connection if possible.

AlexStocks (approved):

wenxuwan:
selectSession function will try to get conn  or create new conn with lock, If we just want to check whether the conn of the gettyRPCClientPool  is healthy, we can add a flag to reduce the use of locks through atomic operations。

wenxuwan:
LGTM

remoting/getty/getty_client.go
```diff
@@ -204,6 +204,12 @@ func (c *Client) Request(request *remoting.Request, timeout time.Duration, respo
 	return perrors.WithStack(err)
 }
 
+// isAvailable returns true if the connection is available, or it can be re-established.
+func (c *Client) IsAvailable() bool {
+	_, _, err := c.selectSession(c.addr)
+	return err == nil
```
zouyx, line 210:
```suggestion
	c, _, err := c.selectSession(c.addr)
	return c!=nil && err == nil
```

zouyx (approved):

remoting/getty/getty_client.go
beiwei30: add this defensive check, pls. check again.

PR #850, opened 2020-11-05
fix bug: pathMap cache need to delete after node removed

cadeeper:
as the title mentioned.

### DIFF
--- a/remoting/zookeeper/listener.go
+++ b/remoting/zookeeper/listener.go
@@ -302,8 +302,8 @@ func (l *ZkEventListener) listenDirEvent(conf *common.URL, zkPath string, listen
 				if l.listenServiceNodeEvent(zkPath) {
 					listener.DataChange(remoting.Event{Path: zkPath, Action: remoting.EventTypeDel})
 					l.pathMapLock.Lock()
+					defer l.pathMapLock.Unlock()
 					delete(l.pathMap, zkPath)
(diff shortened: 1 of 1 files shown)


zouyx (commented):
What will happen if it is not deleted?

cadeeper:
> What will happen if it is not deleted?

node add event won't trigger when the same provider node unregister and regiester again. it will continue on : @zkPath %s has already been listened.

remoting/zookeeper/listener.go
```diff
@@ -301,6 +301,9 @@ func (l *ZkEventListener) listenDirEvent(conf *common.URL, zkPath string, listen
 			go func(zkPath string, listener remoting.DataListener) {
 				if l.listenServiceNodeEvent(zkPath) {
 					listener.DataChange(remoting.Event{Path: zkPath, Action: remoting.EventTypeDel})
+					l.pathMapLock.Lock()
+					delete(l.pathMap, zkPath)
+					l.pathMapLock.Unlock()
```
AlexStocks, line 306: plg using defer here.

cadeeper: done

AlexStocks (approved):

gaoxinge (approved):

cityiron (approved):

fangyincheng (approved):

PR #857, opened 2020-11-09
Fix Warnings: incorrect conversion between integer type

### LINKED ISSUE #839: Security:Incorrect conversion between integer types
**What happened**:

Fix the security problems as below:

image

image

kedadiannao220:
**What this PR does**:

 it is just a warning about [CWE-681](https://github.com/github/codeql-go/blob/main/ql/src/Security/CWE-681/IncorrectIntegerConversion.ql)

Fixes #839

```release-note
None
```

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -487,7 +487,7 @@ func (c URL) GetParamBool(key string, d bool) bool {
 // GetParamInt gets int64 value by @key
 func (c URL) GetParamInt(key string, d int64) int64 {
 	r, err := strconv.ParseInt(c.GetParam(key, ""), 10, 64)
-	if r == 0 || err != nil {
+	if  err != nil {
 		return d
 	}
 	return r
@@ -496,7 +496,7 @@ func (c URL) GetParamInt(key string, d int64) int64 {
 // GetParamInt32 gets int32 value by @key
 func (c URL) GetParamInt32(key string, d int32) int32 {
 	r, err := strconv.ParseInt(c.GetParam(key, ""), 10, 32)
-	if r == 0 || err != nil {
+	if  err != nil {
 		return d
 	}
 	return int32(r)
@@ -505,7 +505,7 @@ func (c URL) GetParamInt32(key string, d int32) int32 {
 // GetParamByIntValue gets int value by @key
 func (c URL) GetParamByIntValue(key string, d int) int {
 	r, err := strconv.ParseInt(c.GetParam(key, ""), 10, 0)
-	if r == 0 || err != nil {
+	if  err != nil {
 		return d
 	}
 	return int(r)
@@ -514,7 +514,7 @@ func (c URL) GetParamByIntValue(key string, d int) int {
 // GetMethodParamInt gets int method param
 func (c URL) GetMethodParamInt(method string, key string, d int64) int64 {
 	r, err := strconv.ParseInt(c.GetParam("methods."+method+"."+key, ""), 10, 64)
-	if r == 0 || err != nil {
+	if err != nil {
 		return d
 	}
 	return r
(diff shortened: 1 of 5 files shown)


common/url.go
```diff
@@ -484,22 +484,49 @@ func (c URL) GetParamBool(key string, d bool) bool {
...
 		return d
 	}
-	return int64(r)
+	return r
+}
+
+// GetParamInt32 gets int32 value by @key
+func (c URL) GetParamInt32(key string, d int32) int32 {
+	r, err := strconv.ParseInt(c.GetParam(key, ""), 10, 32)
+	if r == 0 || err != nil {
```
wongoo, line 499: if `key=0`, then return default value. is it the right logic?

common/url.go
```diff
@@ -484,22 +484,49 @@ func (c URL) GetParamBool(key string, d bool) bool {
...
+		return d
+	}
+	return int32(r)
+}
+
+// GetParamByInt0 gets int value by @key
+func (c URL) GetParamByInt0(key string, d int) int {
```
wongoo, line 506: is it better to rename it to `GetParamByIntValue` ?

common/url.go
```diff
@@ -484,22 +484,49 @@ func (c URL) GetParamBool(key string, d bool) bool {
...
-	r, err := strconv.Atoi(c.GetParam("methods."+method+"."+key, ""))
+	r, err := strconv.ParseInt(c.GetParam("methods."+method+"."+key, ""), 10, 64)
+	if r == 0 || err != nil {
+		return d
+	}
+	return r
+}
+
+// GetMethodParamInt0 gets int method param
+func (c URL) GetMethodParamInt0(method string, key string, d int) int {
```
wongoo, line 524: is it better to rename it to `GetMethodParamIntValue` ?

common/url.go
```diff
@@ -484,22 +484,49 @@ func (c URL) GetParamBool(key string, d bool) bool {
...
+	if r == 0 || err != nil {
+		return d
+	}
+	return int32(r)
+}
+
+// GetParamByInt0 gets int value by @key
+func (c URL) GetParamByInt0(key string, d int) int {
+	r, err := strconv.ParseInt(c.GetParam(key, ""), 10, 0)
+	if r == 0 || err != nil {
```
zouyx, line 508: as wongoo's comment

common/url.go
```diff
@@ -484,22 +484,49 @@ func (c URL) GetParamBool(key string, d bool) bool {
...
+		return d
+	}
+	return int(r)
 }
 
 // GetMethodParamInt gets int method param
 func (c URL) GetMethodParamInt(method string, key string, d int64) int64 {
-	r, err := strconv.Atoi(c.GetParam("methods."+method+"."+key, ""))
+	r, err := strconv.ParseInt(c.GetParam("methods."+method+"."+key, ""), 10, 64)
+	if r == 0 || err != nil {
```
zouyx, line 517: as wongoo's comment

common/url.go
kedadiannao220: `key` is only a flag of `param` value;  this conversion  is about value of  key, not key itself;

e.g.
```
func TestURLGetParamInt(t *testing.T) {
	params := url.Values{}
	params.Set("key", "3")
	u := URL{baseUrl: baseUrl{params: params}}
	v := u.GetParamInt("key", 1)
	assert.Equal(t, int64(3), v)

	u = URL{}
	v = u.GetParamInt("key", 1)
	assert.Equal(t, int64(1), v)
}
``

common/url.go
kedadiannao220: ok , i will change it;

common/url.go
wongoo: maybe u miss-understand what I mean,  if `params.Set("key", "0")` ,  the value of `u.GetParamInt("key", 1)` will be 1, which is not correct.

kedadiannao220: wow, soga;  you are right.....  I think I am fixing a bug  of this module.

kedadiannao220: I searched the history commits,  `r==0` is  avoid to situation of  `param.Set("key", "")`

wongoo (approved):

zouyx (commented):
Change target branch to develop pls

AlexStocks (approved):

watermelo:
Merged at pr #871

PR #874, opened 2020-11-12
Fix: client read timeout caused by default request timeout is 0

peaman:
**What this PR does**:
修复默认request timeout为0导致的错误(client read timeout)

None

如果用户未设置timeout值，在NewDubboInvoker()方法中做了兜底
此修改可以解决问题，但未必是最好的修复方式

None

```release-note

```

### DIFF
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -73,6 +73,9 @@ func NewDubboInvoker(url common.URL, client *remoting.ExchangeClient) *DubboInvo
 	if t, err := time.ParseDuration(requestTimeoutStr); err == nil {
 		requestTimeout = t
 	}
+	if requestTimeout == 0 {
+		requestTimeout = time.Second * 3
+	}
 	return &DubboInvoker{
 		BaseInvoker: *protocol.NewBaseInvoker(url),
 		client:      client,


AlexStocks (approved):

protocol/dubbo/dubbo_invoker.go
```diff
@@ -73,6 +73,9 @@ func NewDubboInvoker(url common.URL, client *remoting.ExchangeClient) *DubboInvo
 	if t, err := time.ParseDuration(requestTimeoutStr); err == nil {
```
cityiron, line 73: I think the code should be 

```go
	if t, err := time.ParseDuration(requestTimeoutStr); err == nil {
		requestTimeout = time.Second * 3
	}
```

cityiron (changes requested):

protocol/dubbo/dubbo_invoker.go
```diff
@@ -73,6 +73,9 @@ func NewDubboInvoker(url common.URL, client *remoting.ExchangeClient) *DubboInvo
...
 		requestTimeout = t
 	}
+	if requestTimeout == 0 {
+		requestTimeout = time.Second * 3
```
fangyincheng, line 77: url.GetParam(key, default)这个方法就可以设置默认值了

protocol/dubbo/dubbo_invoker.go
peaman: it is because  `time.ParseDuration(requestTimeoutStr); err ==nil` is false, so requestTimeout is 0
I think it not solve the problem

protocol/dubbo/dubbo_invoker.go
zouyx: I think this PR is set default value ,rather than how to set default value?

zouyx (commented):
I think you should check: why [this default value](https://github.com/apache/dubbo-go/blob/ab44015ec19d72ca524bc96a77b68e60dc8c8178/config/consumer_config.go#L56) has not set .

protocol/dubbo/dubbo_invoker.go
peaman: I noticed this method，How do you think I should modify it ？

zouyx:
You should need to use code as below for make all variable has default value before set custom value.
```golang
consumerConfig := &config.ConsumerConfig{}
yaml.UnmarshalYML([]byte("{}"), consumerConfig)
```

PR #773, opened 2020-09-22
Imp: replace go/net/url

christophe-liwei:
根据[#769](https://github.com/apache/dubbo-go/pull/769)的讨论，替换`go/net/url`为`common/net/url`。

```
// Set sets the key to value.
// If value is not "", it will replaces existing values.
func (v Values) Set(key, value string) {
	if len(value) == 0 {
		return
	}
	v[key] = []string{value}
}

// Add adds the value to key. If value is not "", it appends to existing
// values associated with key.
func (v Values) Add(key, value string) {
	if len(value) == 0 {
		return
	}
	v[key] = append(v[key], value)
}
```

Values的Set方法替换为value 是非空字符串时才能替换old value

PR #876, opened 2020-11-14
Enhance: upgrade dubbogo URL

georgehao:
**What this PR does**:
Enhance dubbogo URL

I have read some discuss [How to copy a struct which contains a mutex](https://groups.google.com/g/golang-nuts/c/imxjBLNJ9OY). The discuss tell us can't copy a struct which contains `sync` type or variable. 

And the struct contains `sync` type can see anywhere of source code of k8s, but k8s use the struct with pointer. eg: [Cache](https://github.com/kubernetes/kubernetes/blob/d20e3246bade17564860655e7fa57ead922d7307/pkg/controller/endpointslicemirroring/metrics/cache.go#L37)

I have two way to fix the URL.

#### **way1** : move the `paramsLock  sync.RWMutex` to `paramsLock   *sync.RWMutex`

like this:
```
type URL struct {
        // ....
        Path     string
        paramsLock   *sync.RWMutex
}
```

it has some problem:
1. when new URL, we must init the `paramsLock`, otherwise may call the nil pointer panic.
2. some struct may share the same `paramsLock`, will influence each other. 

so it's a bad solution.

#### **way2**: make `common.URL` noCopy. must use the `*common.URL` instead of the `common.URL`. 

this way is the general way. so i select this

I also add `noCopy` struct to `common.URL`, it's will warning us not copy the URL struct. 

#### other
my solution is  similar to the PR #773, but #773 missing some files not modify.
 
I have add the feature of #823 do.

### DIFF
--- a/common/proxy/proxy_factory/default.go
+++ b/common/proxy/proxy_factory/default.go
@@ -72,7 +72,7 @@ func (factory *DefaultProxyFactory) GetAsyncProxy(invoker protocol.Invoker, call
 }
 
 // GetInvoker gets a invoker
-func (factory *DefaultProxyFactory) GetInvoker(url common.URL) protocol.Invoker {
+func (factory *DefaultProxyFactory) GetInvoker(url *common.URL) protocol.Invoker {
 	return &ProxyInvoker{
 		BaseInvoker: *protocol.NewBaseInvoker(url),
 	}
@@ -90,7 +90,7 @@ func (pi *ProxyInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
 
 	url := pi.GetUrl()
 	//get providerUrl. The origin url may be is registry URL.
-	url = *getProviderURL(&url)
+	url = getProviderURL(url)
 
 	methodName := invocation.MethodName()
 	proto := url.Protocol
--- a/common/proxy/proxy_factory/default.go
+++ b/common/proxy/proxy_factory/default.go
@@ -88,9 +88,8 @@ func (pi *ProxyInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
 	result := &protocol.RPCResult{}
 	result.SetAttachments(invocation.Attachments())
 
-	url := pi.GetUrl()
 	//get providerUrl. The origin url may be is registry URL.
-	url = getProviderURL(url)
+	url := getProviderURL(pi.GetUrl())
 
 	methodName := invocation.MethodName()
 	proto := url.Protocol
--- a/config/config_loader_test.go
+++ b/config/config_loader_test.go
@@ -25,7 +25,6 @@ import (
 )
 
 import (
-	cm "github.com/Workiva/go-datastructures/common"
 	"github.com/Workiva/go-datastructures/slice/skip"
 	gxset "github.com/dubbogo/gost/container/set"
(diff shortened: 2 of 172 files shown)


zouyx (commented):
Thanks @georgehao , make `Checks`normal pls.
BTW can i close #773 ?

georgehao:
> Thanks @georgehao , make `Checks`normal pls.
> BTW can i close #773 ?

yes, you can close 773.

common/proxy/proxy_factory/default.go
```diff
@@ -90,7 +90,7 @@ func (pi *ProxyInvoker) Invoke(ctx context.Context, invocation protocol.Invocati
 
 	url := pi.GetUrl()
 	//get providerUrl. The origin url may be is registry URL.
-	url = *getProviderURL(&url)
+	url = getProviderURL(url)
```
cvictory, line 93: 两行可以合并吧

go.mod
```diff
@@ -51,6 +51,7 @@ require (
 
 replace (
 	github.com/envoyproxy/go-control-plane => github.com/envoyproxy/go-control-plane v0.8.0
+	github.com/shirou/gopsutil => github.com/shirou/gopsutil v0.0.0-20181107111621-48177ef5f880
```
cvictory, line 54: 为什么指定版本？ 而且是通过replace方式？

config/config_loader_test.go
```diff
@@ -418,23 +417,23 @@ func (mts *mockMetadataService) addURL(targetMap *sync.Map, url *common.URL) boo
...
-func (m *mockMetadataService) getAllService(services *sync.Map) []common.URL {
+func (m *mockMetadataService) getAllService(services *sync.Map) []*common.URL {
 	// using skip list to dedup and sorting
-	res := make([]common.URL, 0)
+	var res []*common.URL
 	services.Range(func(key, value interface{}) bool {
 		urls := value.(*skip.SkipList)
 		for i := uint64(0); i < urls.Len(); i++ {
-			url := common.URL(urls.ByPosition(i).(Comparator))
+			url := urls.ByPosition(i).(*common.URL)
```
cvictory, line 436: 转换确认下，没有问题。

go.mod
georgehao: @zouyx 贤哥搞的, 贤哥看下吧

config/config_loader_test.go
georgehao: 嗯, 我用URL实现了Comparator, 把config_loader里面实现的Comparator给去掉了

common/proxy/proxy_factory/default.go
georgehao: OK

fangyincheng:
> > Thanks @georgehao , make `Checks`normal pls.
> > BTW can i close #773 ?
> 
> yes, you can close 773.

what about 823?

AlexStocks (approved):

go.mod
AlexStocks: https://github.com/alibaba/sentinel-golang/issues/326

hxmhlt (approved):

fangyincheng (approved):

PR #889, opened 2020-11-23
Fix:  refact heartbeat

fangyincheng:
* client send heartbeat
          now: close session when once timeout
          doing: close session when three times timeout（60s*3） 
* server send heartbeat
          now: print error when error response 
          doing: close session when three times timeout（60s*3）


remoting/getty/config.go
```diff
@@ -188,6 +196,12 @@ func (c *ClientConfig) CheckValidity() error {
 			c.HeartbeatPeriod, time.Duration(config.MaxWheelTimeSpan))
 	}
 
+	if len(c.HeartbeatTimeout) == 0 {
+		c.heartbeatTimeout = 5 * time.Second
```
AlexStocks, line 200: The default hb timeout duration 5s is too short. Pls change it to 90s.

remoting/getty/config.go
```diff
@@ -199,6 +213,17 @@ func (c *ClientConfig) CheckValidity() error {
 func (c *ServerConfig) CheckValidity() error {
 	var err error
 
+	if len(c.HeartbeatPeriod) == 0 {
+		c.heartbeatPeriod = 15 * time.Second
```
AlexStocks, line 217: ur default timeout is 5s while hb period is 15s. very strange.

remoting/getty/config.go
fangyincheng: that is a test data， I will change it to 60s same as `Dubbo`

remoting/getty/config.go
fangyincheng: above

hxmhlt (approved):

zouyx (approved):

AlexStocks (approved):

PR #892, opened 2020-11-25
Ftr:support for loading whole ServiceConfig&ReferenceConfig propertie…

xuchenncepu:


xuchenncepu:
prefix func appears in many other config struct.

PR #994, opened 2021-01-12
fixed invalid timeout config setting

yxcodebase:


zouyx (commented):
why do you disable this code?

AlexStocks:
@yxcodebase pls tell us why u remark "os.Exit(0)"?

yxcodebase:
Exit immediately regardless of whether the timeout is set。

cityiron:
Do not understand, can talk in detail, are you in the group of dingding 31363295

PR #1022, opened 2021-01-26
Fix: consul security problem

zouyx:
**What this PR does**:

* Fix consul security problem #1004 
* Delete  `replace github.com/envoyproxy/go-control-plane` in go.mod , it causes incompatible problem in consul v1.8.7.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #1029, opened 2021-01-31
Revert "Fix: consul security problem"

zouyx:
Reverts apache/dubbo-go#1022

checking fail, revert pls

Can not update consul sdk, it will cause checking fail.

Root problem is as below:

```
# github.com/hashicorp/consul/sdk/testutil
../../../../pkg/mod/github.com/hashicorp/consul/sdk@v0.6.0/testutil/server.go:239:28: cannot use t (type "testing".TB) as type CleanupT in argument to defaultServerConfig:
	"testing".TB does not implement CleanupT (missing Cleanup method)
../../../../pkg/mod/github.com/hashicorp/consul/sdk@v0.6.0/testutil/testlog.go:14:41: cannot use t (type "testing".TB) as type CleanupT in argument to NewLogBuffer:
```

Detail:https://travis-ci.com/github/apache/dubbo-go/builds/215403185

It need to update  golang version for fix this problem, detail as below:

https://github.com/hashicorp/consul/issues/8538.


fangyincheng (approved):

AlexStocks (approved):

zouyx:
Suggest update golang version in dubbo-go 3.0 rather than now.

cityiron:
> Suggest update golang version in dubbo-go 3.0 rather than now.

I think your suggest is right. @LaurenceLiZhixin @AlexStocks What do you think?

PR #1035, opened 2021-02-01
add read configuration from the command line when start

cjphaha:
**What this PR does**:
Add feature :
Developers can add command line parameters when starting the program to set the location of log, provider, consumer and router configuration files.
If no parameters are added at startup, it will be read from the default environment variables **same as before**.
If you set environment variables and add parameters at startup, it will run according to the startup parameters.

Fixes #

+ In https://github.com/dubbogo/dubbo-samples/dubbo-samples/golang/helloworld/dubbo/go-server/app
   You can start server width out setting environment variables, like this:
   ```shell
   go run . -proConf ../profiles/dev/server.yml -logConf ../profiles/dev/log.yml
   ```
   In https://github.com/dubbogo/dubbo-samples/dubbo-samples/golang/helloworld/dubbo/go-client/app
   You can start client like this:
   ```shell
   go run . -conConf ../profiles/dev/client.yml -logConf ../profiles/dev/log.yml
   ```
+ If you don’t want to start in this way, you can set the environment variables in go-server/app like this:
   ```shell
   export CONF_PROVIDER_FILE_PATH="../profiles/dev/server.yml"
   export APP_LOG_CONF_FILE="../profiles/dev/log.yml"
   ```
   then start the server
   ```shell
   go run .
   ```
+ if you set environment variables and parameters like this:
   ```shell
   export CONF_PROVIDER_FILE_PATH="../profiles/dev/server.yml"
   export APP_LOG_CONF_FILE="../profiles/dev/log.yml"
   go run . -conConf ../profiles/dev/client_new.yml
   ```
   under this circumstance “../profiles/dev/client_new.yml” is valid

```release-note

```

### DIFF
--- a/common/logger/logger.go
+++ b/common/logger/logger.go
@@ -64,7 +64,7 @@ func init() {
 	fs := flag.NewFlagSet("log_config", flag.ContinueOnError)
 	logConfFile := fs.String("logConf", "", "default log config path")
 	fs.Parse(os.Args[1:])
-	for len(fs.Args()) != 0{
+	for len(fs.Args()) != 0 {
 		fs.Parse(fs.Args()[1:])
 	}
 	err := InitLog(*logConfFile)
--- a/common/logger/logger.go
+++ b/common/logger/logger.go
@@ -18,6 +18,7 @@
 package logger
 
 import (
+	"flag"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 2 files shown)


Patrick0308 (commented):
I think we should support env parameter to config that we did before too. This change will be 
 backward compatible.

cjphaha:
> I think we should support env parameter to config that we did before too. This change will be
> backward compatible.

Yes, this is my problem. I didn’t think it completely. I will revise it later, thank you!

AlexStocks:
Pls change the target branch from develop to 1.5.

PR #1073, opened 2021-03-04
Ftr: provide a default config

cjphaha:
**What this PR does**:
If the path of the configuration file cannot be obtained from the environment variables or the startup parameters, the system will search for the configuration file from the default location. If the configuration file fails to read, the default parameters will be set

Fixes #

+ Put the config files in profiles folder. by default, the configuration file should be placed like this
```shell
.
├── app
│   ├── client.go
│   └── user.go
└── profiles
   ├── client.yml
   └── log.yml
```
+ You can start the program widthout setting conConf and proConf:
```shell
go run .
```
+ If there is no config files in peofiles path, this pr will provide a default config like  https://github.com/dubbogo/dubbo-samples/tree/master/golang/helloworld/dubbo


flycash (approved):

cjphaha:
> Could you make checking success?

Sorry, Recently I encountered some difficulties, Now I have changed the way of setting default parameters. It can be passed on my local computer.

AlexStocks (approved):

config/config_loader.go
```diff
@@ -86,12 +100,48 @@ func init() {
...
+		// if so, we set a default value for it
+		setDefaultValue(providerConfig)
 		// Even though baseConfig has been initialized, we override it
 		// because we think read from config file is correct config
 		baseConfig = &providerConfig.BaseConfig
 	}
 }
 
+// setDefaultValue set default value for providerConfig or consumerConfig if it is null
+func setDefaultValue(target interface{}) {
```
zouyx, line 113: Can you use annotation in [this line](https://github.com/apache/dubbo-go/blob/274718b650e71ee55a832203e87088f9762d4ff7/config/consumer_config.go#L53) for default value?

AlexStocks: > Can you use annotation in [this line](https://github.com/apache/dubbo-go/blob/274718b650e71ee55a832203e87088f9762d4ff7/config/consumer_config.go#L53) for default value?

in go we can use annotation as Java?

flycash: let me see see and study study. hhhh

LaurenceLiZhixin (approved):

Issue #993, opened 2021-01-11
Get weird error message: `pkg.Unmarshal(len(@data):0) = error:get wrong attachments: <nil>` in the server


spacewander:
**What happened**:
Get weird error message: `pkg.Unmarshal(len(@data):0) = error:get wrong attachments: <nil>` in the server code.

**What you expected to happen**:
Server works as expected.

**How to reproduce it (as minimally and precisely as possible)**:
1. use tengine as http2dubbo proxy, with configuration in the example: https://github.com/alibaba/tengine/tree/master/modules/mod_dubbo
2. write a dubbo server with dubbo-go
3. request with `curl http://127.0.0.1:8080/dubbo -i`

**Anything else we need to know?**:
The same thing doesn't happen with the tengine's Java example. It works as the document states.

AlexStocks:
sorry, we have not tested our dubbogo with tengine. maybe u can try to fix this problem and submit a pr. thx.

spacewander:
Hi, @AlexStocks, any updates?

rockXiaofeng:
@spacewander Hi, just now we used dubbo-samples (helloworld/go-server) to debug the program, and obtained the following information in the apisix error_log:
```
2021/03/09 10:36:45 [error] 2080#23919: *527208 dubbo: parse exception failed unsupported type readObjectDefinition
2021/03/09 10:36:45 [warn] 2080#23919: *527208 dubbo: response decode result failed �Ccom.ikurento.user.User�idnameagetime`A001
                                                                                                                               Alex Stocks�Jx�N
2021/03/09 10:36:45 [warn] 2080#23919: *527208 multi: parse get error 00007F9A50025E50
```

rockXiaofeng:
What is the version of apisix dubbo plugin?

spacewander:
Do you mean the dubbo protocol version? It is "2.0.2".

rockXiaofeng:
Yes,  and pls take a look at the above error log.

rockXiaofeng:
Hi @spacewander , as you said, it does not support object,  and just now we changed the provider response from object to string,  called twice with curl, and then we got an error from apisix error_log:
```
2021/03/10 15:10:17 [warn] 2391#27190: *2669171 multi: multi connection detach not empty 00007FB4FAC3F950
[...]

PR #1087, opened 2021-03-16
Bugfix: nil attachments

rockXiaofeng:
**What this PR does**:
 Give it an initial value when attachments is nil

Fixes https://github.com/apache/dubbo-go/issues/993

```release-note

```

### DIFF
--- a/protocol/dubbo/impl/hessian.go
+++ b/protocol/dubbo/impl/hessian.go
@@ -254,6 +254,10 @@ func unmarshalRequestBody(body []byte, p *DubboPackage) error {
 		return perrors.WithStack(err)
 	}
 
+	if attachments == nil {
+		attachments = map[interface{}]interface{}{constant.INTERFACE_KEY: target}
+	}
+
 	if v, ok := attachments.(map[interface{}]interface{}); ok {
 		v[DUBBO_VERSION_KEY] = dubboVersion
 		req[6] = ToMapStringInterface(v)


AlexStocks:
@pantianying @fangyincheng pls check this pr.

pantianying:
I don't think this is a bug of dubbo-go. If there is no upstream attachment, why initialize one? But I also see no problem with your PR code if it improves protocol compatibility. However, we need to confirm the compatibility with the Java side

pantianying (commented):
I don't think this is a bug of dubbo-go. If there is no upstream attachment, why initialize one? But I also see no problem with your PR code if it improves protocol compatibility. However, we need to confirm the compatibility with the Java side

rockXiaofeng:
> I don't think this is a bug of dubbo-go. If there is no upstream attachment, why initialize one? But I also see no problem with your PR code if it improves protocol compatibility. However, we need to confirm the compatibility with the Java side

OK, I agree with you, but according to the issuer:
```
"Anything else we need to know?:
The same thing doesn't happen with the tengine's Java example. It works as the document states."
```
So, it should be confirmed whether java-server is compatible with this situation.

rockXiaofeng:
I encountered this problem when I used the dubbo plugin of apisix as proxy, and dubbo-go as server provider,  so I submitted this PR and the problem was solved in my side.

rockXiaofeng:
FYI (from com.alibaba/dubbo) : 
```
public class DecodeableRpcInvocation extends RpcInvocation implements Codec, Decodeable {
...
@Override
    public Object decode(Channel channel, InputStream input) throws IOException {
        ObjectInput in = CodecSupport.getSerialization(channel.getUrl(), serializationType)
                .deserialize(channel.getUrl(), input);
 
        String dubboVersion = in.readUTF();
        request.setVersion(dubboVersion);
        setAttachment(Constants.DUBBO_VERSION_KEY, dubboVersion);
 
        setAttachment(Constants.PATH_KEY, in.readUTF());
        setAttachment(Constants.VERSION_KEY, in.readUTF());
 ...
  Map<string, string=""> map = (Map<string, string="">) in.readObject(Map.class);
            if (map != null && map.size() > 0) {
                Map<string, string=""> attachment = getAttachments();
                if (attachment == null) {
                    attachment = new HashMap<string, string="">();
                }
                attachment.putAll(map);
                setAttachments(attachment);
            }
            //decode argument ,may be callback
            for (int i = 0; i < args.length; i++) {
                args[i] = decodeInvocationArgument(channel, this, pts, i, args[i]);
            }
 
            setArguments(args);
 
```

AlexStocks (approved):

fangyincheng (approved):

cityiron (approved):

PR #1137, opened 2021-04-07
Fix: when provider restart,  somtimes consumer  can not discover the new service  

jack15083:
**What this PR does**:
fix when provider restart,  consumer automatic can not discover service when somtimes 

Fixes #

```release-note

```


remoting/zookeeper/listener.go
```diff
@@ -182,7 +182,7 @@ func (l *ZkEventListener) handleZkNodeEvent(zkPath string, children []string, li
 				logger.Warnf("delete zkNode{%s}", node)
 				listener.DataChange(remoting.Event{Path: node, Action: remoting.EventTypeDel})
 				l.pathMapLock.Lock()
-				delete(l.pathMap, zkPath)
+				delete(l.pathMap, node)
```
wenxuwan, line 185: In my opinion ,here and below don't need delete logic. There have a goroutine do this(Line 310), please check it.

jack15083: 细看了下逻辑这里确实不需要删除

PR #1181, opened 2021-04-27
Ftr: enable filter and cluster when client consumer provider directly

kzhan:
客户端直连模式下启用 filter 和 cluster处理


config/reference_config.go
```diff
@@ -20,6 +20,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"github.com/apache/dubbo-go/protocol/protocolwrapper"
```
AlexStocks, line 23: move it to the 3rd import block

kzhan: > move it to the 3rd import block

ok

LaurenceLiZhixin (approved):

gaoxinge (approved):

AlexStocks (approved):

PR #1200, opened 2021-05-15
remove reflect in grpc server

gaoxinge:
**What this PR does**:

- remove reflect for `SetProxyImpl` in grpc server
- update protobuf

```release-note
NONE
```

### DIFF
--- a/protocol/grpc/server.go
+++ b/protocol/grpc/server.go
@@ -20,7 +20,6 @@ package grpc
 import (
 	"fmt"
 	"net"
-	"reflect"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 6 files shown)


zouyx (approved):

AlexStocks (approved):

protocol/grpc/server.go
```diff
@@ -134,18 +136,12 @@ func waitGrpcExporter(providerServices map[string]*config.ServiceConfig) {
 func registerService(providerServices map[string]*config.ServiceConfig, server *grpc.Server) {
 	for key, providerService := range providerServices {
 		service := config.GetProviderService(key)
-
 		ds, ok := service.(DubboGrpcService)
 		if !ok {
```
hxmhlt, line 140: Is 'ds' may be nil ?

hxmhlt (commented):
LGTM

protocol/grpc/server.go
hxmhlt: > Is 'service' may be nil ?

gaoxinge: `service` check exist: https://github.com/apache/dubbo-go/blob/master/config/config_loader.go#L259-L264.

PR #1203, opened 2021-05-15
improve etcd version and change create to put

ztelur:
**What this PR does**:
- improve etcd version
- change Create(k,v) function to Put(k,v) for etcd client, because Create(k,v) only succeed when key is not set

Fixes #
- dubbo-go-samples registry etcd server can't start twice or can't start when etcd key has been set

Notice!:
not remove because github.com/hashicorp/consul depend on it
```
google.golang.org/grpc => google.golang.org/grpc v1.29.1
```


AlexStocks (approved):

LaurenceLiZhixin (approved):

gaoxinge (approved):

gaoxinge (commented):
remove https://github.com/apache/dubbo-go/blob/b343b9a102ad3bf454020e2307208bd5ad986c84/go.mod#L70

PR #1254, opened 2021-06-11
Fix: Graceful shutdown bugs

xavier-niu:
**What this PR does**:

- Cancel calling exporter.Unexport() in registryProtocol.Destory()
- Close zkListener before closing zkClient
- Fix timeout logic for graceful shutdown
- Fix bug that RequestsFinished always be true
- Fix bug that RejectRequest not works

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1257, opened 2021-06-13
Fix: Graceful shutdown bugs(supplement #1254)

xavier-niu:
**What this PR does**:

- Fix a bug that RejectRequest is not working
- Fix a bug that shutdownConfig always be nil, because `init()` is called before `config.Load()`. The changes are listed at below:
    -  To avoid the import circle that happens when `filter_impl` package is imported in `config` package, a new interface, called `config.Setter`, is introduced.
    - `gracefulShutdownFilter`s for both provider and consumer are initialized in `init()` as usual, they, however, are not to get `shutdownConfig` immediately, instead, the `shutdownConfig` is set to filters in `config.GracefulShutdownInit()` via `config.Setter`.

see also #1254

Fixes #

```release-note
NONE
```


config/config_loader_test.go
```diff
@@ -18,6 +18,9 @@
 package config
 
 import (
+	"context"
+	"dubbo.apache.org/dubbo-go/v3/filter"
```
AlexStocks, line 22: move it to the 3rd import block.

filter/filter_impl/graceful_shutdown_filter.go
```diff
@@ -71,11 +65,17 @@ func (gf *gracefulShutdownFilter) OnResponse(ctx context.Context, result protoco
...
 	// although this isn't thread safe, it won't be a problem if the gf.rejectNewRequest() is true.
 	if gf.shutdownConfig != nil && gf.activeCount <= 0 {
-		gf.shutdownConfig.RequestsFinished = true
+		gf.shutdownConfig.RequestsFinished.Store(true)
 	}
 	return result
 }
 
+func (gf *gracefulShutdownFilter) Set(name string, conf interface{}) {
+	if shutdownConfig, ok := conf.(*config.ShutdownConfig); ok && name == config.GracefulShutdownFilterShutdownConfig {
```
AlexStocks, line 74: hey, guy, this code line is so long. would u pls split it to two code lines?

AlexStocks (approved):

config/graceful_shutdown_config.go
```diff
@@ -54,7 +58,7 @@ type ShutdownConfig struct {
 	RejectRequest bool
 	// true -> all requests had been processed. In provider side it means that all requests are returned response to clients
 	// In consumer side, it means that all requests getting response from servers
-	RequestsFinished bool
+	RequestsFinished *atomic.Bool
```
win5do, line 61: RequestsFinished is pointer,  can't be fill with config, requires explicit assignment.

It's better to use value types.

filter/filter_impl/graceful_shutdown_filter.go
```diff
@@ -71,11 +65,24 @@ func (gf *gracefulShutdownFilter) OnResponse(ctx context.Context, result protoco
 	atomic.AddInt32(&gf.activeCount, -1)
 	// although this isn't thread safe, it won't be a problem if the gf.rejectNewRequest() is true.
```
win5do, line 66: This comment can be deleted now.

filter/filter_impl/graceful_shutdown_filter.go
```diff
@@ -61,7 +55,7 @@ func (gf *gracefulShutdownFilter) Invoke(ctx context.Context, invoker protocol.I
 	}
 	atomic.AddInt32(&gf.activeCount, 1)
 	if gf.shutdownConfig != nil && gf.activeCount > 0 {
-		gf.shutdownConfig.RequestsFinished = false
+		gf.shutdownConfig.RequestsFinished.Store(false)
```
win5do, line 58: Atomic use LOCK CPU instruction, may cause performance issues.

xavier-niu: IMO, CAS lock is the best choice in high concurrency scenario, because using `sync.Mutex` may block the Goroutine and results in greater performance losses.

config/graceful_shutdown_config.go
xavier-niu: Yes, it is a bug. The latest commit will solve this problem.

xavier-niu:
Currently, it may have performance issue. So, this pr is converted draft temporarily.

AlexStocks (approved):

LaurenceLiZhixin (approved):

gaoxinge (approved):

config/config_setter.go
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package config
+
+const (
+	GracefulShutdownFilterShutdownConfig = "GracefulShutdownFilterShutdownConfig"
+)
+
+type Setter interface {
```
WilliamLeaves, line 24: maybe it is bertter to add some comment for Setting

WilliamLeaves (approved):

config/config_setter.go
xavier-niu: Yes, some comments will be added lately.

PR #1267, opened 2021-06-19
fix hash function bug

tangweihua163:
**What this PR does**:

fix bug

Fixes #

`hash` function shifts a byte by 24/16/8 bits, always get zero.

```release-note

```


AlexStocks:
why do u change the first charactor of word to its low case?

tangweihua163:
> why do u change the first charactor of word to its low case?

like all other three implements of `LoadBalance` interface, `least_active`, `round_robin` and `random`, `consistent_hash` should not export its detail out. this change is not necessary, but reasonable.

the focal point is the bug in `hash` method of type `consistentHashSelector`, left shifting a byte by more than 8 bits always got zero.

AlexStocks:
I have changed the target branch from master to 3.0.

AlexStocks (approved):

LaurenceLiZhixin (approved):

zhaoyunxing92 (approved):
approved these changes

williamfeng323 (approved):

PR #1289, opened 2021-07-01
[#1288] fix the inconsistent filepath key logic

### LINKED ISSUE #1288: Dubbogo Metadata path is inconsistence with Dubbo.
**What happened**:
When generating the file path, it should encode the interface instead of decoding it. Moreover, current logic will cause multiple services to use the same path because when the decode failed, it will return an empty string. No unique path can be guaranteed for services.

dubbo-go version:
image

dubbo version:
image

getFilePathKey function in base_metadata_identifier.go should return strings like "metadata/org.apache.pkg.mockService/1.0.0/Group/provider/a/b/c"

williamfeng323:
**What this PR does**:
Fix the issue mentioned in #1288 

Fixes #1288 

```release-note

```


AlexStocks (approved):

hxmhlt (approved):
LGTM

gaoxinge (approved):

PR #1293, opened 2021-07-02
performance optimization: change time.After => time.NewTimer

mark4z:
**What this PR does**:
Ues time.Timer instead the time.After
```
// After waits for the duration to elapse and then sends the current time
// on the returned channel.
// It is equivalent to NewTimer(d).C.
// The underlying Timer is not recovered by the garbage collector
// until the timer fires. If efficiency is a concern, use NewTimer
// instead and call Timer.Stop if the timer is no longer needed.
func After(d Duration) <-chan Time {
	return NewTimer(d).C
}
```

Fixes #

```release-note

```

### DIFF
--- a/protocol/grpc/server.go
+++ b/protocol/grpc/server.go
@@ -115,7 +115,8 @@ func waitGrpcExporter(providerServices map[string]*config.ServiceConfig) {
 	t := time.NewTicker(50 * time.Millisecond)
 	defer t.Stop()
 	pLen := len(providerServices)
-	ta := time.After(10 * time.Second)
+	ta := time.NewTimer(10 * time.Second)
+	defer ta.Stop()
 
 	for {
 		select {
@@ -124,7 +125,7 @@ func waitGrpcExporter(providerServices map[string]*config.ServiceConfig) {
 			if pLen == mLen {
 				return
 			}
-		case <-ta:
+		case <-ta.C:
 			panic("wait grpc exporter timeout when start grpc server")
 		}
 	}
--- a/tools/cli/client/client.go
+++ b/tools/cli/client/client.go
@@ -143,11 +143,12 @@ func (t *TelnetClient) processSingleRequest(req *protocol.Request, userPkg inter
 	go t.readInputData(string(inputData), requestDataChannel)
 	go t.readServerData(t.conn, responseDataChannel)
 
-	timeAfter := time.After(t.responseTimeout)
+	timeAfter := time.NewTimer(t.responseTimeout)
+	defer timeAfter.Stop()
 
 	for {
 		select {
-		case <-timeAfter:
+		case <-timeAfter.C:
 			log.Println("request timeout to:", t.tcpAddr)
 			return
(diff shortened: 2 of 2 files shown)


mark4z:
@AlexStocks

LaurenceLiZhixin (approved):

protocol/grpc/server.go
```diff
@@ -115,7 +115,8 @@ func waitGrpcExporter(providerServices map[string]*config.ServiceConfig) {
 	t := time.NewTicker(50 * time.Millisecond)
 	defer t.Stop()
 	pLen := len(providerServices)
-	ta := time.After(10 * time.Second)
+	ta := time.NewTimer(10 * time.Second)
+	defer ta.Stop()
```
AlexStocks, line 119: hey, guy, this is not a good code example. just stop the timer in no ta.C case branch. Another method is as follows:

```go
var flag uberatomic.Bool

defer func() {
    if !flag {
        ta.Stop()
    }
}
...

case time.C:

flag.Set(true)

```

tools/cli/client/client.go
```diff
@@ -143,11 +143,12 @@ func (t *TelnetClient) processSingleRequest(req *protocol.Request, userPkg inter
 	go t.readInputData(string(inputData), requestDataChannel)
 	go t.readServerData(t.conn, responseDataChannel)
 
-	timeAfter := time.After(t.responseTimeout)
+	timeAfter := time.NewTimer(t.responseTimeout)
+	defer timeAfter.Stop()
```
AlexStocks, line 147: the same as above.

protocol/grpc/server.go
mark4z: It is indeed necessary to call B once when the timeout use defer, but it is too complicated to set the flag in each branch. defer is obviously a bit more concise.

mark4z: And dial.dialParallel() in GO SDK also used defer, like:

```go
	// Start the timer for the fallback racer.
	fallbackTimer := time.NewTimer(sd.fallbackDelay())
	defer fallbackTimer.Stop()

	for {
		select {
		case <-fallbackTimer.C:
			fallbackCtx, fallbackCancel := context.WithCancel(ctx)
			defer fallbackCancel()
			go startRacer(fallbackCtx, false)

		case res := <-results:
			if res.error == nil {
				return res.Conn, nil
			}
			if res.primary {
				primary = res
			} else {
				fallback = res
			}
			if primary.done && fallback.done {
				return nil, primary.error
			}
			if res.primary && fallbackTimer.Stop() {
				// If we were able to stop the timer, that means it
				// was running (hadn't yet started the fallback), but
				// we just got an error on the primary path, so start
				// the fallback immediately
```

AlexStocks: good. thx for your info.

AlexStocks (approved):

zhaoyunxing92 (approved):
approved these changes

PR #1298, opened 2021-07-03
Rft: Metadata service

xavier-niu:
**What this PR does**:

- Rename package names of metadata service both for local and remote.
- Update README.

Fixes #

This PR is opened after talking with @flycash. IMO, name problems should be resolved ASAP in release candidate stages.

```release-note
Users may encounter some errors when calling `inmemory.MetadataService` or `remote.MetadataService`, dubbo-go should be updated to the latest version to address that problem.
```


flycash:
I am not sure if this is a good idea because those package has been exposed to users.

xavier-niu:
> I am not sure if this is a good idea because those package has been exposed to users.

In fact, the remote.RemoteMetedataService was renamed in the RC2. I think the package name should be changed ASAP if the name could be determined and be immutable in the future, it could help the people who are new to dubbo-go know the components quickly.

If the compatibility with older systems is one of the important things, I think `inmemory` package could be reserved and marked as deprecated in the comments.

flycash:
I understand your concern. So I say I am not sure if it's a good idea. @AlexStocks @zouyx what do you think?

AlexStocks:
@flycash I think there are not so many people using this package now. It's a good time for us to change the package name.

AlexStocks (approved):

flycash (approved):

hxmhlt (approved):

PR #1516, opened 2021-10-12
Update triple and some relies version, add consumer registry timeout configuration item

LaurenceLiZhixin:


LaurenceLiZhixin:
3.0 Samples is no needs to run time.Sleep() to wait for service discovery, and the max wait time can be configured in consumer level, with api or config file.

PR #1571, opened 2021-11-08
[fix] start application need prioritizing reference service

XiaoWeiKIN:
why ?

因为我对外提供服务时，需要依赖引入的上游服务。如果先启动 provider 侧，服务注册到注册中心已经被下游发现，引入上游服务还未完成，会造成应用“冷启动”报错。


LaurenceLiZhixin:
@AlexStocks @AlbumenJ @justxuewei What do you think?

justxuewei:
I think this is a valuable suggestion. But the hardest things is how to config that? Hard-coding launch sequence in config file is not a smart way in my opinion.

LaurenceLiZhixin (approved):

justxuewei (approved):

Mulavar (approved):

PR #1604, opened 2021-11-20
fix tps filter panic bug(pre-check the configuration of tps at the st…

Mulavar:
…art phase)

**What this PR does**:

Fixes #

```release-note

```

### DIFF
--- a/config/method_config.go
+++ b/config/method_config.go
@@ -18,11 +18,17 @@
 package config
 
 import (
+	"fmt"
+	"strconv"
+)
+
+import (
 	"github.com/creasty/defaults"
 )
 
 import (
 	"dubbo.apache.org/dubbo-go/v3/common/constant"
+	"dubbo.apache.org/dubbo-go/v3/common/extension"
 )
 
 // MethodConfig defines method config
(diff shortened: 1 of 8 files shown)


config/method_config.go
```diff
@@ -70,9 +76,34 @@ func initProviderMethodConfig(sc *ServiceConfig) error {
 }
 
 // check set default value and verify
-func (mc *MethodConfig) check() error {
-	if err := defaults.Set(mc); err != nil {
+func (m *MethodConfig) check() error {
+	qualifieldMethodName := m.InterfaceName + "#" + m.Name
+	if m.TpsLimitStrategy != "" {
+		_ = extension.GetTpsLimitStrategyCreator(m.TpsLimitStrategy)
```
flycash, line 82: why we call this? you don't use the return value.

flycash: I prefer that you add one more function to check if the creator exist. And then you log error here.

Mulavar: call this to check if we enable the corresponding strategy, maybe adding one more function is a better way to detect this, I'll fix it.

LaurenceLiZhixin:
@Mulavar Please check this conflict.

AlexStocks (approved):

PR #1688, opened 2021-12-27
[ISSUE #1687] Fix the problem that the heartbeat cannot be reported

### LINKED ISSUE #1687: When using the Polaris registry, InstanceRegisterRequest needs to set the value of the TTL field
**What happened**:

If the TTL field is not set, polaris will think that this instance does not need to perform the heartbeat health check operation, then after the instance goes offline, the instance cannot be converted to unhealthy normally.

chuntaojun:
**What this PR does**:

Fixes #1687 

```release-note

```

### DIFF
--- a/registry/polaris/registry.go
+++ b/registry/polaris/registry.go
@@ -46,7 +46,8 @@ var localIP = ""
 
 const (
 	// RegistryConnDelay registry connection delay
-	RegistryConnDelay = 3
+	RegistryConnDelay                  = 3
+	defaultHeartbeatIntervalSec = 5
 )
 
 func init() {
@@ -302,7 +303,7 @@ func createRegisterParam(url *common.URL, serviceName string) *api.InstanceRegis
 		},
 	}
 
-	req.SetTTL(5)
+	req.SetTTL(defaultHeartbeatIntervalSec)
 
 	return req
 }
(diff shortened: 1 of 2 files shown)


registry/polaris/registry.go
```diff
@@ -301,6 +301,10 @@ func createRegisterParam(url *common.URL, serviceName string) *api.InstanceRegis
 			Metadata:  metadata,
 		},
 	}
+
+	req.SetTTL(5)
```
justxuewei, line 305: Why do you use 5 instead of other number here?

chuntaojun: Nacos default heartbeat reporting time at 5 seconds, so here is a direct reference

justxuewei: Ok, could you define a constant and replace 5 with it here?

chuntaojun: ok

justxuewei:
Please perform `go fmt ./...` on your local environment, and push the code again to pass the CI.

AlexStocks (approved):

zhaoyunxing92 (approved):

liu-song:
https://github.com/apache/dubbo-go/blob/master/registry/polaris/service_discovery.go#L97，这里log里面的 nacos  可以顺带改下

AlexStocks (approved):

LaurenceLiZhixin (approved):

PR #1955, opened 2022-07-06
Imp: remove useless time format

### LINKED ISSUE #1954: Unnecessary time formatting in NewHillClimbingUpdater of adaptive service
**What happened:**
When I tested the adaptive service, I found that the cpu load on the adaptive service was much higher, I used pprof to analyse it and found that a lot of CPU time was spent on the time formatting in the `NewHillClimbingUpdater` function.
image
I found that even though `adaptive-service-verbose` is set to false, the code still converts the time to a string and passes it to `VerboseDebugf`, which is completely unnecessary
https://github.com/apache/dubbo-go/blob/b7e348319f82444e18ecda284e20656d7ed3b861/filter/adaptivesvc/limiter/hill_climbing.go#L125-L135
**What you expected to happen:**
When `adaptive-service-verbose` is set to false, formatting time is no longer executed.
 
**How to reproduce it (as minimally and precisely as possible):**
The following is set in conf:
```
    adaptive-service: true
    adaptive-service-verbose: false
```

**Anything else we need to know?:**

CoolIceV:
**What this PR does**: 
Whether this log is output or not, the date will be formatted and cost a lot time, now it will be output without formatting as `2022-07-06 19:37:39.4692053 +0800 CST m=+0.775520901`

Fixes #1954 

- [x] All ut passed (run 'go test ./...' in project root)
- [x] After go-fmt ed , run 'go fmt project' using goland.
- [x] Golangci-lint passed, run 'sudo golangci-lint run' in project root.
- [x] After import formatted, (using [imports-formatter](https://github.com/dubbogo/tools#5-how-to-get-imports-formatter) to run 'imports-formatter .' in project root, to format your import blocks, mentioned in [CONTRIBUTING.md](https://github.com/apache/dubbo-go/blob/master/CONTRIBUTING.md) above) 
- [x] Your new-created file needs to have [apache license](https://raw.githubusercontent.com/dubbogo/resources/master/tools/license/license.txt) at the top, like other existed file does.
- [x] All integration test passed. You can run integration test locally (with docker env). Clone our [dubbo-go-samples](https://github.com/apache/dubbo-go-samples) project and replace the go.mod to your dubbo-go, and run 'sudo sh start_integration_test.sh' at root of samples project root. (M1 Slice is not Support)

### DIFF
--- a/filter/adaptivesvc/limiter/hill_climbing.go
+++ b/filter/adaptivesvc/limiter/hill_climbing.go
@@ -130,7 +130,7 @@ func NewHillClimbingUpdater(limiter *HillClimbing) *HillClimbingUpdater {
 		seq:       limiter.seq.Add(1) - 1,
 	}
 	VerboseDebugf("[NewHillClimbingUpdater] A new request arrived, seq: %d, inflight: %d, time: %s.",
-		u.seq, inflight, u.startTime.String())
+		u.seq, inflight, u.startTime)
 	return u
 }
 


justxuewei:
Hi Coollce, thanks for your contribution, if `String()` is removed, how much the performance can be improved?

sanxun0325 (approved):

CoolIceV:
I  tested it roughly. When the QPS was high, there was about a ten or so milliseconds' improvement, but when the QPS was low, there didn't seem to be much optimization.

sanxun0325 (commented):
%s print time, eventually the String() function is called as well。  If must print the time, you can make a special note。
this should not matter in debug print log mode

CoolIceV:
> %s print time, eventually the String() function is called as well。 If must print the time, you can make a special note。 this should not matter in debug print log mode

I've now just delayed the String() call, meaning that if I don't want to print the log, I won't call the String() method, the original code would have called it whether or not the log was printed.
Original:
image
Modified:
image

CoolIceV:
> 如果只是打印当前时间的话是不是用log的时间就可以了

这个需要看log中的时间与所想要打印的时间会有多大的差距，这个本来就是debug用的，我觉得现在这样处理就可以，不用太纠结这个

justxuewei (approved):

AlexStocks (approved):

PR #1961, opened 2022-07-08
Fix: fix bug when creating metrics objects concurrently

### LINKED ISSUE #1953: Prometheus reporters fail to prevent creating concurrently
**What happened:**
When I want to collect custom metrics in filter, some metrics never changed. I think it may be caused by concurrent creation of metrics objects, the code tries to use syncMap's Load to avoid duplicate creation, but this is useless. 
https://github.com/apache/dubbo-go/blob/b7e348319f82444e18ecda284e20656d7ed3b861/metrics/prometheus/reporter.go#L296-L309
When I debug, I find that the Register() returns an AlreadyExist error, but this error is incorrectly ignored

**What you expected to happen:**
The same metrics object should be created only once and needs to be guaranteed to be registered successfully.

**How to reproduce it (as minimally and precisely as possible):**
image

**Anything else we need to know?:**
In addition to counter, metrics such as Gauge should also have this problem

CoolIceV:
**What this PR does**: 
fix bug when creating metrics objects concurrently

Fixes #1953 

### DIFF
--- a/metrics/prometheus/reporter.go
+++ b/metrics/prometheus/reporter.go
@@ -252,11 +252,11 @@ func (reporter *PrometheusReporter) setGauge(gaugeName string, toSetValue float6
 	if len(labelMap) == 0 {
 		// gauge
 		if val, exist := reporter.userGauge.Load(gaugeName); !exist {
-			newGauge := newGauge(gaugeName, reporter.namespace)
-			err := prom.DefaultRegisterer.Register(newGauge)
+			gauge := newGauge(gaugeName, reporter.namespace)
+			err := prom.DefaultRegisterer.Register(gauge)
 			if err == nil {
-				reporter.userGauge.Store(gaugeName, newGauge)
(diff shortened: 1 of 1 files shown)


AlexStocks (approved):

metrics/prometheus/reporter.go
```diff
@@ -253,10 +253,16 @@ func (reporter *PrometheusReporter) setGauge(gaugeName string, toSetValue float6
 		// gauge
 		if val, exist := reporter.userGauge.Load(gaugeName); !exist {
 			newGauge := newGauge(gaugeName, reporter.namespace)
```
Mulavar, line 255: maybe it's better to use other variable names to avoid confusion

justxuewei (approved):

AlexStocks:
@Mulavar pls review agaiin.

Mulavar (approved):

PR #1999, opened 2022-07-26
[#1998]fix a bug where filter was not currently working properly in triple protocol

Mulavar:
…fix some spelling mistake

**What this PR does**: 
fix a bug where filter was not currently working properly and fix some spelling mistake

**Which issue(s) this PR fixes**: 
Fixes #1988 

**You should pay attention to items below to ensure your pr passes our ci test**
We do not merge pr with ci tests failed

- [ ] All ut passed (run 'go test ./...' in project root)
- [ ] After go-fmt ed , run 'go fmt project' using goland.
- [ ] Golangci-lint passed, run 'sudo golangci-lint run' in project root.
- [ ] After import formatted, (using [imports-formatter](https://github.com/dubbogo/tools#5-how-to-get-imports-formatter) to run 'imports-formatter .' in project root, to format your import blocks, mentioned in [CONTRIBUTING.md](https://github.com/apache/dubbo-go/blob/master/CONTRIBUTING.md) above) 
- [ ] Your new-created file needs to have [apache license](https://raw.githubusercontent.com/dubbogo/resources/master/tools/license/license.txt) at the top, like other existed file does.
- [ ] All integration test passed. You can run integration test locally (with docker env). Clone our [dubbo-go-samples](https://github.com/apache/dubbo-go-samples) project and replace the go.mod to your dubbo-go, and run 'sudo sh start_integration_test.sh' at root of samples project root. (M1 Slice is not Support)

PR #2001, opened 2022-07-30
Fix: token filter type assertion

AnneYang720:
**What this PR does**: 
Fix type assertion in token filter.
remoteTkn must be assert to []string[0], because of http2 header limitation.


Mulavar:
I'm sorry that issue #1998 has been fixed, the purpose of this pr is the same as the other pr #1999 .
I saw your pr, and maybe you forgot that the token filter should be compatible with the dubbo protocol too, so we need to check both type assertion with string and []string.
@AnneYang720

PR #2022, opened 2022-08-18
Fix service discovery related issues and add mesh proxy mode support

chickenlj:


chickenlj:
Samples would be added to dubbo-go-samples soon

config/registry_config.go
```diff
@@ -167,6 +169,122 @@ func (c *RegistryConfig) toURL(roleType common.RoleType) (*common.URL, error) {
...
+	//if roleType == common.PROVIDER {
+	//	registryURLs = genCompatibleRegistries(registryURLs)
+	//}
+
+	return registryURLs
+}
+
+//// GenCompatibleRegistries check register-mode to see if we need to do double registration for migration or compatibility reasons,
+//// if so, generate registry url of that kind.
+//func genCompatibleRegistries(registryURLs []*common.URL) []*common.URL {
```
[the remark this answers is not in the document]
chickenlj: done

chickenlj:
https://github.com/apache/dubbo-go-samples/pull/411

Merge this pull request first to pass the CI.

Mulavar:
please squash your commits with git rebase.

config/registry_config.go
```diff
@@ -41,20 +41,22 @@ import (
...
+	Username          string            `yaml:"username" json:"username,omitempty" property:"username"`
+	Password          string            `yaml:"password" json:"password,omitempty"  property:"password"`
+	Simplified        bool              `yaml:"simplified" json:"simplified,omitempty"  property:"simplified"`
+	Preferred         bool              `yaml:"preferred" json:"preferred,omitempty" property:"preferred"` // Always use this registry first if set to true, useful when subscribe to multiple registriesConfig
+	Zone              string            `yaml:"zone" json:"zone,omitempty" property:"zone"`                // The region where the registry belongs, usually used to isolate traffics
+	Weight            int64             `yaml:"weight" json:"weight,omitempty" property:"weight"`          // Affects traffic distribution among registriesConfig, useful when subscribe to multiple registriesConfig Take effect only when no preferred registry is specified.
+	Params            map[string]string `yaml:"params" json:"params,omitempty" property:"params"`
+	RegistryType      string            `yaml:"registry-type"`
```
Mulavar, line 57: Is this a required attribute? If not, should we add `default` tag

chickenlj:
> please squash your commits with git rebase.

Why is this necessary?  I think 'squash and merge' would do the same thing for us if you worry too many commits would be  brought in the repo after merging, it will compose all commits into one commit and the only thing that needs to do when using that is to replace the commit record listed there to one single conclusion.

config/registry_config.go
chickenlj: No explicit default value is required, just leaving it with the zero value is fine.  That way would also make it easier to tell whether the user has set the value.

chickenlj: It's the framework's responsibility to check if `RegistryType == ""`, if so, the framework will know that the user does not specify the value and does expect action.

Mulavar:
>  the only thing that needs to do when using that is to replace the commit record listed there to one single conclusion.
yeah, that's what I want to mean. If you think this is not a problem, I am Ok with this.

Mulavar (approved):

common/metadata_info.go
```diff
@@ -205,7 +205,7 @@ func (si *ServiceInfo) GetMethods() []string {
 		s := si.Params[constant.MethodsKey]
 		return strings.Split(s, ",")
 	}
-	methods := make([]string, 8)
+	methods := make([]string, 0)
```
binbin0325, line 208: make([]string, 0, 8) maybe it's better

common/extension/metadata_report_factory.go
```diff
@@ -31,7 +31,8 @@ func SetMetadataReportFactory(name string, v func() factory.MetadataReportFactor
 // GetMetadataReportFactory finds the MetadataReportFactory with @name
 func GetMetadataReportFactory(name string) factory.MetadataReportFactory {
 	if metaDataReportFactories[name] == nil {
-		panic("metadata report for " + name + " is not existing, make sure you have import the package.")
+		//panic("metadata report for " + name + " is not existing, make sure you have import the package.")
```
binbin0325, line 34: It might be better to just delete it

config/registry_config.go
```diff
@@ -152,8 +154,10 @@ func (c *RegistryConfig) toURL(roleType common.RoleType) (*common.URL, error) {
 	if c.RegistryType == "service" {
 		// service discovery protocol
 		registryURLProtocol = constant.ServiceRegistryProtocol
-	} else {
+	} else if c.RegistryType == "interface" {
```
binbin0325, line 157: interface constant，might be better

config/registry_config.go
```diff
@@ -167,6 +171,85 @@ func (c *RegistryConfig) toURL(roleType common.RoleType) (*common.URL, error) {
...
+					target = true
+					break
+				}
+			}
+		}
+
+		if target {
+			if urls, err := registryConf.toURLs(roleType); err != nil {
+				logger.Errorf("The registry id: %s url is invalid, error: %#v", k, err)
+				panic(err)
```
binbin0325, line 230: 这里一定要panic吗？

chickenlj:
Kindly remind: please use 'Squash and Merge' when merging.

config/registry_config.go
chickenlj: Yes, I think so. If one or more registries were specified and panic occurs when parsing them, then we should panic to stop the process from being started.

common/extension/metadata_report_factory.go
chickenlj: done

.asf.yaml
```diff
@@ -3,3 +3,29 @@ notifications:
...
+        # Enable wiki for documentation
+        wiki: true
+        # Enable issue management
+        issues: true
+        # Enable projects for project management boards
+        projects: true
+    protected_branches:
+        master:
+            # only disable force push
+            foo: bar
```
justxuewei, line 25: Is it possible to enable some extra rules here? Like merging a PR with at least three approves and CIs passed?

chickenlj: Yes, it is possible to add those restrictions, but I personally don't think it's necessary to add that for the current state of our project. Giving more flexibility is more important now.

Also I guess this should not be a blocker for this pull request.

common/url.go
```diff
@@ -515,6 +530,23 @@ func (c *URL) GetParam(s string, d string) string {
...
+	var r string
+	ok := true
+	if len(c.params) > 0 {
+		r = c.params.Get(s)
+	}
+	if len(r) == 0 {
+		ok = false
+	}
+
+	return r, ok
```
baerwang, line 547: 这个是不是多余了？直接判断是否为空字符串不就好了吗？

common/url.go
```diff
@@ -515,6 +530,23 @@ func (c *URL) GetParam(s string, d string) string {
 	return r
 }
 
+// GetParamNoDefault gets value by key, return nil,false if no value found mapping to the key
+func (c *URL) GetParamNoDefault(s string) (string, bool) {
```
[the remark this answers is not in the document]
chickenlj: > 改名叫 GetParam 就可以

GetParam 这个名字已经被占用了，并且没有 bool 返回值，没啥好办法了。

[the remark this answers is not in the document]
justxuewei: You might consider using `GetParamWithoutDefault` to replace that.

Xezeloh:
GetNonDefaultParam

common/url.go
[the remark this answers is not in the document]
Xezeloh: GetNonDefaultParam怎么样

PR #2032, opened 2022-08-25
[WIP] Feat: Adaptive Service

### LINKED ISSUE #1834: GSoC 2022: Explore a new generation of flexible loadbalance algorithms
Flexible loadbalance is an intelligent cluster scheduling algorithm, which can dynamically allocate traffic according to real-time machine load and hardware configuration, so as to maintain the best performance of the whole system. We will set up a flexible service performance test environment to simulate the real cluster status, including instance capacity expansion, timeout, offline due to failure, etc.

So if you are interested in this project and the objective described in the proposal, please leave comments below so we can further exchange opinions.

Please don't forget to apply for GSoC from its official website when the application slot opens.

CoolIceV:
**What this PR does**: 
Provides a flexible loadbalance in unstable environments.

Fixes #1834

### DIFF
--- a/cluster/cluster/adaptivesvc/cluster_invoker.go
+++ b/cluster/cluster/adaptivesvc/cluster_invoker.go
@@ -19,12 +19,11 @@ package adaptivesvc
 
 import (
 	"context"
-	"strconv"
+	"errors"
+	"time"
 )
 
 import (
-	"github.com/dubbogo/gost/log/logger"
-
 	perrors "github.com/pkg/errors"
 )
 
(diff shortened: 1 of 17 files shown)


CoolIceV:
Now only the `dynamic load balancer` on the consumer side is implemented, which takes into account the recent request success rate and request latency, the effect is as follows.

Flexible services are turned on at the marked line position. After that the QPS of different configured services is no longer the same, and the CPU load and latency of lower configured services are significantly reduced.

qps
cpu
rtt99

The next step is to implement an `adaptive concurrent limiter` on the provider side.

justxuewei:
I am thinking you could submit several PRs, instead of an all-in-one PR. For example: one PR for p2c, one PR for adaptive service.

CoolIceV:
Simple test results for adaptive concurrency limiter.
image

So far, a simple adaptive concurrency limiter has been implemented that automatically limits the concurrency when the CPU load is too high, keeping the number of requests being processed stable.

However, the following issues still need to be fixed.
1. the latency still rises when the QPS is too high and the service still may get crashed.
2. the adaptive policy is too simple to adapt to more complex situations and it needs to be improved with reference to `brpc`

CoolIceV:
> Nit: I am thinking you could submit several PRs, instead of an all-in-one PR. For example: one PR for p2c, one PR for adaptive service.

The commits are a bit confusing right now, I'll sort it out later and break it down into multiple PRs.

cluster/cluster/adaptivesvc/cluster_invoker.go
```diff
@@ -65,45 +64,20 @@ func (ivk *adaptiveServiceClusterInvoker) Invoke(ctx context.Context, invocation
...
+	_ = metrics.SlidingWindowCounterMetrics.SetMethodMetrics(invoker.GetURL(), invocation.MethodName(), metrics.Requests, 1)
 	// invoke
 	invocation.SetAttachment(constant.AdaptiveServiceEnabledKey, constant.AdaptiveServiceIsEnabled)
+	startTime := time.Now().UnixNano()
 	result := invoker.Invoke(ctx, invocation)
-
+	rtt := time.Now().UnixNano() - startTime
 	// if the adaptive service encounters an error, DO NOT
 	// update the metrics.
 	if clsutils.IsAdaptiveServiceFailed(result.Error()) {
```
justxuewei, line 75: If the two test branches have the same logic, I think you should merge them.

cluster/cluster/adaptivesvc/cluster_invoker.go
```diff
@@ -65,45 +64,20 @@ func (ivk *adaptiveServiceClusterInvoker) Invoke(ctx context.Context, invocation
...
-		if str, strOK := remainingIface.(string); strOK {
-			remainingStr = str
-		} else if strArr, strArrOK := remainingIface.([]string); strArrOK && len(strArr) > 0 {
-			remainingStr = strArr[0]
-		}
-	}
-	if remainingStr == "" {
-		logger.Errorf("[adasvc cluster] The %s field type of value %v should be string.",
-			constant.AdaptiveServiceRemainingKey, remainingIface)
+	} else if errors.Is(result.Error(), context.DeadlineExceeded) {
```
justxuewei, line 77: For the ddl execeeded error, it is worth wrapping a function, called `IsDeadlineExceeded()`, in the `clsutils` package.

cluster/loadbalance/p2c/loadbalance.go
```diff
@@ -89,49 +85,62 @@ func (l *p2cLoadBalance) Select(invokers []protocol.Invoker, invocation protocol
...
+
+func successRateWeight(url *common.URL, methodName string) float64 {
+	requestsIface, _ := metrics.SlidingWindowCounterMetrics.GetMethodMetrics(url, methodName, metrics.Requests)
+	acceptsIface, _ := metrics.SlidingWindowCounterMetrics.GetMethodMetrics(url, methodName, metrics.Accepts)
+
+	requests := metrics.ToFloat64(requestsIface)
+	accepts := metrics.ToFloat64(acceptsIface)
+	r := (1 + accepts) / (1 + requests)
+
+	if r > 1 {
```
justxuewei, line 140: What would make r greater than 1? IMO, if it does, something is going wrong. So an error output should be dumpped.

PR #2152, opened 2022-12-08
Fix: Replace assignment behavior with copy operation to avoid OOM problem

### LINKED ISSUE #2031: common.MergeURL占用内存一直上升
pprof001 (1)
image
image

Lvnszn:
**What this PR does**: 
There is a risk of OOM in the assignment of params in MergeURL, which can be avoided by using copy behavior to avoid memory overflow. Initially, it is suspected that the problem is caused by mutual references between serviceURL and referenceURL.

After executing the test logic for more than 4 hours, no memory usage problems were found in MergeURL, even when restarting zk.
image

Here, the paramsKey will also be allocated on the heap through the closure, so the closure logic is split and the problem of closure references is solved by passing parameters.

image

I tested in @mengcaicheng's test environment and noticed that the memory usage of MergeURL has decreased significantly, but the total memory is still in a high state. Initially, it looks like the Clone logic is complex with closure logic inside.

image

image

Fixes #2031

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -35,11 +35,9 @@ import (
 
 	gxset "github.com/dubbogo/gost/container/set"
 
+	"github.com/google/uuid"
 	"github.com/jinzhu/copier"
-
 	perrors "github.com/pkg/errors"
-
-	"github.com/satori/go.uuid"
 )
 
 import (
@@ -203,7 +201,7 @@ func WithToken(token string) Option {
 		if len(token) > 0 {
 			value := token
 			if strings.ToLower(token) == "true" || strings.ToLower(token) == "default" {
-				u, _ := uuid.NewV4()
+				u, _ := uuid.NewUUID()
 				value = u.String()
 			}
 			url.SetParam(constant.TokenKey, value)
@@ -684,7 +682,8 @@ func (c *URL) ToMap() map[string]string {
 // will be added into result.
 // for example, if serviceURL contains params (a1->v1, b1->v2) and referenceURL contains params(a2->v3, b1 -> v4)
 // the params of result will be (a1->v1, b1->v2, a2->v3).
-// You should notice that the value of b1 is v2, not v4.
+// You should notice that the value of b1 is v2, not v4
+// except constant.LoadbalanceKey, constant.ClusterKey, constant.RetriesKey, constant.TimeoutKey.
 // due to URL is not thread-safe, so this method is not thread-safe
 func MergeURL(serviceURL *URL, referenceURL *URL) *URL {
 	// After Clone, it is a new URL that there is no thread safe issue.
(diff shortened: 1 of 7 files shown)


georgehao (approved):

common/url.go
```diff
@@ -711,8 +711,16 @@ func MergeURL(serviceURL *URL, referenceURL *URL) *URL {
...
 	// finally execute methodConfigMergeFcn
 	for _, method := range referenceURL.Methods {
-		for _, fcn := range methodConfigMergeFcn {
-			fcn("methods." + method)
+		for _, paramKey := range []string{constant.LoadbalanceKey, constant.ClusterKey, constant.RetriesKey, constant.TimeoutKey} {
+			if v := referenceURL.GetParam(paramKey, ""); len(v) > 0 {
+				params[paramKey] = []string{v}
+			}
+
+			if len(mergedURL.GetParam("methods."+method+"."+paramKey, "")) == 0 {
```
justxuewei, line 719: Please consider creating a variable to avoid concatenating '"methods."+method+"."+paramKey' many times, like 'key := "methods."+method+"."+paramKey'.

common/url.go
```diff
@@ -711,8 +711,17 @@ func MergeURL(serviceURL *URL, referenceURL *URL) *URL {
...
-		for _, fcn := range methodConfigMergeFcn {
-			fcn("methods." + method)
+		for _, paramKey := range []string{constant.LoadbalanceKey, constant.ClusterKey, constant.RetriesKey, constant.TimeoutKey} {
+			if v := referenceURL.GetParam(paramKey, ""); len(v) > 0 {
+				params[paramKey] = []string{v}
+			}
+
+			methodsKey := "methods." + method + "." + paramKey
+			if len(mergedURL.GetParam(methodsKey, "")) == 0 {
+				if v := referenceURL.GetParam(method+"."+paramKey, ""); len(v) > 0 {
```
justxuewei, line 721: Here should use `methodsKey`.

common/url.go
```diff
@@ -711,8 +711,17 @@ func MergeURL(serviceURL *URL, referenceURL *URL) *URL {
...
+					params[method+"."+paramKey] = []string{v}
```
justxuewei, line 722: Here should use `methodsKey`.

justxuewei (approved):
Lgtm, thanks!

remoting/zookeeper/listener.go
```diff
@@ -63,16 +63,13 @@ func NewZkEventListener(client *gxzookeeper.ZookeeperClient) *ZkEventListener {
...
-	go func(zkPath string, listener remoting.DataListener) {
-		if l.listenServiceNodeEvent(zkPath, listener) {
-			listener.DataChange(remoting.Event{Path: zkPath, Action: remoting.EventTypeDel})
-			l.pathMapLock.Lock()
-			delete(l.pathMap, zkPath)
-			l.pathMapLock.Unlock()
-		}
-		logger.Warnf("ListenServiceNodeEvent->listenSelf(zk path{%s}) goroutine exit now", zkPath)
-	}(zkPath, listener)
+	if l.listenServiceNodeEvent(zkPath, listener) {
```
justxuewei, line 66: If `go func` is removed from here, then the code will be synced and is blocked until `l.listenServiceNodeEvent` returns. Why should it be removed?

justxuewei:
Plus, the ci tests failed, please fix or rerun them.

remoting/zookeeper/listener.go
Lvnszn: When calling this function, it is started asynchronously, so there is no problem with always posting

Lvnszn:
open another pr #2182

PR #2182, opened 2023-01-15
Fix: Replace assignment behavior with copy operation to avoid OOM pro…

### LINKED ISSUE #2031: common.MergeURL占用内存一直上升

Lvnszn:
**What this PR does**: 
There is a risk of OOM in the assignment of params in MergeURL, which can be avoided by using copy behavior to avoid memory overflow. Initially, it is suspected that the problem is caused by mutual references between serviceURL and referenceURL.

After executing the test logic for more than 4 hours, no memory usage problems were found in MergeURL, even when restarting zk.
image

Here, the paramsKey will also be allocated on the heap through the closure, so the closure logic is split and the problem of closure references is solved by passing parameters.

image

I tested in @mengcaicheng's test environment and noticed that the memory usage of MergeURL has decreased significantly, but the total memory is still in a high state. Initially, it looks like the Clone logic is complex with closure logic inside.

image

image

Fixes #2031


binbin0325 (approved):

zhaoyunxing92 (approved):

xuhuiliang-maybe:
When will we release it? Our memory won't hold

justxuewei:
@xuhuiliang-maybe We've released `v3.0.5-rc1` which contains this bugfix, please have a try. Btw, releasing `v3.0.5` is in the Apache releasing progress, please stay tuned.

PR #2246, opened 2023-03-11
[Bugfix] Custom triple server message size

### LINKED ISSUE #2176: the msg size of provider is invalid
## 问题
grpc: received message larger than max (7292359 vs. 4194304)，提示传送的数据超过大小限制
#1644 
参考这个issue，进行配置，但是并没有用
image

mark4z:
According to the pr of #1654, for the client side, the function works normally, but for the server side, the order of Provider Export is not always the same, resulting in sometimes max-server-send-msg-size not taking effect, and for the server side, it is not possible to customize the message size based on the request, so a more appropriate The more appropriate setting should be placed on the protocol.
This pr also fixes the grpc protocol, which can also be used to customize the message size in this way.

**What this PR does**: 
Custom triple/grpc server message size
For message size , The available units are, b/kb/kib/mib/mb/gib/gb, 1kb = 1000b and 1kib = 1024b

Provider:
```yaml
dubbo:
  protocols:
    triple:
      name: tri
      port: 20000
      max-server-send-msg-size: 4mib
      max-server-recv-msg-size: 4mib
```
Consumer:
```yaml
  consumer:
    references:
      GreeterClientImpl:
        protocol: tri
        params:
          max-call-send-msg-size: 4mib
          max-call-recv-msg-size: 4mib
```

Fixes #2176 

- [x] All ut passed (run 'go test ./...' in project root)
- [x] After go-fmt ed , run 'go fmt project' using goland.
- [x] Golangci-lint passed, run 'sudo golangci-lint run' in project root.
- [x] Your new-created file needs to have [apache license](https://raw.githubusercontent.com/dubbogo/resources/master/tools/license/license.txt) at the top, like other existed file does.
- [x] All integration test passed. You can run integration test locally (with docker env). Clone our [dubbo-go-samples](https://github.com/apache/dubbo-go-samples) project and replace the go.mod to your dubbo-go, and run 'sudo sh start_integration_test.sh' at root of samples projM1 Slice is not Support)

### DIFF
--- a/config/protocol_config.go
+++ b/config/protocol_config.go
@@ -31,6 +31,9 @@ type ProtocolConfig struct {
 	Ip     string      `yaml:"ip"  json:"ip,omitempty" property:"ip"`
 	Port   string      `default:"20000" yaml:"port" json:"port,omitempty" property:"port"`
 	Params interface{} `yaml:"params" json:"params,omitempty" property:"params"`
+
+	MaxServerSendMsgSize string `yaml:"max-server-send-msg-size" json:"max-server-send-msg-size,omitempty"`
+	MaxServerRecvMsgSize string `default:"4mib" yaml:"max-server-recv-msg-size" json:"max-server-recv-msg-size,omitempty"`
 }
 
 // Prefix dubbo.config-center
@@ -74,6 +77,16 @@ func (pcb *ProtocolConfigBuilder) SetPort(port string) *ProtocolConfigBuilder {
 
 func (pcb *ProtocolConfigBuilder) SetParams(params interface{}) *ProtocolConfigBuilder {
 	pcb.protocolConfig.Params = params
+	return pcb
+}
+
+func (pcb *ProtocolConfigBuilder) SetMaxServerSendMsgSize(maxServerSendMsgSize string) *ProtocolConfigBuilder {
+	pcb.protocolConfig.MaxServerSendMsgSize = maxServerSendMsgSize
+	return pcb
+}
+
+func (pcb *ProtocolConfigBuilder) SetMaxServerRecvMsgSize(maxServerRecvMsgSize string) *ProtocolConfigBuilder {
+	pcb.protocolConfig.MaxServerRecvMsgSize = maxServerRecvMsgSize
 	return pcb
 }
 
--- a/config/protocol_config.go
+++ b/config/protocol_config.go
@@ -32,7 +32,10 @@ type ProtocolConfig struct {
 	Port   string      `default:"20000" yaml:"port" json:"port,omitempty" property:"port"`
 	Params interface{} `yaml:"params" json:"params,omitempty" property:"params"`
 
+	// MaxServerSendMsgSize max size of server send message, 1mb=1000kb=1000000b 1mib=1024kb=1048576b.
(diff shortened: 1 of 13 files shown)


protocol/grpc/client.go
```diff
@@ -61,7 +61,16 @@ func NewClient(url *common.URL) (*Client, error) {
 	// If not, will return NoopTracer.
 	tracer := opentracing.GlobalTracer()
 	dialOpts := make([]grpc.DialOption, 0, 4)
-	maxMessageSize, _ := strconv.Atoi(url.GetParam(constant.MessageSizeKey, "4"))
+
+	// set max send and recv msg size
+	maxCallRecvMsgSize := constant.DefaultMaxCallRecvMsgSize
+	if recvMsgSize, err := humanize.ParseBytes(url.GetParam(constant.MaxCallRecvMsgSize, "0")); err == nil && recvMsgSize != 0 {
```
justxuewei, line 67: Why does here use "0" as default? For MaxServerRecvMsgSize, the default value is "".

```go
if recvMsgSize, err := humanize.ParseBytes(url.GetParam(constant.MaxServerRecvMsgSize, "")); err == nil && recvMsgSize != 0 {
	maxServerRecvMsgSize = int(recvMsgSize)
}

**Review by justxuewei — COMMENTED** [review]
Thanks @mark4z, a few comments.

File: protocol/dubbo3/dubbo3_protocol.go [hunk 2]
```diff
@@ -242,16 +242,16 @@ func (dp *DubboProtocol) openServer(url *common.URL, tripleCodecType tripleConst
 		}
 	}
 
-	if maxCall := url.GetParam(constant.MaxServerRecvMsgSize, ""); maxCall != "" {
-		if size, err := strconv.Atoi(maxCall); err == nil && size != 0 {
-			opts = append(opts, triConfig.WithGRPCMaxServerRecvMessageSize(size))
-		}
+	maxServerRecvMsgSize := constant.DefaultMaxServerRecvMsgSize
+	if recvMsgSize, err := humanize.ParseBytes(url.GetParam(constant.MaxServerRecvMsgSize, "")); err == nil && recvMsgSize != 0 {
```
justxuewei, line 246: Plus, `recvMsgSize` and `sendMsgSize` should be greater than 0. So the code should be:

```suggestion
	if recvMsgSize, err := humanize.ParseBytes(url.GetParam(constant.MaxServerRecvMsgSize, "")); err == nil && recvMsgSize > 0 {
	// do something...
}
```

config/protocol_config.go
```diff
@@ -31,6 +31,9 @@ type ProtocolConfig struct {
 	Ip     string      `yaml:"ip"  json:"ip,omitempty" property:"ip"`
 	Port   string      `default:"20000" yaml:"port" json:"port,omitempty" property:"port"`
 	Params interface{} `yaml:"params" json:"params,omitempty" property:"params"`
+
+	MaxServerSendMsgSize string `yaml:"max-server-send-msg-size" json:"max-server-send-msg-size,omitempty"`
```
justxuewei, line 35: Please add a comment for its unit, like "1mib=1024b, 1mb=1000b", or  "Please refer to go-humanize package for its unit.".

mark4z: okay

zhaoyunxing92 (approved):

PR #2306, opened 2023-04-23
3.0 remove directory consumer configuration listener

binbin0325:
What this PR does:
remove directory consumer configuration listener

Which issue(s) this PR fixes:
Fixes https://github.com/apache/dubbo-go/issues/2151


justxuewei:
Why does you comment out that code instead of removing them?

binbin0325:
> Why does you comment out that code instead of removing them?

It's ok to delete it, but it's better to see if there are any other problems.

registry/directory/directory.go
```diff
@@ -90,9 +90,9 @@ func NewRegistryDirectory(url *common.URL, registry registry.Registry) (director
 		logger.Warnf("fail to create router chain with url: %s, err is: %v", url.SubURL, err)
 	}
 
-	dir.consumerConfigurationListener = newConsumerConfigurationListener(dir)
-	dir.consumerConfigurationListener.addNotifyListener(dir)
-	dir.referenceConfigurationListener = newReferenceConfigurationListener(dir, url)
+	//dir.consumerConfigurationListener = newConsumerConfigurationListener(dir)
```
justxuewei, line 93: Add a `FIXME` here.

chickenlj:
应该优化相关的错误处理流程，而不是直接删除相关代码逻辑。我先关闭这个pr

PR #2307, opened 2023-04-23
Ignore node not exist error message and continue retry

chickenlj:
When consumer starts before any provider is available, there're error messages that keep printing out on the Consumer side when subscribing/listening to the Zookeeper node.

```text
Get children of path /dubbo/xxxService/providers with watcher failed, the error is zk: node does not exist.
```


chickenlj:
The failed test has nothing to do with changes in this pull request, it's from the Polaris module which needs to be further investigated.

AlbumenJ (approved):

PR #2313, opened 2023-05-11
docs: Change the PATH of cli-v2 in README & RAEDME_CN

shiw-yang:
The link to cli-v2 in README is no longer valid. It has been updated to the latest path. 
Both the README and RAEDME_CN have been modified.


chickenlj (approved):
LGTM

PR #2344, opened 2023-06-27
fix: solve config bool field zero value bug by using pointer

ev1lQuark:
Apply #2338 to main branch.


justxuewei (approved):
Lgtm, thanks!

chickenlj (approved):
LGTM.

PR #2350, opened 2023-07-07
add metrics base api interface

FoghostCn:
add metrics base api interface
  * add base api for collect and expose metric data
  * support metric extension for more indicators

add app info metric and prometheus metric registry impl
  * add metric dubbo_application_info_total


chickenlj (approved):
LGTM.

PR #2354, opened 2023-07-10
fix: disable metrics filter by default instead

ev1lQuark:
Enable multiple dubbo go instances, the metrics function is enabled by default and the default port of the Prometheus server is 9000, which will cause port occupation problems. It is better to turn off the metrics function by default, and leave it to the user to decide whether to enable it and enable port configuration.

---

启用多个 dubbo go 实例，metrics 功能默认开启且 Prometheus server 默认端口为9000，会产生端口占用的问题。不如默认关闭 metrics 功能，是否启用以及启用端口配置交由用户来做决定。


justxuewei (approved):
Lgtm, thanks!

justxuewei (approved):
Lgtm, thanks!

PR #2368, opened 2023-08-02
Triple adaptation and refactor demo

DMwangnima:
This PR introduces new Triple protocol based on Connect library without generic and Triple adaptation.
Main change could be seen in /consumer, /provider and /protocol/triple.
This is a POC.


chickenlj:
Please run go fmt to fix the code format issue.

protocol/triple/internal/client/cmd/main.go
```diff
@@ -0,0 +1,115 @@
...
+	"dubbo.apache.org/dubbo-go/v3/protocol/triple/internal/proto/greettriple"
+	"github.com/dubbogo/gost/log/logger"
+)
+
+func main() {
+	//client initialization
+	cli, err := client.NewClient(
+		client.WithURL("127.0.0.1:8080"),
+	)
+	dubboCli, err := greettriple.NewGreetServiceClient(cli)
```
chickenlj, line 17: Will `GreetService` be better than `GreetServiceClient`?

client/client.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	for _, opt := range opts {
+		opt(refCfg)
+	}
+	return &Client{
+		cfg: refCfg,
+	}, nil
+}
+
+func NewConsumerConfig(opts ...ConsumerOption) (*ConsumerConfig, error) {
```
chickenlj, line 124: I think it's better to expose only `NewClientXxx` varies in this package to users. Move `NewConsumerConfig` and `NewMethodConfig` to a new place or the previous place where it is from.

protocol/triple/internal/client/cmd/main.go
```diff
@@ -0,0 +1,115 @@
...
+	"dubbo.apache.org/dubbo-go/v3/client"
+	_ "dubbo.apache.org/dubbo-go/v3/imports"
+	greet "dubbo.apache.org/dubbo-go/v3/protocol/triple/internal/proto"
+	"dubbo.apache.org/dubbo-go/v3/protocol/triple/internal/proto/greettriple"
+	"github.com/dubbogo/gost/log/logger"
+)
+
+func main() {
+	//client initialization
+	cli, err := client.NewClient(
```
chickenlj, line 14: Is `dubbogo.yml` still needed to run with this API?

client/options.go
```diff
@@ -0,0 +1,352 @@
...
+	perrors "github.com/pkg/errors"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+type CallOptions struct {
+	RequestTimeout string
+	Retries        string
+}
```
chickenlj, line 16: The options exposed here should be further investigated to make sure it's a 100% copy of the `config` package in regard of functionality.

chickenlj (changes requested):

protocol/triple/internal/proto/greettriple/greet.triple.go
```diff
@@ -0,0 +1,385 @@
...
+		return nil, err
+	}
+	return &GreetServiceClientImpl{
+		cli: cli,
+	}, nil
+}
+
+func SetConsumerService(srv common.RPCService) {
+	config.SetClientInfoService(&GreetService_ClientInfo, srv)
+}
```
chickenlj, line 79: How about we remove all the unnecessary old code stubs from the generated file?
For users who rely on old stubs, they can still use the old compiler to generate the code.

chickenlj (changes requested):

protocol/triple/internal/client/cmd/main.go
DMwangnima: Yes, NewGreetService would be better than NewGreetServiceClient cause two client concepts create confusion.

protocol/triple/internal/client/cmd/main.go
DMwangnima: No, dubbogo.yml is only needed for classic configuration (SetConsumerService, config.Load). In this case, just create client with options.

protocol/triple/internal/proto/greettriple/greet.triple.go
DMwangnima: The code generated by old compiler still depends on grpc library. Maybe there is a scenario that user wants to upgrade all services with new compiler? In this case, user also needs a classical config loading way to solve the complex configuration problem.(Assume that we have not provided NewClient with dubbogo.yml)

client/options.go
DMwangnima: Yes, It's a placeholder that waits for professionals to think about what fields should be exposed.

PR #2369, opened 2023-08-04
Fix zookeeper and nacos issues working as registry, metadata and configcenter

chickenlj:
As titled, I tested dubbo-go Nacos and Zookeeper implementation working as registry, metadata, and configcenter in dubbo admin. 

The changes in this pull request are bugfixes made to different parts that finally make admin work.


registry/servicediscovery/service_mapping_change_listener_impl.go
```diff
@@ -63,6 +65,9 @@ func (lstn *ServiceMappingChangedListenerImpl) OnEvent(e observer.Event) error {
 		err error
 		reg registry.Registry
 	)
+
+	lstn.mux.Lock()
```
Lvnszn, line 69: 这个加锁之后中间 return 没有 unlock

chickenlj: 我来加上defer

metadata/report/delegate/delegate_report.go
```diff
@@ -19,6 +19,7 @@ package delegate
 
 import (
 	"encoding/json"
+	gxset "github.com/dubbogo/gost/container/set"
```
Lvnszn, line 22: go get -u github.com/dubbogo/tools/cmd/imports-formatter
imports-formatter

Makefile
```diff
@@ -68,6 +68,11 @@ license: clean prepareLic
 .PHONY: verify
 verify: clean license test
 
+.PHONE: fmt
+fmt:
+	$(GO_GET) -u github.com/dubbogo/tools/cmd/imports-formatter
+	imports-formatter
+
```
justxuewei, line 75: Please add this function to a new pull request. Keep a pull request to implement only one function as much as possible to make code review easier.

justxuewei:
Could you split this pull request into several pull requests? Or could you give some description about this pull request?

Makefile
chickenlj: I will create a new pull request for this part.

chickenlj:
> 

Please see the description I leave above.  

I understand it would be ideal if we submit a pull request containing only changes fixing exactly one issue. In my case here, I don't have a very clear boundary of which issue I am targeting, so the standard I use is to make Nacos being able to work together with Admin.

zhaoyunxing92 (approved):

PR #2374, opened 2023-08-08
Nacos support subscribe to all with '*'

FinalT:


registry/nacos/registry.go
```diff
@@ -169,36 +169,95 @@ func (nr *nacosRegistry) Subscribe(url *common.URL, notifyListener registry.Noti
...
+				logger.Warnf("event listener game over.")
+				return perrors.New("nacosRegistry is not available.")
+			}
 
-	for {
-		if !nr.IsAvailable() {
-			logger.Warnf("event listener game over.")
-			return perrors.New("nacosRegistry is not available.")
-		}
+			services, err := nr.getAllSubscribeServiceNames()
```
chickenlj, line 181: Do we need a scheduled pulling task here? What if the service name list in Nacos server changes after we subscribed to '*', for example, if a new service name is registered, will we still get notified by subscribing to '*'?

binbin0325 (commented):
In what scenarios should subscribe to all services

registry/nacos/registry.go
chickenlj: This can be added later.

chickenlj (approved):
LGTM.

PR #2376, opened 2023-08-09
refactor: make rt metric more performance and easy to use

FoghostCn:
1. optimize rt calculation time
  The current implementation rt will be calculated every time the sample data is added, which is performance-intensive and unnecessary. 
  new implementation will only be calculated when Prometheus pulls data

2. add more unit tests for metrics
3. add Rt interface to `MetricRegisrty` simplfy  rt processing

  for example:
  ```go
  var MetricRegisrty registry
  registry.Rt(&MetricId{Name:"dubbo_request_rt", Desc: "Request RT"}).Observe(100)

  // next time also call it in this way, no need to manually cache it
  registry.Rt(&MetricId{Name:"dubbo_request_rt", Desc: "Request RT"}).Observe(100)
  ```
  it will produce 5 metrics in prometheus:
  ```plain
  # HELP dubbo_request_rt_avg Average request RT
  # TYPE dubbo_request_rt_avg gauge
  dubbo_request_rt_avg{app="dubbo",version="1.0.0"} 45
  # HELP dubbo_request_rt_last Last request RT
  # TYPE dubbo_request_rt_last gauge
  dubbo_request_rt_last{app="dubbo",version="1.0.0"} 90
  # HELP dubbo_request_rt_max Max request RT
  # TYPE dubbo_request_rt_max gauge
  dubbo_request_rt_max{app="dubbo",version="1.0.0"} 90
  # HELP dubbo_request_rt_min Min request RT
  # TYPE dubbo_request_rt_min gauge
  dubbo_request_rt_min{app="dubbo",version="1.0.0"} 0
  # HELP dubbo_request_rt_sum Sum request RT
  # TYPE dubbo_request_rt_sum gauge
  dubbo_request_rt_sum{app="dubbo",version="1.0.0"} 450
  ```

### DIFF
--- a/metrics/prometheus/registry.go
+++ b/metrics/prometheus/registry.go
@@ -39,16 +39,16 @@ func init() {
 }
 
 type promMetricRegistry struct {
-	r prom.Registerer // for convenience of testing
+	r    prom.Registerer // for convenience of testing
 	vecs sync.Map
 }
 
-func (p *promMetricRegistry) getOrComputeVec(key string, supplier func()interface{}) interface{} {
+func (p *promMetricRegistry) getOrComputeVec(key string, supplier func() interface{}) interface{} {
 	v, ok := p.vecs.Load(key)
 	if !ok {
 		v, ok = p.vecs.LoadOrStore(key, supplier())
 		if !ok {
-			p.r.MustRegister(v.(prom.Collector))// only registe collector which stored success
+			p.r.MustRegister(v.(prom.Collector)) // only registe collector which stored success
 		}
 	}
 	return v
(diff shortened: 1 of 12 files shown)


metrics/prometheus/rt_vec.go
```diff
@@ -0,0 +1,137 @@
...
+	tags map[string]string
+	agg  *aggregate.TimeWindowAggregator
+}
+
+func (v *valueAgg) Observe(val float64) {
+	v.agg.Add(val)
+}
+
+func buildKey(m map[string]string, labNames []string) string {
+	var k string
```
Lvnszn, line 72: var buffer bytes.Buffer

    for i := 0; i < 1000; i++ {
        buffer.WriteString("a")
    }

    return buffer.String()

Lvnszn (approved):

ev1lQuark (approved):

ev1lQuark:
LGTM

metrics/prometheus/registry.go
```diff
@@ -35,97 +34,84 @@ import (
...
-		p.mtx.Unlock()
-	}
-	s := vec.With(m.Tags)
-	return &summary{ps: s.(prom.Summary)}
+	}).(*prom.SummaryVec)
+	return vec.With(m.Tags)
 }
 
-func (p *promMetricRegistry) Export() {
+func (p *promMetricRegistry) Rt(m *metrics.MetricId) metrics.ObservableMetric {
```
ev1lQuark, line 97: I think it is more appropriate to call it AggregateRT, and RT should be a set of  global metrics without using time window.

FoghostCn: Maybe I should change this to a normal rt implementation instead of an aggregate rt implementation, because the aggregate rt implementation is quite different from this

PR #2378, opened 2023-08-11
implement triple protocol server-side adaptation

DMwangnima:
Changed provider package to server package and wait for server package finishing.


chickenlj (approved):
LGTM.

PR #2383, opened 2023-08-15
Fix: cache manager test

FinalT:


registry/servicediscovery/store/cache_manager_test.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func TestCacheManager(t *testing.T) {
-	cm, err := NewCacheManager("test", "test_cache", time.Second, 10)
+	cm, err := NewCacheManager("test", "test_cache", time.Second*3, 10)
```
justxuewei, line 30: Why does the time interval affect the results? Does 3 seconds guarantee these tests pass?

FinalT: I dont think it would pass every time, so I update the logic to make surethat the next cm created will be able to read the cache file

chickenlj (approved):
LGTM.

justxuewei (approved):
Lgtm, thanks!

PR #2387, opened 2023-08-18
feat: optimize the integration of otel tracing

### LINKED ISSUE #2386: Optimize the integration of otel tracer
- The Dubbo-go kernel requires additional improvements to the configuration loading 
and component initialization for tracing. By enhancing these areas, applications can 
more easily and comprehensively control Dubbo-go's tracing capabilities through 
simple modifications to the configuration file.
- Currently, Dubbo-go only supports the propagator that adheres to the W3C standards 
for context propagation. Therefore, additional support for the B3 standard is required. 
To implement the B3 protocol, the implementation in opentelemetry-go-contrib can 
be used.
- The configuration of tracing needs to be further aligned with Dubbo-java to improve 
the observability of tracing. This can be achieved by adding support for configuration 
items such as sampling probability to achieve a more refined sampling strategy.
- In terms of observability backends, Dubbo-go currently only supports Jaeger, but it 
should also add support for other backends such as OTel and Zipkin. This would allow 
users to choose their preferred backend for tracing and enable Dubbo-go to integrate 
with various observability systems

ev1lQuark:
Resolves #2386 

Features:
- Simplify the initialization process of the tracing module, by configuration instead of programming
- Support more tracing exporters, include stdout, Jaeger, Zipkin, and Otlp(a bug to be fixed)
- Support for the W3C standand and the B3 standard of context propagation
- Allow user to config sampling probability to achieve a more refined sampling strategy

Bugs:
- Otlp-grpc use grpc health check which has the namespace conflict with dubbo3 protocol


otel/trace/exporter.go
```diff
@@ -0,0 +1,110 @@
...
+type ExporterConfig struct {
+	Exporter         string
+	Endpoint         string
+	SampleMode       string
+	SampleRatio      float64
+	Propagator       string
+	ServiceNamespace string
+	ServiceName      string
+	ServiceVersion   string
+}
```
FoghostCn, line 43: should we use url instead custom config as other module does,maybe we should listen to other people's opinions

FoghostCn (approved):

imports/imports.go
```diff
@@ -66,6 +66,10 @@ import (
 	_ "dubbo.apache.org/dubbo-go/v3/metadata/service/remote"
 	_ "dubbo.apache.org/dubbo-go/v3/metrics/app_info"
 	_ "dubbo.apache.org/dubbo-go/v3/metrics/prometheus"
+	_ "dubbo.apache.org/dubbo-go/v3/otel/trace/jaeger"
+	_ "dubbo.apache.org/dubbo-go/v3/otel/trace/zipkin"
+	//_ "dubbo.apache.org/dubbo-go/v3/otel/trace/otlp" // FIXME: otlp-grpc use grpc health check which has the namespace conflict with dubbo3 protocol
```
HubQin, line 71: May be fixed by #2395

ev1lQuark: Thanks!

PR #2396, opened 2023-08-25
refactor: change metric config  to url like other module

FoghostCn:
1. refactor: metric module use url as config class, `metrics.ReporterConfig` will be removed in the future
2. split metric constant to a single file

The configuration class of the metric module is a copy of the configuration class of the config package. This commit will change the configuration of the metric module to be dependent on `common.URL`, in order to have a consistent style with other modules.


justxuewei:
What is the benifit of using `URL` to init metrics config?

I am thinking the `URL` is abused. It makes others confused since it is used to registry, invoker, etc. And it don't have clear semantics. They are distinguished by its protocol. Think about it, if we don't know its protocol, we can't know what it's used for.

If you still want to use a type of `URL`, please consider defining a new type, like `MetricsAddress`, or something same like that.

FoghostCn:
> What is the benifit of using `URL` to init metrics config?
> 
> I am thinking the `URL` is abused. It makes others confused since it is used to registry, invoker, etc. And it don't have clear semantics. They are distinguished by its protocol. Think about it, if we don't know its protocol, we can't know what it's used for.
> 
> If you still want to use a type of `URL`, please consider defining a new type, like `MetricsAddress`, or something same like that.

1. `metrics.ReporterConfig` is a copy of `config.MetricConfig`,they have duplicate fields, if we add a config field for metric module in yaml, we have to add filed in both `config.MetricConfig` and `metrics.ReporterConfig`,it is ugly.
2. `common.URL` is design to be a unified format for configuration information,All extension points carry configuration information by passing URL,so I think we should passing URL as configuration for metric module.
3. other module like logger、metadata、cluster etc all use URL for configuration,I think this is a default behavior.

chickenlj:
Both ways can meet the requirements of passing metrics configurations for now. I think URL as the current unified design for now is acceptable. 

We can revisit the design later in the following refactor plan.

chickenlj (approved):
LGTM.

PR #2404, opened 2023-08-29
feat: cover Triple stream testing

DMwangnima:
1. Add Triple stream testing and set Grpc protocol as default client protocol.
2. Turn all any to interface{}.


chickenlj (approved):
LGTM.

PR #2406, opened 2023-08-30
feat: add global dubbo instance conception

DMwangnima:
This PR adds global dubbo **Instance** conception (**Instance** may be replaced by a more ideal conception).

In the former PRs, we introduced client layer mapping to config.ReferenceConfig. Then, server layer mapping to config.ServiceConfig will be introduced too. These layers can provide the basic RPC function for users. But for users who hope to inject global configurations (eg. config.ApplicationConfig) and initialize common modules needed by both client and server sides (eg. config.RegistryConfig), it is not enough.

As a result, we introduce a top layer conception **Instance** mapping to config.RootConfig. It is responsible for setting global configurations and common modules. Then, user-side code would changed like this:
```go
import "dubbo.apache.org/dubbo-go/v3"

ins, err := dubbo.NewInstance()
if err != nil {
    panic(err)
}

cli, err := ins.NewClient()
if err != nil {
    panic(err)
}

// then use generated code with cli
```

PS. This layer also needs compatible functions like https://github.com/apache/dubbo-go/pull/2399.


chickenlj:
image


image

chickenlj (approved):
LGTM.

chickenlj:
I suggest we put all configs into one unified package - `global` can currently be the right place.

One principle we must always keep in mind is we need to separate 'config' from 'action', all the new configs must only include getter/setters but no functionalities.

PR #2412, opened 2023-09-01
refactor(root_config): remove redundant nil check in `Init`

Juneezee:
From the Go specification:

> "3. If the map is nil, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/7FQ6fOesRuY

### DIFF
--- a/config/root_config.go
+++ b/config/root_config.go
@@ -167,12 +167,9 @@ func (rc *RootConfig) Init() error {
 	}
 
 	// init registry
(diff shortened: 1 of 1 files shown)


justxuewei (approved):
Lgtm, thanks @Juneezee!

chickenlj (approved):
LGTM.

PR #2413, opened 2023-09-01
feat: copy bare configurations from config to global

DMwangnima:
Based on https://github.com/apache/dubbo-go/pull/2406, we decided to move all bare configurations from config package to **global** package( this package name would be replaced with a more ideal one in the future). After that, the user-side code would change like this:
```go
// global conception
// configure global configurations and common modules
	dubbo.WithApplication(
		global.WithApplication_Name("dubbo_test"),
	),
	dubbo.WithRegistry("nacos",
		global.WithRegistry_Address("127.0.0.1:8848"),
	),
	dubbo.WithMetric(
		global.WithMetric_Enable(true),
	),
)
if err != nil {
	panic(err)
}
// configure the params that only client layer cares
cli, err := ins.NewClient(
	client.WithRetries(3),
)
if err != nil {
	panic(err)
}
// then use this cli with code generated by protoc-gen-triple
```
For user-side, users make use of global.WithXXX_XXX(eg. WithApplication_Name) to create specific configuration. Then users invoke dubbo.WithXXX or client.WithXXX to inject these configurations. By this approach, every module could rely on global package and there is no circular dependencies problem.

introducing a global package seems confusing, but we have strong reason. For compatibility, config.Load must be supported in new dubbo-go version. It means that config package as the entrance should rely on other modules. If client, server and dubbo packages make use of configurations from client package directly, there would be circular dependencies. So we have to introduce this intermediate package.

In the future, we hope that bare configurations from client package would be moved to global package and related actions and logics of configuration would be moved to related module (eg. all functions belong to config.RegistryConfig would be moved to registry package and exported as API).


chickenlj (approved):
LGTM.

PR #2414, opened 2023-09-02
feat: enhance user-side configuration style

DMwangnima:
Please see protocol/triple/internal/client for details.

We moved all bare configurations from config package to global package almostly. Please refer to https://github.com/apache/dubbo-go/pull/2406 https://github.com/apache/dubbo-go/pull/2413. These configurations contain **ReferenceConfig**. As a result, user-side code would change like this if users make use of client layer directly:
```go
cli, err := client.NewClient(
    client.WithReferenceConfig(
        global.WithReference_URL("127.0.0.1:20000"),
    ),
)
```
For users getting started, understanding the conception of Reference is a little bit difficult. They would be more receptive to this form:
```go
cli, err := client.NewClient(
    client.WithURL("127.0.0.1:20000"),
)
```
They do not need to understand **Reference** and import **global** package. All the contents they need belong to **client** package.

We provide this type of API for **client** package because **client** package is a very clear module. Configurations only belong to a module could be provided with <module_name>.WithXXX. And configurations relied on by many modules could be provided with global.With<config_name>_XXX.


chickenlj (approved):
LGTM.

PR #2415, opened 2023-09-02
feat(metric): Add prometheus pushgateway support

FoghostCn:
1. removed metrics configurations:
    * `Mode`: only supports pull mode before,now use `prometheus.exporter.enabled` and `prometheus.pushgateway.enabled` instead, and pull mode is the default behavior if metrics.enable is true
    * `Namespace`: does not support modification. it is useless now
    * `PushGatewayAddress`: never supported before. `prometheus.pushgateway.*` is the new config
2. removed file:
    * `common/extension/metrics.go` and `common/extension/metrics_test.go`, since metric extension way is changed,they are useless now. see #2392 for more information .
    * `metrics/prometheus/reporter.go`,move export method to `metricRegistry.export`
3. Add prometheus pushgateway support, we can enabled it by config:
```yaml
  metrics:
    prometheus:
      pushgateway:
        enabled: true
        base-url: "localhost:9999"
        job: myjob
        username: dubbo
        password: 1234
        push-interval: 10
```


zhaoyunxing92 (approved):

ev1lQuark (approved):
Nice work! LGTM.

PR #2421, opened 2023-09-10
feat: add server side

L2ncE:


common/url.go
```diff
@@ -110,10 +111,11 @@ type URL struct {
...
 
-	Path     string // like  /com.ikurento.dubbo.UserProvider
-	Username string
-	Password string
-	Methods  []string
+	Path       string // like  /com.ikurento.dubbo.UserProvider
+	Username   string
+	Password   string
+	Methods    []string
+	MethodInfo []server.MethodInfo
```
DMwangnima, line 118: Maybe we should discuss with others about introducing new field into URL.

chickenlj: Yes, we should revisit the necessity of adding this new field and its possible affect.

chickenlj: I think we can use `URL.Attributes` to deliver `MethodInfo`, `URL.Attributes` is designed for delivering objects for runtime use.

chickenlj (changes requested):

chickenlj:
I will merge this pull request first for later development convenience. We will definitely need to check the URL part change introduced in this pull request.

PR #2424, opened 2023-09-13
feat(metrics): add switch for metric collector

FoghostCn:
add switch for metric collector, include metadata、registry、config-center、rpc

split `PrometheusRegistry.Export` method into two method to avoid large method body


ev1lQuark (approved):
LGTM.

PR #2427, opened 2023-09-16
feat: modularize Graceful_Shutdown functionality

DMwangnima:
In the original implementation of Graceful_Shutdown, all the processing logics exist in **RootConfig.Start()** (**gracefulShutdownInit**). As shown below:
```go
func (rc *RootConfig) Start() {
	startOnce.Do(func() {
		gracefulShutdownInit()
		rc.Consumer.Load()
		rc.Provider.Load()
		exportMetadataService()
		registerServiceInstance()
	})
}
```
It is not convenient for just creating a **Client** because this behavior is global and embedded in **config** package.
After modularizing Graceful_Shutdown functionality, **Client** and **Server** modules could invoke **Init** and **RegisterProtocol** directly. Here is the code sample:
```go
graceful_shutdown.Init(&global.ShutdownConfig{
    Timeout: "300ms"
})
// iterate protocols configured
for protocolName... {
    graceful_shutdown.RegisterProtocol(protocolName)
}
```
About extensibility, we can add RegisterXXX functions and XXX represents certain resource that need to be destroyed.

### DIFF
--- a/graceful_shutdown/common.go
+++ b/graceful_shutdown/common.go
@@ -18,24 +18,17 @@
 package graceful_shutdown
 
 import (
-	"dubbo.apache.org/dubbo-go/v3/config"
-	"dubbo.apache.org/dubbo-go/v3/global"
-	"go.uber.org/atomic"
+	"github.com/dubbogo/gost/log/logger"
+	"time"
 )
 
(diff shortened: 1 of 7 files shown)


chickenlj (approved):
LGTM.

PR #2436, opened 2023-09-28
fix: config random port

maguowei:
https://github.com/apache/dubbo-go/pull/2384 is missing a condition check, so it still cannot use random ports.

@chickenlj @AlexStocks


chickenlj (approved):
LGTM.

Issue #2428, opened 2023-09-17
[zookeeper][service_discovery] rootPath support customize，not just "/services"


xiaomiusa87:
**What would you like to be added**:
Use zookeeper as registryCenter，expected customize 'rootPath'，not just "/services"

**Why is this needed**:
We use 'rootPath' to manage multiple clusters， want to call Java through dubbo-go

FoghostCn:
why not use different application name or group name to manage multiple clusters

xiaomiusa87:
> why not use different application name or group name to manage multiple clusters

application name or group name is business attributes，has already been used. 
we are middleware developer ，plan multiple clusters by business type.  may use multi zk or one zk，so need use different 
 rootPath

PR #2437, opened 2023-10-06
zk registry support customize 'rootPath'

xiaomiusa87:
[#2428] zk registry support customize 'rootPath'

### DIFF
--- a/registry/zookeeper/service_discovery.go
+++ b/registry/zookeeper/service_discovery.go
@@ -64,10 +64,14 @@ type zookeeperServiceDiscovery struct {
 
 // newZookeeperServiceDiscovery the constructor of newZookeeperServiceDiscovery
 func newZookeeperServiceDiscovery(url *common.URL) (registry.ServiceDiscovery, error) {
-	rp := url.GetParam(constant.ZookeeperRootPath, rootPath)
+	group := url.GetParam(constant.RegistryGroupKey, rootPath)
+	if !strings.HasPrefix(group, constant.PathSeparator) {
+		group = constant.PathSeparator + group
+	}
+
 	zksd := &zookeeperServiceDiscovery{
 		url:                 url,
-		rootPath:            rp,
+		rootPath:            group,
 		instanceListenerMap: make(map[string]*gxset.HashSet),
 	}
 	if err := zookeeper.ValidateZookeeperClient(zksd, url.Location); err != nil {
@@ -75,7 +79,7 @@ func newZookeeperServiceDiscovery(url *common.URL) (registry.ServiceDiscovery, e
 	}
 	zksd.WaitGroup().Add(1) // zk client start successful, then wg +1
 	go zookeeper.HandleClientRestart(zksd)
-	zksd.csd = curator_discovery.NewServiceDiscovery(zksd.client, rp)
+	zksd.csd = curator_discovery.NewServiceDiscovery(zksd.client, group)
 	return zksd, nil
 }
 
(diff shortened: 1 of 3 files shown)


DMwangnima:
This improvement needs to be supported by the configuration API or customized filter so that we need to consider the implications.

registry/zookeeper/service_discovery.go
```diff
@@ -64,17 +64,18 @@ type zookeeperServiceDiscovery struct {
 
 // newZookeeperServiceDiscovery the constructor of newZookeeperServiceDiscovery
 func newZookeeperServiceDiscovery(url *common.URL) (registry.ServiceDiscovery, error) {
+	rp := url.GetParam(constant.ZookeeperRootPath, rootPath)
```
DMwangnima, line 67: This param should be injected by **config/ReferenceConfig.Refer** and **config/ServiceConfig.Export**.

xiaomiusa87: I have looked at the source code of Dubbo Java, and rootPath is only related to the registryCenter ZookeeperServiceDiscovery

RootPath is only used during ZookeeperServiceDiscovery initialization, and providers and consumers do not care about this param

DMwangnima: Which configuration style do you prefer to specify **ZookeeperRootPath** if you do not let providers and consumers inject this param?

xiaomiusa87: Sorry, I added a constant based on **ZookeeperKey**. What's the problem with this？
It is transparently transmitted to zookeeperServiceDiscovery through **RegistryConfig.Params**

DMwangnima: My bad, I did not describe it well. I thought you were going to add a generic feature. Using **RegistryConfig.Params** could work. **ZookeeperRootPath** should be "zookeeper.rootPath" like keys related to nacos(https://github.com/apache/dubbo-go/blob/5ac057d3ca6c3c0ea1fa355c6dcdf64a0813c6b9/common/constant/key.go#L246). Maybe we should add comments on **RegistryConfig.Params** since it is not intuitive to other users if there is no document explaining it.

xiaomiusa87: ok, **ZookeeperRootPath** has been modified to "zookeeper. rootPath",  and **RegistryConfig.Params** also added comments .  
Help review again ~

chickenlj: `RegistryConfig.group` is also designed for users to specify the root path. 

So I think `ZookeeperRootPath = "zookeeper.rootPath"` is not necessary, directly using `group` would be fine. Or at least we need to check `group` if 'rootPath' is not set.

chickenlj: The java implementation checked both rootPath and group. Maybe for some legacy consideration, for dubbogo, I personally think simplify check group is enough for RegistryConfig.group is designed for this scenario.

```java
 public static String getRootPath(URL registryURL) {
        String group = ROOT_PATH.getParameterValue(registryURL);
        if (group.equalsIgnoreCase(DEFAULT_GROUP)) {
            group = GROUP_PATH.getParameterValue(registryURL);
            if (!group.startsWith(PATH_SEPARATOR)) {
                group = PATH_SEPARATOR + group;
            }
        }
        return group;
    }
```

xiaomiusa87: I see that currently zookeeper service_discovery. go does not use **group**, thinking about the principle of minimum change, without considering group.
Do I need to add the **group** parameter logic as well? 
Like Dubbo Java:
```Java
public static String getRootPath(URL registryURL) {
    String group = ROOT_PATH.getParameterValue(registryURL);
    if (group.equalsIgnoreCase(DEFAULT_GROUP)) {
        group = GROUP_PATH.getParameterValue(registryURL);
        if (!group.startsWith(PATH_SEPARATOR)) {
            group = PATH_SEPARATOR + group;
        }
    }
    return group;
}
```

xiaomiusa87: replace zk **rootPath** with  **RegistryConfig.group**
Please review again

chickenlj (approved):
LGTM.

PR #2445, opened 2023-10-11
feat: simplify registry and protocol configuration

DMwangnima:
This PR aims to solve issue(https://github.com/apache/dubbo-go/issues/2444) and unify options design.
For instance, in former implementation, we use DefaultOptions and apply those Option to DefaultOptions directly.
```go
// dubbo.WithRegistry
func WithRegistry(key string, opts ...registry.Option) ClientOption {
	regOpts := registry.DefaultOptions()
	for _, opt := range opts {
		opt(regOpts)
	}

	return func(cliOpts *ClientOptions) {
		if cliOpts.Registries == nil {
			cliOpts.Registries = make(map[string]*global.RegistryConfig)
		}
		cliOpts.Registries[key] = regOpts.Registry
		cliOpts.Registries[regOpts.ID] = regOpts.Registry
	}
}
```
We should move this logic to registry module so that extensions would be easy to add. As a result, introducing  ```NewOptions(opts ...Option) *Options```  in each module. Then **dubbo.WithRegistry** would become:
```go
func WithRegistry(opts ...registry.Option) InstanceOption {
	regOpts := registry.NewOptions(opts...)

	return func(insOpts *InstanceOptions) {
		if insOpts.Registries == nil {
			insOpts.Registries = make(map[string]*global.RegistryConfig)
		}
		insOpts.Registries[regOpts.ID] = regOpts.Registry
	}
}
```
We could add some logics into NewOptions and do not expose these details to other modules.
```go
func NewOptions(opts ...Option) *Options {
	defOpts := defaultOptions()
	for _, opt := range opts {
		opt(defOpts)
	}

	if defOpts.Registry.Protocol == "" {
		panic(fmt.Sprintf("Please specify registry, eg. WithZookeeper()"))
	}
	if defOpts.ID == "" {
		defOpts.ID = defOpts.Registry.Protocol
	}

	return defOpts
}
```

### DIFF
--- a/protocol/options.go
+++ b/protocol/options.go
@@ -18,6 +18,7 @@
 package protocol
 
 import (
+	"fmt"
 	"strconv"
 )
 
@@ -27,10 +28,28 @@ import (
 
 type Options struct {
 	Protocol *global.ProtocolConfig
+
+	ID string
 }
 
-func DefaultOptions() *Options {
+func defaultOptions() *Options {
 	return &Options{Protocol: global.DefaultProtocolConfig()}
+}
+
+func NewOptions(opts ...Option) *Options {
+	defOpts := defaultOptions()
+	for _, opt := range opts {
+		opt(defOpts)
+	}
+
+	if defOpts.Protocol.Name == "" {
+		panic(fmt.Sprintf("Please specify registry, eg. WithTriple()"))
+	}
+	if defOpts.ID == "" {
+		defOpts.ID = defOpts.Protocol.Name
+	}
+
+	return defOpts
 }
 
 type Option func(*Options)
@@ -62,6 +81,14 @@ func WithREST() Option {
 func WithTriple() Option {
 	return func(opts *Options) {
 		opts.Protocol.Name = "tri"
+	}
+}
+
+// WithID specifies the id of protocol.Options. Then you could configure server.WithProtocolIDs and
+// server.WithServer_ProtocolIDs to specify which protocol you need to use in multi-protocols scenario.
+func WithID(id string) Option {
+	return func(opts *Options) {
+		opts.ID = id
 	}
 }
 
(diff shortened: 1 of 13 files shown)


protocol/options.go
```diff
@@ -27,12 +28,30 @@ import (
...
 }
 
+func NewOptions(opts ...Option) *Options {
+	defOpts := defaultOptions()
+	for _, opt := range opts {
+		opt(defOpts)
+	}
+
+	if defOpts.Protocol.Name == "" {
+		panic(fmt.Sprintf("Please specify registry, eg. WithTriple()"))
```
chickenlj, line 46: Should we default to 'triple' protocol if the protocol name is not present?

So, this raises another question, what's the current default behavior if users do not specify any information about the protocol at all? 

``server.NewServer()`

protocol/triple/internal/server/cmd_instance_with_registry/main.go
```diff
@@ -31,11 +31,11 @@ func main() {
 	// configure global configurations and common modules
 	ins, err := dubbo.NewInstance(
 		dubbo.WithName("dubbo_test"),
-		dubbo.WithRegistry("zk",
+		dubbo.WithRegistry(
 			registry.WithZookeeper(),
 			registry.WithAddress("127.0.0.1:2181"),
```
chickenlj, line 36: Is 'registry.WithAddress("zookeeper://127.0.0.1:2181")' a valid method to configure Registry, a combination of WithZookeeper() and WithAddress("127.0.0.1:2181")? We should support this format if we want to give a consistent experience to users.

chickenlj (approved):
LGTM.

protocol/options.go
DMwangnima: For unifying design principle of API, setting default value to **triple** protocol and deleting ```WithTriple()``` make sense. But the consideration is the same as registry configuration.
For instance, setting **zookeeper** as default registry is not intuitive to users because **zookeeper** is not a de facto standard. Users often need to be aware that the concrete registry they are using.
Likewise, **triple** is not a well-known protocol so that users need to configure it clearly.

But now I think setting **triple** as default protocol could help us promote it. **triple** could replace **http** and **grpc** in dubbo ecosphere. These is only **dubbo** protocol that we need to take into separate consideration.

DMwangnima: If users do not specify any information about the protocol at all, server would not serve any protocol. Maybe we should raise an error when users do not provide any protocol configurations when initializing **ServerOptions**?

Issue #2442, opened 2023-10-11
Summarized todos for #2441


chickenlj:
This issue is a collection of todos extracted from https://github.com/apache/dubbo-go/issues/2441

- [ ] Find a better name for `global` package, possible names 'cfg' , 'gConfig', `_global`, etc.
- [x] Move `XxxOption` definition in global package to separate packages https://github.com/apache/dubbo-go/issues/2446
  - [x] protocol
  - [x] registry
  - [x] metrics
  - [x] config-center
  - [x] metadata
  - [x] logger
  - [x] metrics & tracing
  - [x] tls, custom, and profile will not be supported in the first release
- [x] Decide params that should be put in CallOption, timeout, and more? https://github.com/apache/dubbo-go/issues/2448
   - [x] timeout
   - [x] retries
   - [x] xxx?
- [ ] Examine properties in various Configs, like `ServerOption.FilterConf`, etc.
- [ ] Mechanism for passing extra headers and trailers #2435
- [ ] Completely delete `dubbo3` and `grpc` packages
- [ ] RefreshService? reflectionService and healthService
- [x] Simplified version of `dubbo.WithRegistry()`, with default key support.  https://github.com/apache/dubbo-go/issues/2444
- [x] Unify naming of `server.WithRegistry(id, ...)` and `server.WithRegistryIds()` https://github.com/apache/dubbo-go/issues/2444

chickenlj:
https://github.com/apache/dubbo-go/issues/2443

PR #2447, opened 2023-10-12
Fix #2446, move XxxOption definition in global package to separate packages

### LINKED ISSUE #2446: Move XxxOption definition in global package to separate packages

chickenlj:
See #2446 and https://github.com/apache/dubbo-go/issues/2442 for details


logger/options.go
```diff
@@ -0,0 +1,137 @@
+/*
```
DMwangnima, line 1: LoggerConfig has many default values so that ```WithZap(), WithLevelInfo(), WithFormatText(), WithAppenderConsole()``` could be deleted?

config_center/options.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+func WithAppID(id string) Option {
+	return func(opts *Options) {
+		opts.Center.AppID = id
+	}
+}
+
+func WithTimeout(timeout string) Option {
```
DMwangnima, line 116: This should be
```go
func WithTimeout(timeout time.Duration) Option {
    return func(opts *Options) {
        opts.Center.Timeout = timeout
    }
}
```

config_center/options.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+func WithParams(params map[string]string) Option {
+	return func(opts *Options) {
+		opts.Center.Params = params
+	}
+}
+
+func WithFileExtension(extension string) Option {
```
DMwangnima, line 128: This field could be enumerated. Please see (https://github.com/apache/dubbo-go/blob/8b041402f7d7cebb82effc0f6213f3f19f17c9ed/common/constant/file/suffix.go#L22).

metadata/options.go
```diff
@@ -0,0 +1,102 @@
...
+func WithEtcd() Option {
+	return func(opts *Options) {
+		opts.Metadata.Protocol = constant.EtcdV3Key
+	}
+}
+
+func WithAddress(address string) Option {
+	return func(opts *Options) {
+		if i := strings.Index(address, "://"); i > 0 {
+			opts.Metadata.Protocol = address[1:i]
```
DMwangnima, line 68: ditto.

config_center/options.go
```diff
@@ -0,0 +1,132 @@
...
+func WithFile() Option {
+	return func(opts *Options) {
+		opts.Center.Protocol = constant.FileKey
+	}
+}
+
+func WithAddress(address string) Option {
+	return func(opts *Options) {
+		if i := strings.Index(address, "://"); i > 0 {
+			opts.Center.Protocol = address[1:i]
```
DMwangnima, line 68: ditto.

metrics/options.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+func WithAggregationBucketNum(num int) Option {
+	return func(opts *Options) {
+		opts.Metric.Aggregation.BucketNum = num
+	}
+}
+
+func WithAggregationTimeWindowSeconds(seconds int) Option {
```
FoghostCn, line 55: `Duration` may be better

protocol/options.go
```diff
@@ -42,11 +41,13 @@ func NewOptions(opts ...Option) *Options {
 		opt(defOpts)
 	}
 
-	if defOpts.Protocol.Name == "" {
-		panic(fmt.Sprintf("Please specify registry, eg. WithTriple()"))
-	}
 	if defOpts.ID == "" {
-		defOpts.ID = defOpts.Protocol.Name
+		if defOpts.Protocol.Name == "" {
```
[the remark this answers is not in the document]
chickenlj: I think it's better to leave `defOpts.Protocol` as it is. To let framework to decide which protocol to use based on it's default protocol policy.

`defOpts.Protocol.ID` is a special case here, it needs to be set to work as the mapping key.

logger/options.go
chickenlj: I will replace these methods with some more generic methods instead. I did check all the possible values of each type, and they seems to very limited as I listed according to the implementation.

chickenlj: But I will still keep `WithZap()` and `WithLogurus()` as they are currently the two drivers we support.

metrics/options.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func WithEnabled() Option {
+	return func(opts *Options) {
+		b := true
+		opts.Metric.Enable = &b
+	}
+}
+
+func WithPort(port string) Option {
```
DMwangnima, line 114: This should be
```
func WithPort(port int) Option {
    return func(opts *Options) {
        opts.Metric.Port = strconv.Itoa(port)
    }
}
```
?

metrics/options.go
```diff
@@ -0,0 +1,124 @@
...
+		opts.Metric.Prometheus.Pushgateway.PushInterval = int(interval.Seconds())
+	}
+}
+
+func WithEnabled() Option {
+	return func(opts *Options) {
+		b := true
+		opts.Metric.Enable = &b
+	}
+}
```
FoghostCn, line 112: maybe we can remove this option, because it is equivalent to `WithPrometheusExporterEnabled` for now

chickenlj:
There're currently two parts that need to be confirmed:
1. otel/tracing config
2. metrics config

I will submit new patches once we reach conclusions. Hopefully, it will be done within this week.

PR #2456, opened 2023-10-25
Fix triple lint

FinalT:


common/constant/key.go
```diff
@@ -19,6 +19,8 @@ package constant
 
 type DubboCtxKey string
 
+type TripleCtxKey string
```
DMwangnima, line 22: I suggest that moving this type to protocol/triple/triple_protocol/handler_compat.go since Triple protocol should try not to rely on dubbo-go as much as possible.

common/constant/key.go
```diff
@@ -434,3 +436,8 @@ const (
 	DefaultMetaFileName  = "dubbo.metadata."
 	DefaultEntrySize     = 100
 )
+
```
DMwangnima, line 439: Ditto.

PR #2463, opened 2023-10-27
Update metrics and tracing configs

chickenlj:
followup of #2447


compat.go
```diff
@@ -273,24 +268,32 @@ func compatMetricConfig(c *global.MetricConfig) *config.MetricConfig {
...
+		Enable:             c.Enable,
+		Port:               c.Port,
+		Path:               c.Path,
+		Prometheus:         compatMetricPrometheusConfig(c.Prometheus),
+		Aggregation:        compatMetricAggregationConfig(c.Aggregation),
+		Protocol:           c.Protocol,
+		EnableMetadata:     c.EnableMetadata,
+		EnableRegistry:     c.EnableRegistry,
+		EnableConfigCenter: c.EnableConfigCenter,
+		EnableRpc:          c.EnableRpc,
```
FoghostCn, line 280: Shouldn’t EnableRpc be deleted?

chickenlj: Yeah, but I worry it will introduce new issues. So we can do that in a separate patch later.

FoghostCn: OK

FoghostCn (approved):
LGTM

PR #2465, opened 2023-10-27
refactor: metrics config

### LINKED ISSUE #2464: MetricsConfig has too much enable fields
目前，MetricsConfig 下有太多的 enable 控制字段
9b22e045bbc669f0b28d1cde1904ac3b
可以修改为如下策略：
1. Enable，Enable=true 则开启 rpc metrics，默认 false
2. EnableRegistry/EnableMetadata/EnableConfigCenter，Enable=true 后依然默认关闭，通过 EnableRegistry=true 来打开
3. Prometheus.Exporter.Enable，Enable=true 后默认打开
4. 移除 EnableRpc，RPC 应该是 metrics 功能启用后，dubbogo 提供的基本可观测能力

ev1lQuark:
Resolves #2464

### DIFF
--- a/config/metric_config.go
+++ b/config/metric_config.go
@@ -133,7 +133,6 @@ func (mc *MetricConfig) DynamicUpdateProperties(newMetricConfig *MetricConfig) {
 // prometheus://localhost:9090?&histogram.enabled=false&prometheus.exporter.enabled=false
 func (mc *MetricConfig) toURL() *common.URL {
 	url, _ := common.NewURL("localhost", common.WithProtocol(mc.Protocol))
-	url.SetParam(constant.PrometheusExporterEnabledKey, strconv.FormatBool(*mc.Enable)) // for compatibility
 	url.SetParam(constant.PrometheusExporterMetricsPortKey, mc.Port)
 	url.SetParam(constant.PrometheusExporterMetricsPathKey, mc.Path)
 	url.SetParam(constant.ApplicationKey, mc.rootConfig.Application.Name)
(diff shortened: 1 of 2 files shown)


chickenlj (approved):
LGTM.

config/metric_config.go
```diff
@@ -60,7 +59,7 @@ type PrometheusConfig struct {
 }
 
 type Exporter struct {
-	Enabled *bool `default:"false" yaml:"enabled" json:"enabled,omitempty" property:"enabled"`
+	Enabled *bool `default:"true" yaml:"enabled" json:"enabled,omitempty" property:"enabled"`
```
FoghostCn, line 62: this will cause `url.getParam(constant.PrometheusExporterEnabledKey)` allways true, see https://github.com/apache/dubbo-go/blob/main/config/metric_config.go#L157

ev1lQuark: I think the `toURL()` method should not add `&&` or `||` to change enabled logic, instead, only performs a format conversion operation.

FoghostCn: `metrics.Init(mc.toURL())` executed only when mc.Enable is true , I have no problem here.

FoghostCn (approved):
LGTM

config/metric_config.go
FoghostCn: `metrics.Init(mc.toURL())` will executed only when `mc.Enable` is true, so I have no problem here.

ev1lQuark: Got it, thanks.

PR #2467, opened 2023-10-28
[3.0] fix(jsonrpc): Limit header size to avoid unexpected OOM

justxuewei:
This patch limits the max size of header to 8Mib to avoid OOM issues.


chickenlj (approved):
LGTM.

PR #2473, opened 2023-10-28
feat：add triple healthcheck server

FinalT:
Add health check server for the triple service


DMwangnima (commented):
Nice work! I'll consider the pros and cons of exposing ```SetProServices```. And I would give you detailed feedback tomorrow.

chickenlj (approved):
LGTM.

chickenlj:
> Nice work! I'll consider the pros and cons of exposing `SetProServices`. And I would give you detailed feedback tomorrow.

Yeah. We need to provide a higher-level API that can shield the details like ServiceDefinition from end users. This might need help of proto-gen-go-triple.

PR #2483, opened 2023-11-03
feat:sort triple logic

FinalT:
1.add triple RefreshService
2.move export service to server.Serve()

related issue #2477


chickenlj (approved):
LGTM.

protocol/triple/server.go
```diff
@@ -144,6 +147,44 @@ func (s *Server) Start(invoker protocol.Invoker, info *server.ServiceInfo) {
 	}()
 }
 
+// RefreshService refreshes Triple Service
+func (s *Server) RefreshService(invoker protocol.Invoker, info *server.ServiceInfo) {
```
DMwangnima, line 151: Maybe we could define a common function to initialize ```hanOpts``` since there are much repeated logic.

DMwangnima:
LGTM.

PR #2484, opened 2023-11-04
feat: support configuring client.WithURL without protocol prefix

DMwangnima:
This PR aims to support another type of configuring ```client.WithURL```:
```go
client.WithURL("127.0.0.1:20000")

client.WithURL("tri://127.0.0.1:20000")
```
Both of them could work well.

### DIFF
--- a/protocol/triple/internal/client/cmd_instance/main.go
+++ b/protocol/triple/internal/client/cmd_instance/main.go
@@ -36,7 +36,7 @@ func main() {
 	}
 	// configure the params that only client layer cares
 	cli, err := ins.NewClient(
-		client.WithURL("tri://127.0.0.1:20000"),
+		client.WithURL("127.0.0.1:20000"),
 	)
 	if err != nil {
 		panic(err)
--- a/client/action.go
+++ b/client/action.go
@@ -135,44 +135,8 @@ func (opts *ClientOptions) refer(srv common.RPCService, info *ClientInfo) {
 	updateOrCreateMeshURL(opts)
 
 	// retrieving urls from config, and appending the urls to opts.urls
-	if ref.URL != "" { // use user-specific urls
-		/*
-			 Two types of URL are allowed for opts.URL:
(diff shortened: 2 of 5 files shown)


protocol/triple/internal/client/cmd_instance/main.go
```diff
@@ -36,7 +36,7 @@ func main() {
 	}
 	// configure the params that only client layer cares
 	cli, err := ins.NewClient(
-		client.WithURL("tri://127.0.0.1:20000"),
+		client.WithURL("127.0.0.1:20000"),
```
chickenlj, line 39: How will the protocol going to be used determined if no protocol is set?

chickenlj: So it will check `ref.Protocol`? What if `ref.Protocol` is also not specified.

DMwangnima: Client would use Triple protocol by default. Actually, the ideal configuration form would be like:
```
cli, err := ins.NewClient(
    client.WithURL("127.0.0.1:20000"),
    client.WithTriple(),
)
```
Since we decided to make Triple as default protocol, there is no need to provide ```client.WithTriple``` to users. Then, there would be other protocols that could be supported by new API in the future, we would provide ```client.WithXXX``` to users.

DMwangnima: > So it will check `ref.Protocol`? What if `ref.Protocol` is also not specified.
```global.DefaultReferenceConfig``` has set triple as default protocol.
By the way, that is the problem we need to talk. **What is the correct location that we should assign default values?**
For instance, we set default protocol in **global.ReferenceConfig**. But it is too much granularity of impact. I think the right to assign default values should be in the corresponding module.

chickenlj: I see. I will create an issue for this to make sure we can get back to this later.

chickenlj (approved):
LGTM.

Issue #2479, opened 2023-10-31
Remaining todos for dubbo-go v3.2


DMwangnima:
Blocking todos:
1. go version upgrading
2. protoc-gen-triple and protoc-gen-go-triple
Update the name of protoc-gen-triple to protoc-gen-go-triple and make it support generating old triple stub code. Then we can tag it with a higher version.
3. Triple supports sending and receiving Headers and Trailers
4. Triple adaptation layer supports old triple client stub code
5. Samples and Docs

Non-blocking todos:
1. Tracing interceptor for Triple adaptation layer
2. TLS support of dubbo-go and Triple adaptation layer
3. Exposed functionalities of Server layer

chickenlj:
Delete 'protocol/dubbo3' and 'protocol/grpc'

PR #2489, opened 2023-11-07
feat: update protoc-gen-triple

FinalT:
1.rename protoc-gen-triple to protoc-gen-go-triple
2.support generating old triple stub code
3.tag a higher version
related issues #2479


protocol/triple/triple-tool/protoc-gen-go-triple/main.go
```diff
@@ -51,19 +53,41 @@ func main() {
 		os.Exit(1)
 	}
 
-	protogen.Options{}.Run(
+	var flags flag.FlagSet
+	useOld := flags.Bool("useOldVersion", false, "print the version and exit")
```
DMwangnima, line 57: @chickenlj We need to talk about this flag to make it more intuitive. And we use ```go-triple_opt=useOldVersion=true``` as temporary placeholder.

chickenlj: `go-triple_opt=useOldVersion=true`

This style looks good to me as it's the same as `go-grpc_opt=xxx`

protocol/triple/triple-tool/protoc-gen-go-triple/main.go
```diff
@@ -51,19 +53,41 @@ func main() {
...
-			for _, file := range plugin.Files {
-				if file.Generate {
-					tripleGo, err := generator.ProcessProtoFile(file.Proto)
-					if err != nil {
-						return err
-					}
-					return generator.GenTripleFile(tripleGo)
-				}
+			if *useOld {
+				return genOldTriple(plugin)
```
chickenlj, line 66: Will the generated `oldTriple` sources be put in the same place and given the same package name as in earlier versions?

FinalT: Keep the same as before

chickenlj (approved):
LGTM.

PR #2494, opened 2023-11-08
feat:support dubbo.load()

FinalT:


dubbo.go
```diff
@@ -137,3 +145,63 @@ func (ins *Instance) NewServer(opts ...server.ServerOption) (*server.Server, err
 	}
 	return srv, nil
 }
+
+func (ins *Instance) Start() (err error) {
```
DMwangnima, line 149: There is no need to expose this method?

dubbo.go
```diff
@@ -137,3 +145,63 @@ func (ins *Instance) NewServer(opts ...server.ServerOption) (*server.Server, err
...
+		if err = ins.LoadProvider(); err != nil {
+			return
+		}
+	})
+	return err
+}
+
+// LoadProvider loads the service provider.
+func (ins *Instance) LoadProvider() error {
+	srv, err := ins.NewServer()
```
DMwangnima, line 164: Just consider what options should be passed to ```NewServer``` and ```srv.Register``` referring to ```ProviderConfig``` and ```ReferenceConfig```.

dubbo.go
```diff
@@ -137,3 +145,63 @@ func (ins *Instance) NewServer(opts ...server.ServerOption) (*server.Server, err
...
+		if err = srv.Register(definition.Handler, definition.Info, definition.Opts...); err != nil {
+			return err
+		}
+	}
+	return srv.Start()
+}
+
+// LoadConsumer loads the service consumer.
+func (ins *Instance) LoadConsumer() error {
+	cli, err := ins.NewClient()
```
DMwangnima, line 179: ditto. Refer to ```ConsumerConfig``` and "ReferenceConfig".

dubbo.go
```diff
@@ -137,3 +145,63 @@ func (ins *Instance) NewServer(opts ...server.ServerOption) (*server.Server, err
...
+	if err != nil {
+		return err
+	}
+	// refer services
+	for _, definition := range consumerServices {
+		if err = cli.Init(definition.Info); err != nil {
+			return err
+		}
+		definition.Info.ClientInjectFunc(definition.Svc, cli)
```
DMwangnima, line 188: Excellent!

server/server.go
```diff
@@ -181,6 +182,15 @@ func (s *Server) Serve() error {
 	select {}
 }
 
+func (s *Server) Start() error {
+	if err := s.exportServices(); err != nil {
+		return err
+	}
+	metadata.ExportMetadataService()
+	registry_exposed.RegisterServiceInstance(s.cfg.Application.Name, s.cfg.Application.Tag, s.cfg.Application.MetadataType)
```
DMwangnima, line 190: This part is used for what?

dubbo.go
```diff
@@ -137,3 +145,63 @@ func (ins *Instance) NewServer(opts ...server.ServerOption) (*server.Server, err
...
+		}
+		if err = ins.LoadProvider(); err != nil {
+			return
+		}
+	})
+	return err
+}
+
+// LoadProvider loads the service provider.
+func (ins *Instance) LoadProvider() error {
```
DMwangnima, line 163: ditto.

dubbo.go
```diff
@@ -137,3 +145,63 @@ func (ins *Instance) NewServer(opts ...server.ServerOption) (*server.Server, err
...
+	for _, definition := range providerServices {
+		if err = srv.Register(definition.Handler, definition.Info, definition.Opts...); err != nil {
+			return err
+		}
+	}
+	return srv.Start()
+}
+
+// LoadConsumer loads the service consumer.
+func (ins *Instance) LoadConsumer() error {
```
DMwangnima, line 178: ditto.

server/server.go
chickenlj: I have the same question, we already have `server.Serve()`. And also `dubbo.Start`, `dubbo.LoadProvider`,`dubbo.LoadConsumer` are not necesary.

dubbo.go
FinalT: Yeah, I don't think it needs to be exposed

server/server.go
FinalT: Non-blocking start, same as the original

FinalT: I need a func to start without blocking

chickenlj: How about you we make the new method private and delegate the main logic of `Serve` to this method by calling `ins.start()`?

DMwangnima: ```config.Load``` is non-blocking. We could put ```server.Serve()``` in a goroutine to make ```dubbo.Load()``` non-blocking. Then there would be enough space for us to consider whether introducing a non-blocking method ```server.Start()``` is necessary.

PR #2502, opened 2023-11-13
Triple client&server api

chickenlj:
1. ConsumerConfig and ProviderConfig serve as the default configuration across services.
```go
type ConsumerConfig struct {
	ReferenceConfig
}
```

```go
type ProviderConfig struct {
	ServiceConfig
}
```

2. Both the client and server have the same configuration hierarchy.
* dubbo.NewInstance(InstanceOption)
* instance.NewServer(ServerOption) / instance.NewClient(ClientOption)
* server.RegisterService(ServiceOption) / client.GetGreetService(ReferenceOption)
* greetService.greet(CallOption)


client/options.go
```diff
@@ -131,239 +130,542 @@ func (cliOpts *ClientOptions) init(opts ...ClientOption) error {
...
+	}
+}
+
+func WithProtocolJsonRPC() ReferenceOption {
+	return func(opts *ReferenceOptions) {
+		opts.Reference.Protocol = "jsonrpc"
+	}
+}
+
+func WithProtocol(protocol string) ReferenceOption {
```
DMwangnima, line 356: Since we have enumerated protocols that we support, maybe there is no need to expose this API?

client/options.go
```diff
@@ -131,239 +130,542 @@ func (cliOpts *ClientOptions) init(opts ...ClientOption) error {
...
+func WithClientProtocolJsonRPC() ClientOption {
 	return func(opts *ClientOptions) {
-		opts.Reference.MeshProviderPort = port
+		opts.Consumer.Protocol = "jsonrpc"
 	}
 }
 
-// ---------- For framework ----------
-// These functions should not be invoked by users
+func WithClientProtocol(protocol string) ClientOption {
```
DMwangnima, line 644: ditto.

client/options_test.go
```diff
@@ -17,49 +17,37 @@
...
-	}
-
-	for _, test := range tests {
-		newOpts := defaultClientOptions()
-		assert.Nil(t, newOpts.init(test.opts...))
-		assert.Nil(t, newOpts.processURL(&common.URL{}))
-		test.justify(t, newOpts)
-	}
-}
+//func TestWithURL(t *testing.T) {
```
DMwangnima, line 20: After merging this PR, we can modify this test file.

common/dubboutil/copier.go
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package dubboutil
+
+import (
+	"reflect"
+)
+
+func CopyFields(sourceValue reflect.Value, targetValue reflect.Value) {
```
DMwangnima, line 24: Java-Style but very efficient code :)

client/client.go
```diff
@@ -45,7 +47,7 @@ type ClientInfo struct {
 	Meta             map[string]interface{}
 }
 
-func (cli *Client) call(ctx context.Context, paramsRawVals []interface{}, interfaceName, methodName, callType string, opts ...CallOption) (protocol.Result, error) {
+func (cli *Client) call(ctx context.Context, paramsRawVals []interface{}, interfaceName, methodName, group, version, callType string, opts ...CallOption) (protocol.Result, error) {
```
DMwangnima, line 50: We need to consider that whether ```group``` and ```version``` should be required fields. 
Since this function is relevant to non-idl API, we need to keep it simple.
For instance, if users do not need to configure ```group``` and ```version```,
```
cli.CallUnary(ctx, req, resp, interfaceName, methodName, "", "") 
```
Empty strings are very redundant.
I prefer to use ```CallOption``` to carry ```group``` and ```version```.

client/client.go
```diff
@@ -57,52 +59,65 @@ func (cli *Client) call(ctx context.Context, paramsRawVals []interface{}, interf
...
+func (cli *Client) CallBidiStream(ctx context.Context, interfaceName, methodName, group, version string, opts ...CallOption) (interface{}, error) {
+	res, err := cli.call(ctx, nil, interfaceName, methodName, group, version, constant.CallBidiStream, opts...)
 	if err != nil {
 		return nil, err
 	}
 	return res.Result(), res.Error()
 }
 
-func (cli *Client) Init(info *ClientInfo) error {
+func (cli *Client) Init(interfaceName string, info *ClientInfo, opts ...ReferenceOption) (string, string, error) {
```
DMwangnima, line 104: ```ClientInfo``` contains ```InterfaceName```. And we also need to consider about non-idl API style since this function is also important.

client/client.go
chickenlj: Good point. I was also thinking of using CallOption in the first place. I will move group and version to CallOption later.

client/client.go
chickenlj: Got

client/options.go
FoghostCn: What if the user custom protocol?

chickenlj: Yeah, I think we should provide both for flexibility.

client/client.go
chickenlj: Putting group and version into CallOption has other issues. But still it's currently the best way I can think of.

ev1lQuark (approved):
LGTM

protocol/triple/health/triple_health/health.triple.go
```diff
@@ -94,22 +94,30 @@ type HealthImpl struct {
...
+	opts = appendGroupVersion(opts, c)
+	stream, err := c.cli.CallServerStream(ctx, req, "grpc.health.v1.Health", "Watch", opts...)
 	if err != nil {
 		return nil, err
 	}
 	rawStream := stream.(*triple_protocol.ServerStreamForClient)
 	return &HealthWatchClient{rawStream}, nil
 }
 
+func appendGroupVersion(opts []client.CallOption, c *HealthImpl) []client.CallOption {
```
DMwangnima, line 115: What if users passing ```WithGroup``` and ```WithVersion``` directly? Should we set priority or just ignore?

client/options.go
```diff
@@ -267,9 +273,9 @@ func WithLoadBalanceP2C() ReferenceOption {
 	}
 }
 
-func WithLoadBalanceXDSRingHash() ReferenceOption {
+func WithLoadBalance(lb string) ReferenceOption {
```
DMwangnima, line 276: Not only do we need to enumerate, but we also need to leave a generic configuration API?

DMwangnima:
I would merge this PR and do some fine tuning.

client/options.go
chickenlj: Yes, I realize this is necessary when developing samples. This becomes especially useful when considering Dubbo's high extensibility, we need to make sure our API can support user self-extended LoadBalance policies.

protocol/triple/health/triple_health/health.triple.go
chickenlj: Yeah, I have the same concern when implementing this feature.

chickenlj:
> I would merge this PR and do some fine tuning.

Thanks. Please remember to file new issues for todos recognized.

PR #2503, opened 2023-11-13
Fix url merge, merge attributes

chickenlj:
followup of #2484


FinalT (approved):

FinalT:
LGTM.

common/url.go
```diff
@@ -818,6 +818,14 @@ func MergeURL(serviceURL *URL, referenceURL *URL) *URL {
 			mergedURL.Methods[i] = method
 		}
 	}
+
+	// merge attributes
+	if mergedURL.attributes == nil {
+		mergedURL.attributes = make(map[string]interface{}, len(referenceURL.attributes))
+	}
+	for attrK, attrV := range referenceURL.attributes {
+		mergedURL.attributes[attrK] = attrV
```
FinalT, line 827: if mergedURL contains attributes (a1 -> v1) and referenceURL contains attributes (a1 -> v2), mergedURL's attributes will be overridden.I think we should check to see if the key exists before setting

chickenlj: Yes, that's a good point. I will think about the override and priority here.

PR #2504, opened 2023-11-14
Fix url attributes merge

FinalT:
fix pr #2503


chickenlj (approved):
LGTM.

chickenlj (changes requested):
How about we make `MergeURL` a method of *URL?

Because currently, the behaviour or priority of attribute keys is not determined since it depends on the sequence of parameters - the first URL always gets high priority, but not all maintainers know about this rule, so if they don't correctly put these two URLs, the result is probably not what they expect.

DMwangnima:
> How about we make `MergeURL` a method of *URL?
> 
> Because currently, the behaviour or priority of attribute keys is not determined since it depends on the sequence of parameters - the first URL always gets high priority, but not all maintainers know about this rule, so if they don't correctly put these two URLs, the result is probably not what they expect.

+1. We need to distinguish between baseURL and addURL.

chickenlj (approved):
LGTM.

PR #2506, opened 2023-11-17
Completely delete `old triple` and `dubbo3` package

chickenlj:
The `new triple` package now provides full support for `old triple` and `dubbo3`, so they can be deleted now.


FinalT:
The `config` package has some logic that depends on these two packages, We need to remove some logic from conifg before merging

chickenlj:
> The `config` package has some logic that depends on these two packages, We need to remove this logic from conifg before merging

done, please check again. An issue is created for future reference https://github.com/apache/dubbo-go/issues/2514

PR #2507, opened 2023-11-17
Reallocate protoc plugin and polish package documentations

chickenlj:


cmd/protoc-gen-go-triple/gen/cmd/root.go
FinalT, line 1: This file can be deleted

FinalT (approved):

FinalT:
LGTM.

PR #2531, opened 2023-12-02
feat: Adapt config.load

FinalT:
Adapt healthCheck Service and reflection Service


common/constant/key.go
```diff
@@ -389,14 +389,14 @@ const (
 
 // reflection service
 const (
-	ReflectionServiceTypeName  = "DubbogoServerReflectionServer"
-	ReflectionServiceInterface = "grpc.reflection.v1alpha.ServerReflection"
+	ReflectionServiceTypeName  = "ReflectionServer"
+	ReflectionServiceInterface = "dubbo.reflection.v1alpha.ReflectionServer"
```
DMwangnima, line 393: Maybe we should no modify this Interface since it is relevant to http path that this service would expose.

common/constant/key.go
```diff
@@ -389,14 +389,14 @@ const (
...
 )
 
 // healthcheck service
 const (
-	HealthCheckServiceTypeName  = "DubbogoHealthServer"
-	HealthCheckServiceInterface = "grpc.health.v1.Health"
+	HealthCheckServiceTypeName  = "HealthTripleServer"
+	HealthCheckServiceInterface = "dubbo.health.v1.Health"
```
DMwangnima, line 399: ditto.

common/constant/key.go
FinalT: This will conflict with grpc

common/constant/key.go
FinalT: This will conflict with grpc

config/service_config.go
```diff
@@ -311,7 +317,11 @@ func (s *ServiceConfig) Export() error {
 
 			for _, regUrl := range regUrls {
 				setRegistrySubURL(ivkURL, regUrl)
-				invoker := proxyFactory.GetInvoker(regUrl)
+				if info == nil {
```
DMwangnima, line 320: We could extract a function such as ```generateInvoker```.

config/service_config.go
```diff
@@ -329,7 +339,11 @@ func (s *ServiceConfig) Export() error {
 					logger.Warnf("SetMetadataServiceURL error = %s", err)
 				}
 			}
-			invoker := proxyFactory.GetInvoker(ivkURL)
+			if info == nil {
```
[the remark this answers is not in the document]
FinalT: done.

config/service_config.go
FinalT: done.

config/service.go
```diff
@@ -60,6 +61,18 @@ func SetProviderService(service common.RPCService) {
 	proServices[ref] = service
 }
 
+// SetProviderServiceWithInfo is called by init() of implement of RPCService
```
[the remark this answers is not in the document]
FinalT: done

config/service.go
```diff
@@ -34,6 +34,7 @@ var (
 	conServices                  = map[string]common.RPCService{} // service name -> service
 	proServicesLock              = sync.Mutex{}                   // used to guard proServices map
```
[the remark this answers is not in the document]
FinalT: done

PR #2533, opened 2023-12-07
fix：set invoker

FinalT:


DMwangnima:
Unit test is necessary :)

PR #2545, opened 2023-12-25
fix: resolve triple bugs based on dubbo-go-samples

DMwangnima:
This PR aims to fix bugs based on dubbo-go-samples.
It contains:
1. Triple-Hessian2 support
2. InterfaceName override
3. dubbo3 attachment compatibility


chickenlj:
Here's the wrapper definition in the java repo https://github.com/apache/dubbo/blob/3.2/dubbo-rpc/dubbo-rpc-triple/src/main/proto/triple_wrapper.proto

protocol/triple/triple_protocol/codec.go
```diff
@@ -19,6 +19,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	hessian "github.com/apache/dubbo-go-hessian2"
+	"github.com/dubbogo/grpc-go/encoding"
+	"github.com/dubbogo/grpc-go/encoding/proto_wrapper_api"
```
chickenlj, line 24: How about we make a copy of the definitions we needed in `dubbogo/grpc-go`? I think it would be better to not depend on  `dubbogo/grpc-go` except for some compatibility needs.

DMwangnima: I would use new protoc-gen-go-triple to generate new code.

FinalT:
LGTM.

FinalT (approved):

Issue #2537, opened 2023-12-12
dubbogo会为metadata多启一个随机端口，但java版没有这个端口


libaoj1985:
**What would you like to be added**:

**Why is this needed**:

baerwang:
<img width="2141" alt="image" src="https://github.com/apache/dubbo-go/assets/52104949/6221ad1c-e7a8-4859-8d7a-f14ee836407c">
<img width="2201" alt="image" src="https://github.com/apache/dubbo-go/assets/52104949/4730aa19-d523-46f8-99d2-af8edddaf10b">
<img width="769" alt="image" src="https://github.com/apache/dubbo-go/assets/52104949/5fce1f20-5da1-435a-9ed8-773fa678b337">

PR #2548, opened 2023-12-27
fix: metadata port reuse

FinalT:
https://github.com/apache/dubbo-go/issues/2537

### DIFF
--- a/metadata/service/exporter/configurable/exporter.go
+++ b/metadata/service/exporter/configurable/exporter.go
@@ -93,7 +93,7 @@ func getMetadataPort() string {
 		if ok {
 			port = protocolConfig.Port
 		} else {
-			logger.Warnf("[Metadata Service] Dubbo-go %s version's MetadataService does not support triple protocol,"+
+			logger.Warnf("[Metadata Service] Dubbo-go %s version's MetadataService only support dubbo protocol,"+
 				"MetadataService will use random port",
 				dubbo.Version)
 		}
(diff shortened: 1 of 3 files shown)


metadata/service/exporter/configurable/exporter.go
```diff
@@ -84,6 +85,22 @@ func (exporter *MetadataServiceExporter) Export(url *common.URL) error {
...
 
+func getMetadataPort() string {
+	rootConfig := config.GetRootConfig()
+	port := rootConfig.Application.MetadataServicePort
+	if port == "" {
+		protocolConfig, ok := rootConfig.Protocols[constant.DefaultProtocol]
+		if ok {
+			port = protocolConfig.Port
+		} else {
+			logger.Warnf("[Metadata Service] Dubbo-go %s version's MetadataService does not support triple protocol,"+
```
chickenlj, line 96: The warning message would better be "[Metadata Service] Dubbo-go %s version's MetadataService only support dubbo protocol, ".  Because it might be other protocols beyond dubbo and triple.

chickenlj (approved):

metadata/service/exporter/configurable/exporter.go
FinalT: Yes, that's better

chickenlj:
@FoghostCn  PTAL.

chickenlj (approved):
LGTM.

PR #2552, opened 2023-12-30
fix bug #2546

Jafeyyu:
fix bug #2546, change timeout of DubboInvoker from Milliseconds to Nanoseconds

### DIFF
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -176,7 +176,7 @@ func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration 
 			timeout, _ = time.ParseDuration(mTimeout)
 		}
 		// set timeout into invocation
-		ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Milliseconds())))
+		ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Nanoseconds())))
 	}
 	return timeout
 }
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -175,9 +175,9 @@ func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration 
 		if len(mTimeout) != 0 {
 			timeout, _ = time.ParseDuration(mTimeout)
 		}
-		// set timeout into invocation
-		ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Nanoseconds())))
 	}
+	// set timeout into invocation
+	ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Nanoseconds())))
 	return timeout
 }
 
--- a/protocol/dubbo3/dubbo3_invoker.go
+++ b/protocol/dubbo3/dubbo3_invoker.go
@@ -239,12 +239,12 @@ func (di *DubboInvoker) getTimeout(invocation *invocation_impl.RPCInvocation) ti
 	if len(timeout) != 0 {
 		if t, err := time.ParseDuration(timeout); err == nil {
 			// config timeout into attachment
-			invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(t.Milliseconds())))
+			invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(t.Nanoseconds())))
 			return t
 		}
 	}
 	// set timeout into invocation at method level
-	invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(di.timeout.Milliseconds())))
+	invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(di.timeout.Nanoseconds())))
 	return di.timeout
 }
 
--- a/protocol/dubbo3/dubbo3_invoker.go
+++ b/protocol/dubbo3/dubbo3_invoker.go
@@ -239,12 +239,12 @@ func (di *DubboInvoker) getTimeout(invocation *invocation_impl.RPCInvocation) ti
 	if len(timeout) != 0 {
 		if t, err := time.ParseDuration(timeout); err == nil {
 			// config timeout into attachment
-			invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(t.Nanoseconds())))
+			invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(t.Milliseconds())))
 			return t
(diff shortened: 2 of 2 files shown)


protocol/dubbo/dubbo_invoker.go
```diff
@@ -176,7 +176,7 @@ func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration
 			timeout, _ = time.ParseDuration(mTimeout)
 		}
 		// set timeout into invocation
-		ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Milliseconds())))
+		ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Nanoseconds())))
```
DMwangnima, line 179: I think we should move this statement to the end of this function, like this:
```
func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration {
	timeout := di.timeout                                                //default timeout
	if attachTimeout, ok := ivc.GetAttachment(constant.TimeoutKey); ok { //check invocation timeout
		timeout, _ = time.ParseDuration(attachTimeout)
	} else { // check method timeout
		methodName := ivc.MethodName()
		if di.GetURL().GetParamBool(constant.GenericKey, false) {
			methodName = ivc.Arguments()[0].(string)
		}
		mTimeout := di.GetURL().GetParam(strings.Join([]string{constant.MethodKeys, methodName, constant.TimeoutKey}, "."), "")
		if len(mTimeout) != 0 {
			timeout, _ = time.ParseDuration(mTimeout)
		}
	}
        // set timeout into invocation
	ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Milliseconds())))
	return timeout
}
```

DMwangnima: Btw, I think the priority of timeout configuration is Method Level > Service Level > Consumer Level. Is it the same with dubbo? @chickenlj 
If yes, we should modify the logic here. Since the priority of Service Level Timeout is great than Method Level Timeout here.

DMwangnima (changes requested):

protocol/dubbo3/dubbo3_invoker.go
```diff
@@ -239,12 +239,12 @@ func (di *DubboInvoker) getTimeout(invocation *invocation_impl.RPCInvocation) ti
 	if len(timeout) != 0 {
 		if t, err := time.ParseDuration(timeout); err == nil {
 			// config timeout into attachment
-			invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(t.Milliseconds())))
+			invocation.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(t.Nanoseconds())))
```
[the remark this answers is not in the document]
Jafeyyu: I only use Dubbo protocol, so I'm not sure if I need to modify this part

protocol/dubbo/dubbo_invoker.go
Jafeyyu: > I think we should move this statement to the end of this function, like this:
> 
> ```
> func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration {
> 	timeout := di.timeout                                                //default timeout
> 	if attachTimeout, ok := ivc.GetAttachment(constant.TimeoutKey); ok { //check invocation timeout
> 		timeout, _ = time.ParseDuration(attachTimeout)
> 	} else { // check method timeout
> 		methodName := ivc.MethodName()
> 		if di.GetURL().GetParamBool(constant.GenericKey, false) {
> 			methodName = ivc.Arguments()[0].(string)
> 		}
> 		mTimeout := di.GetURL().GetParam(strings.Join([]string{constant.MethodKeys, methodName, constant.TimeoutKey}, "."), "")
> 		if len(mTimeout) != 0 {
> 			timeout, _ = time.ParseDuration(mTimeout)
> 		}
> 	}
>         // set timeout into invocation
> 	ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Milliseconds())))
> 	return timeout
> }
> ```

在java服务端中，获取超时时间是这样处理的
```java
        Long timeout = RpcUtils.convertToNumber(invocation.getObjectAttachmentWithoutConvert(TIMEOUT_KEY));

        if (timeout == null) {
            timeout = (long) invoker.getUrl()
                    .getMethodPositiveParameter(RpcUtils.getMethodName(invocation), TIMEOUT_KEY, DEFAULT_TIMEOUT);
        }
```
是先从attachment中获取，如果timeout为空，则获取方法级的超时，都取不到，使用默认值
所以按照这个逻辑，attachment中的timeout是可以为空的，`ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Milliseconds())))`是否要放到最后，就取决于是否要将DubboInvoker的默认超时写入到attachment中，因为我刚接触dubbo-go才几天时间，所以这块我还没有深入了解，至于优先级，那就只能请各位大佬讨论一下，尽量保持java和go的优先级是一致的，谢谢

DMwangnima: Based on dubbo processing logic, there is no problem with priority issue.

I think we should move ```ivc.SetAttachment(constant.TimeoutKey, strconv.Itoa(int(timeout.Milliseconds())))``` to the end. 

If users pass ```Reference.RequestTimeout``` directly, ```getTimeout``` could retrieve attachTimeout successfully, but it does not overwrite the value corresponding to ```constant.TimeoutKey``` with the parsed timeout value.

Jafeyyu: 我对这样修改时没有意见的，但是这个与我提的bug #2546 已经不是一个问题了，是需要我直接按照你说的进行修改，还是后续再提交PR进行修改呢

DMwangnima: It is okay. I would add additional notes to this issue.

Jafeyyu: I have updated the PR

DMwangnima:
LGTM. cc @chickenlj

DMwangnima (approved):

protocol/dubbo/dubbo_invoker.go
chickenlj: > Btw, I think the priority of timeout configuration is Method Level > Service Level > Consumer Level. Is it the same with dubbo? 

In apache/dubbo, Consumer Level timeout always has higher priority than Provider Level timeout, so the first rule is `Consumer Level > Provider Level`.  Both Consumer and Provider have Method Level and Service Level timeout, so for both sides there's the second rule - `Method Level  > Service Level`.

chickenlj (approved):
LGTM.

PR #2557, opened 2024-01-03
update protoc-gen-go-triple

FinalT:
Optimized file generation logic
Keep pb files and triple files in the same directory
Update README


chickenlj:
What's this for? Please provide more details about this change.

chickenlj (approved):
LGTM.

chickenlj:
We need to update the `generated files` in apache/dubbo-go-samples after this patch is merged.

FinalT:
> We need to update the `generated files` in apache/dubbo-go-samples after this patch is merged.

I will update the generated files after this patch is merged.

Issue #2558, opened 2024-01-03
I can not remove this waring.


majintao:
2024-01-03 18:47:45.742	WARN	tag/router.go:87	[-]	config center does not start, please check if the configuration center has been properly configured in dubbogo.yml%!(EXTRA []interface {}=[])	{"SW_CTX": "[xf-app,639f89eeaa2511eebcf552f3f496997d@192.168.24.77,N/A,N/A,-1]"}

majintao:
func Start(conf Conf) error {
	if conf.LogLevel == "" {
		conf.LogLevel = "error"
	}
	logger.SetLoggerLevel(conf.LogLevel)

	applicationConfig := config.NewApplicationConfigBuilder().SetName(conf.AppName).Build()

	protocolConfigs := make(map[string]*config.ProtocolConfig, 0)
	if conf.ProtocolConfig.Name != "" {
		// 根据配置设置启动参数
		clientConfig := getty.GetDefaultClientConfig()
		params := conf.ProtocolConfig.Params
		if params.GrPoolSize != 0 {
			clientConfig.GrPoolSize = params.GrPoolSize
		}

		if params.ConnectionNumber != 0 {
			clientConfig.ConnectionNum = params.ConnectionNumber
		}

		if params.MaxMsgLen != 0 {
			clientConfig.GettySessionParam.MaxMsgLen = params.MaxMsgLen
		}

		if params.SessionName != "" {
			clientConfig.GettySessionParam.SessionName = params.SessionName
		}

		dubboPort := strconv.Itoa(conf.ProtocolConfig.Port)
		if dubboPort == "" {
			dubboPort = "-1"
		}
		dubboProtocolConfig := config.NewProtocolConfigBuilder().
			SetName(conf.ProtocolConfig.Name).SetPort(dubboPort).SetParams(clientConfig).Build()
		protocolConfigs[conf.ProtocolConfig.Name] = dubboProtocolConfig
	} else {
		// 默认配置
		defaultProtocolName := "dubbo"
		defaultDubboProtocolConfig := config.NewProtocolConfigBuilder().
			SetName(defaultProtocolName).SetPort("-1").Build()
		protocolConfigs[defaultProtocolName] = defaultDubboProtocolConfig
	}

	consumerConfig := config.NewConsumerConfigBuilder().SetCheck(false).SetReferences(References).Build()
	providerConfig := config.NewProviderConfigBuilder().SetServices(Services).Build()

[...]

PR #2559, opened 2024-01-04
Fix:modify router warn log

FinalT:
related #2558
I think if users don't set config center ,we should print a info log rather than a warning, telling the user that the router will not be enabled


FoghostCn:
I see more warn log, can you fix it all
```
WARN    tag/router.go:87        config center does not start, please check if the configuration center has been properly configured in dubbogo.yml
WARN    condition/dynamic_router.go:127 config center does not start, please check if the configuration center has been properly configured in dubbogo.yml
WARN    condition/dynamic_router.go:176 condition router get providerApplication is empty, will not subscribe to provider app rules.
```

chickenlj (approved):
LGTM.

Issue #2550, opened 2023-12-28
客户端方法重试次数跟 java 实现不一样


zhaoyunxing92:
### Environment

- Server: Dubbo-go, v3.1.0
- Client: Dubbo-go, v3.1.0
- Protocol: dubbo
- Registry: zk
- 
### Issue description

两边实现不一样导致客户端重试计算出来值不一样
## Java

```java
    private int calculateInvokeTimes(String methodName) {
        int len = getUrl().getMethodParameter(methodName, RETRIES_KEY, DEFAULT_RETRIES) + 1;
        RpcContext rpcContext = RpcContext.getClientAttachment();
        Object retry = rpcContext.getObjectAttachment(RETRIES_KEY);
        if (retry instanceof Number) {
            len = ((Number) retry).intValue() + 1;
            rpcContext.removeAttachment(RETRIES_KEY);
        }
        if (len <= 0) {
            len = 1;
        }

        return len;
    }
```
## Go

```go
func getRetries(invokers []protocol.Invoker, methodName string) int {
	if len(invokers) <= 0 {
		return constant.DefaultRetriesInt
	}

	url := invokers[0].GetURL()
	// get reties
	retriesConfig := url.GetParam(constant.RetriesKey, constant.DefaultRetries)
	// Get the service method loadbalance config if have
	if v := url.GetMethodParam(methodName, constant.RetriesKey, ""); len(v) != 0 {
		retriesConfig = v
	}

	retries, err := strconv.Atoi(retriesConfig)
	if err != nil || retries < 0 {
		logger.Error("Your retries config is invalid,pls do a check. And will use the default retries configuration instead.")
		retries = constant.DefaultRetriesInt
	}

	if retries > len(invokers) {
		retries = len(invokers)
	}
	return retries
}
```

chickenlj:
Hi @zhaoyunxing92, would you like to fix this issue?

PR #2564, opened 2024-01-04
fix: retry time set(#2550)

zhaoyunxing92:


Lvnszn (approved):

chickenlj (approved):
LGTM.

justxuewei:
Does this patch need to backport to 3.1 branch?

cluster/cluster/failover/cluster_invoker.go
```diff
@@ -114,23 +113,7 @@ func getRetries(invokers []protocol.Invoker, methodName string) int {
...
-		logger.Error("Your retries config is invalid,pls do a check. And will use the default retries configuration instead.")
-		retries = constant.DefaultRetriesInt
-	}
-
-	if retries > len(invokers) {
-		retries = len(invokers)
-	}
-	return retries
+	return url.GetMethodParamIntValue(methodName, constant.RetriesKey,
+		url.GetParamByIntValue(constant.RetriesKey, constant.DefaultRetriesInt))
```
chickenlj, line 118: Should we check the case when `retries` is set as negative?

chickenlj (changes requested):

Issue #2565, opened 2024-01-06
Solving Circular Dependency in Dubbo-Go


FinalT:
**Why is this needed**:
This issue was brought up by #2539. Although the export logic of the server layer was separated, the generated triple files still depend on the server layer. This means that our internal service module still depends on the server layer, and our underlying modules depend on these internal services. As the server layer is an upper-level module, it depends on these lower-level modules, resulting in a circular dependency.

**What would you like to be added**:
However, the underlying modules typically only use a few interfaces of the internal services, and the number of internal services is not large. It is not necessary to depend on all packages of the internal services. One possible solution is to create an "internal" package as the top-level module, where the service methods are defined. Then, the internal services can still be registered through the server layer, and the methods can be further registered in the `internal` package. This way, other modules do not need to directly depend on the internal services.

Refer to grpc-go to resolve circular dependencies

chickenlj:
There're some terms that I cannot understand well, `internal service`, `underlying service`, `internal package`, `server layer`. Could you please explain the part in dubbo-go they stand for, maybe giving an example for each of them?

> One possible solution is to create an "internal" package as the top-level module, where the service methods are defined. Then, the internal services can still be registered through the server layer, and the methods can be further registered in the internal package. This way, other modules do not need to directly depend on the internal services.

For this solution, could you please give a simple example of maybe how it solves the specific issue in https://github.com/apache/dubbo-go/issues/2539?

FinalT:
Take healthcheck for example,Here are the current dependencies between modules,It's easy to make circular dependencies if other modules need to make changes to the service health state in the future
![image](https://github.com/apache/dubbo-go/assets/125108794/d2f331d2-a98d-482a-acef-2564e187d5e0)
This is the relationship between modules after transformation
![image](https://github.com/apache/dubbo-go/assets/125108794/65df24d2-0a0a-4930-826b-40edf846d1b7)

PR #2576, opened 2024-01-12
feat:Solving Circular Dependency in Dubbo-Go

FinalT:
refer #2565

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -27,9 +27,9 @@ import (
 var (
 	// HealthSetServingStatusServing is used to set service serving status
 	// the initialization place is in /protocol/triple/health/healthServer.go
-	HealthSetServingStatusServing func(service string)
+	HealthSetServingStatusServing = func(service string) {}
 	// ReflectionRegister is used to register reflection service provider
 	// the initialization place is in /protocol/triple/reflection/serverreflection.go
-	ReflectionRegister func(reflection reflection.ServiceInfoProvider)
+	ReflectionRegister = func(reflection reflection.ServiceInfoProvider) {}
 	// todo: add metadata func
 )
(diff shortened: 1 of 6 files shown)


internal/internal.go
```diff
@@ -0,0 +1,33 @@
...
+// except internal symbols to avoid circular dependencies.
+package internal
+
+import (
+	"dubbo.apache.org/dubbo-go/v3/internal/reflection"
+)
+
+var (
+	// HealthSetServingStatusServing is used to set service serving status
+	HealthSetServingStatusServing = func(service string) {}
```
DMwangnima, line 29: ```HealthSetServingStatusServing func(service string)``` is better? Since ```func(service string) {}``` would let others think this part has been initialized.

internal/internal.go
```diff
@@ -0,0 +1,33 @@
...
+// the top-level dubbo-go package.  It must not import any dubbo-go symbols
+// except internal symbols to avoid circular dependencies.
+package internal
+
+import (
+	"dubbo.apache.org/dubbo-go/v3/internal/reflection"
+)
+
+var (
+	// HealthSetServingStatusServing is used to set service serving status
```
DMwangnima, line 28: Add a comment to specify the initialization place is in /protocol/triple/health/...

internal/internal.go
```diff
@@ -0,0 +1,33 @@
...
+	HealthSetServingStatusServing = func(service string) {}
+	// ReflectionRegister is used to register reflection service provider
```
DMwangnima, line 30: ditto.

internal/internal.go
```diff
@@ -0,0 +1,33 @@
...
+	ReflectionRegister = func(reflection reflection.ServiceInfoProvider) {}
```
DMwangnima, line 31: ditto.

internal/internal.go
FinalT: These services are not required, so the default implementation is empty to avoid affecting the tests of other modules.

DMwangnima (approved):
LGTM.

Issue #2430, opened 2023-09-20
goroutine scheduling issues in service offline scenarios


DMwangnima:
### Environment

- Server: Dubbo-go, v3.0.1
- Client: Dubbo-go, v3.0.1
- Protocol: dubbo
- Registry: zookeeper

### Issue description
**Invoking Process:**
1. ZookeeperServiceDiscovery.DataChange =>
(1) GetInstances
Get the latest instance of this service. Because of the service not actually calling **UnRegister** (refer to graceful_shutdown) when the service shuts down, but waits for the ephemeral node to be removed by the zookeeper node, there is a probability of obtaining the offline instance.
(2) ServiceInstanceChangedListenerImpl.OnEvent
Callback. Returning instances to ServiceInstanceChangedListernerImpl.

2. ServiceInstanceChangedListenerImpl.OnEvent =>
(1) RegistryDirectory.NotifyAll
Callback. Returning urls of instances to RegistryDirectory

3. RegistryDirectory.NotifyAll
<img width="738" alt="image" src="https://github.com/apache/dubbo-go/assets/33331974/4f4e2d00-7367-47ff-9d0c-2e1b5aec0e35">

**refreshAllInvokers** is executed asynchronously, in which the assembly of the invoker is performed. In a large-scale offline scenario, the offline of each instance will trigger the entire process. **NotifyAll** is also called several times, and the order of calls between multiple **refreshAllInvokers** cannot be guaranteed. There may be calls to offline services.

This bug refers to https://github.com/apache/dubbo-go/issues/2328.

### Logs

<details><summary>Click me to check logs</summary>

```
Copy logs to here.
```

</details>

PR #2584, opened 2024-01-18
fix:call offline services

FinalT:
refer #2430
In the case that a large number of services go offline in a short period of time, the dubbo service will call the offline service
This is a temporary solution, we will later reproduce the scenario and test its performance


chickenlj (approved):
LGTM.

PR #2589, opened 2024-01-30
feat: inject attachments for triple

DMwangnima:


protocol/triple/server.go
```diff
@@ -409,3 +422,14 @@ func createServiceInfoWithReflection(svc common.RPCService) *server.ServiceInfo
...
+// generateAttachments transfer http.Header to map[string]interface{} and make all keys lowercase
+func generateAttachments(header http.Header) map[string]interface{} {
+	attachments := make(map[string]interface{}, len(header))
+	for key, val := range header {
+		lowerKey := strings.ToLower(key)
+		attachments[lowerKey] = val
+	}
+
+	return attachments
+}
```
FoghostCn, line 435: Will there be some HTTP protocol headers here? Do we need to filter them out?

DMwangnima: Old triple would expose HTTP protocol headers too.

FoghostCn: ok

Issue #2335, opened 2023-06-09
MetricConfig 无法设置 Enable 为 false


jiangy:
### Environment

- Server: Dubbo-go, v3.0.5
- Client: 
- Protocol: 
- Registry: 

### Issue description
这个 Enable 的默认值处理有问题，如果设置为 false 的话会当作 0 值被 defaults 设置为默认值 true，也就是说用户永远无法设置 false

![image](https://github.com/apache/dubbo-go/assets/2594973/f078c9aa-e863-46f7-a950-e8804f0c8e28)

### Logs

<details><summary>Click me to check logs</summary>

```
Copy logs to here.
```

</details>

ev1lQuark:
Thank you, this is the [PR](https://github.com/apache/dubbo-go/pull/2310/files#diff-b302e39f38f857c8e3d9d38c2b7329c70ab2408d249f82987790af3dfd9cfd35) that fixes the bug. But it hasn't been released yet.

ev1lQuark:
This [PR](https://github.com/apache/dubbo-go/pull/2336) resolve the bug.

justxuewei:
@ev1lQuark Please backport the bugfixes to the main branch.

PR #2594, opened 2024-02-04
Fix Prometheus Pushgateway can not enable

FoghostCn:
Fix Prometheus Pushgateway can not enable

### DIFF
--- a/metrics/options.go
+++ b/metrics/options.go
@@ -76,6 +76,13 @@ func WithPrometheusExporterEnabled() Option {
 	}
 }
 
+func WithPrometheusPushgatewayEnabled() Option {
+	return func(opts *Options) {
+		enabled := true
+		opts.Metrics.Prometheus.Pushgateway.Enabled = &enabled
+	}
+}
+
 func WithPrometheusGatewayUrl(url string) Option {
 	return func(opts *Options) {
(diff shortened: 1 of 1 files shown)


metrics/options.go
```diff
@@ -76,6 +76,13 @@ func WithPrometheusExporterEnabled() Option {
 	}
 }
 
+func WithPrometheusPushgatewayEnabled() Option {
+	return func(opts *Options) {
+		enabled := true
+		opts.Metrics.Prometheus.Pushgateway.Enabled = &enabled
```
baerwang, line 82: why use pointer?

FoghostCn: cause https://github.com/apache/dubbo-go/issues/2335

baerwang: There are multiple PRs in this issue that have resolved this issue

FoghostCn: Yes,and this pr just add a method to modify the configuration

justxuewei:
Could you backport this patch to `release-3.1`?

justxuewei (changes requested):
Block temporarily.

FoghostCn:
> Could you backport this patch to `release-3.1`?

`release-3.1` has no options support,so this pr no need to backport to `release-3.1`

justxuewei (approved):
Lgtm！

Issue #2592, opened 2024-02-02
[3.2] The configuration default value flow problem


FoghostCn:
### Issue description

The configuration default value flow problem, instanceOptions -> ServerOptions/ClientOptions -> ServiceOptions is the default value flow direction, and should not flow in the reverse direction. Each level inherits the default value of the previous level and can be modified at the current level. It should not affect the previous level. level value, the current implementation will synchronize back to InstanceOptions when modifying some configurations at the server layer.

配置的默认值流向问题，instanceOptions -> ServerOptions/ClientOptions -> ServiceOptions 为默认值流向，而不应该反向流动，每一级继承上一级的默认值，可以在当前层级修改，不应该影响上一级的值，目前的实现 在 server 层修改一些配置会同步回 InstanceOptions

setcy:
I'm interested in this issue, can it be assigned to me?

Here is my proposal:

## Background
```go
// NewClient is like client.NewClient, but inject configurations from RootConfig and
// ConsumerConfig
func (ins *Instance) NewClient(opts ...client.ClientOption) (*client.Client, error) {
	if ins == nil || ins.insOpts == nil {
		return nil, errors.New("Instance has not been initialized")
	}

	var cliOpts []client.ClientOption
	conCfg := ins.insOpts.Consumer
	appCfg := ins.insOpts.Application
	regsCfg := ins.insOpts.Registries
	sdCfg := ins.insOpts.Shutdown
	metricsCfg := ins.insOpts.Metrics
	otelCfg := ins.insOpts.Otel
        ...
}
```

## Proposal
In the current implementation of the `NewClient`, `NewServer` method, since configurations are passed directly to methods via pointers. Since the pointer refers to the same object, any modifications to the configuration options are reflected back on the original Instance object.

To address the aforementioned issues, I consider to use deep copied configurations to create clients, and implement respective deep copy functions in each type of configuration item. The design code will look like this:
```go
	conCfg := ins.ConsumerOpts()
	appCfg := ins.ApplicationOpts()
	regsCfg := ins.RegistriesOpts()
	sdCfg := ins.ShutdownOpts()
	metricsCfg := ins.MetricsOpts()
	otelCfg := ins.OtelOpts()
```

PR #2596, opened 2024-02-14
feat: deep copy the configuration when instanceOptions flow to ServerOptions/ClientOptions

setcy:
#2592

### DIFF
--- a/global/tls_config.go
+++ b/global/tls_config.go
@@ -32,7 +32,7 @@ type TLSConfig struct {
 
 func DefaultTLSConfig() *TLSConfig {
 	// please refer to /config/tls_config.go NewTLSConfigBuilder
-	return nil
+	return &TLSConfig{}
 }
 
 // Clone a new TLSConfig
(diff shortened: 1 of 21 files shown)


global/tls_config.go
```diff
@@ -32,7 +32,17 @@ type TLSConfig struct {
 
 func DefaultTLSConfig() *TLSConfig {
 	// please refer to /config/tls_config.go NewTLSConfigBuilder
-	return nil
+	return &TLSConfig{}
```
setcy, line 35: To avoid panic, I changed the DefaultTLSConfig function to return a new instance of TLSConfig instead of nil. This should be fine, right?

setcy: Well, based on the CI results, it seems this can't be changed.

chickenlj (approved):
LGTM.

DMwangnima (commented):
Nice Job @setcy !
There are some points that we could discuss:
1. ```Clone``` relys on  handwriting individual fields. It is easy to miss corresponding clone procedure when we add fields. I think we should add test to ensure that the missing would not happen.
2. The issue had illustrated the modifying problem. Maybe we could add a test in dubbo_test.go to cover this scenario?(i.e. Modify some parts of configuration in Client/Server layer, and the configurations in upper layer did not be modified)

Issue #2602, opened 2024-03-03
Declare Authors under the Apache License


justxuewei:
https://github.com/apache/dubbo-go/pull/2405#issuecomment-1974822321 /cc @jizhuozhi

PR #2603, opened 2024-03-04
chore: Include CloudWeGo copyright notices

justxuewei:
There are some components, such as iwrr load balancer, introduced from
CloudWeGo/kitex. This patch is updating to include CloudWeGo copyright
notices in the `Notice` file. You may refer to
[#notice](https://www.apache.org/legal/src-headers.html#notice) for the
specification of ASF policy.

Fixes: #2602


jizhuozhi:
LGTM

Issue #2614, opened 2024-03-09
Failed to start using nacos as configuration center


Parksosaurus:
### Environment
- Server:  Dubbo-go, v3.1.1
- Client:  Dubbo-go, v3.1.1
- Protocol: Triple
- ConfiigCenter: Nacos: v2.2.3
- Registry: Zookeeper, v3.5.10

### Issue description
When I used nacos as the configuration center for configuration, I found that the service could not be successfully registered. This causes the client to start an error.

### Logs

<details><summary>Click me to check logs</summary>

server:
```
2024-03-09 19:04:46     WARN    server/action.go:164    The service greet.GreetService'svcOpts '[]' protocols don't has right protocolConfigs, Please check your configuration center and transfer protocol 
2024-03-09 19:04:46     WARN    server/action.go:164    The service grpc.health.v1.Health'svcOpts '[]' protocols don't has right protocolConfigs, Please check your configuration center and transfer protocol 
2024-03-09 19:04:46     WARN    server/action.go:164    The service grpc.reflection.v1alpha.ServerReflection'svcOpts '[]' protocols don't has right protocolConfigs, Please check your configuration center and transfer protocol 
```

client:
```
2024-03-09 19:05:24     INFO    nacos/impl.go:78        [Nacos ConfigCenter] New Nacos ConfigCenter with Configuration: &{BaseDynamicConfiguration:{} url:0x140001762a0 rootPath: wg:{noCopy:{} state1:0 state2:0} cltLock:{state:0 sema:0} done:0x14000240360 client:<nil> keyListeners:{mu:{state:0 sema:0} read:{v:<nil>} dirty:map[] misses:0} parser:<nil>}, url = nacos://127.0.0.1:8848?config-center.appId=dubbo&config-center.cluster=&config-center.group=myGroup&config-center.namespace=&config-center.timeout=10s&nacos.access=&nacos.group=myGroup&nacos.namespaceId=&nacos.password=&nacos.secret=&nacos.timeout=10s&nacos.username=&remote-client-name=dubbo.config-center-nacos-127.0.0.1%3A8848
2024-03-09T19:05:24.130+0800    INFO    nacos_client/nacos_client.go:65 logDir:</Users/xducodert/Library/Caches/JetBrains/GoLand2023.1/tmp/GoLand/log> cacheDir:</Users/xducodert/Library/Caches/JetBrains/GoLand2023.1/tmp/GoLand/cache>
panic: runtime error: index out of range [0] with length 0

```

[...]

PR #2616, opened 2024-03-10
fix: unable to merge configurations from the config_center

FinalT:
refer:#2614

### DIFF
--- a/options.go
+++ b/options.go
@@ -90,6 +90,7 @@ func (rc *InstanceOptions) init(opts ...InstanceOption) error {
 		log.Infof("[Config Center] Config center doesn't start")
 		log.Debugf("config center doesn't start because %s", err)
 	} else {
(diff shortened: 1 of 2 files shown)


options.go
```diff
@@ -90,6 +90,7 @@ func (rc *InstanceOptions) init(opts ...InstanceOption) error {
 		log.Infof("[Config Center] Config center doesn't start")
 		log.Debugf("config center doesn't start because %s", err)
 	} else {
+		compatInstanceOptions(rcCompat, rc)
```
chickenlj, line 93: Is this the reverse operation of `compatRootConfig(c *InstanceOptions) *config.RootConfig`? This is to copy configurations in RootConfig into InstanceOptions while `compatRootConfig(c *InstanceOptions) *config.RootConfig` does the opposite.

PR #2618, opened 2024-03-11
fix: log

Parksosaurus:
When I was writing the example of dubbo-go-samples, I found that there was a problem with printing log.
image


FinalT (approved):
LGTM.

Issue #2620, opened 2024-03-12
sentinel.TraceError和sentinel.Exit没有被调用


r27153733:
### Environment

- Server: Dubbo-go, v3.1.1
- Client: Dubbo-go, v3.1.1
- Protocol: Triple

### Issue description

context.WithValue 只能往子ctx存value，这导致OnResponse无法通过ctx.Value获取entry，导致entry没退出，sentinel.TraceError也不会被调用。

DMwangnima:
Can you provide more detailed descriptions like the concrete code location or code screenshot?

r27153733:
> 您能否提供更详细的描述，例如具体代码位置或代码截图？

![图片](https://github.com/apache/dubbo-go/assets/69096367/be285e59-9889-4496-973f-e7aa76b8a537)
![图片](https://github.com/apache/dubbo-go/assets/69096367/2f918915-5606-452a-bac0-986283cb55c9)
![图片](https://github.com/apache/dubbo-go/assets/69096367/e9bbf598-7c0b-4e56-bc09-5a26b2c07fb9)
![图片](https://github.com/apache/dubbo-go/assets/69096367/0d45851c-17d1-4725-a6ab-99e5532dd21a)

PR #2621, opened 2024-03-12
Fix/sentinel filter

r27153733:
#2620


filter/sentinel/filter.go
```diff
@@ -110,6 +101,12 @@ type sentinelProviderFilter struct{}
 
 func newSentinelProviderFilter() filter.Filter {
 	if sentinelProvider == nil {
+		initOnce.Do(func() {
+			err := sentinel.InitDefault()
+			if err != nil {
+				panic(err)
+			}
```
justxuewei, line 108:
```suggestion
if err := sentinel.InitDefault(); err != nil {
    panic(err)
}
```

filter/sentinel/filter.go
```diff
@@ -129,8 +126,9 @@ func (d *sentinelProviderFilter) Invoke(ctx context.Context, invoker protocol.In
 	if b != nil {
 		// interface blocked
 		return sentinelDubboProviderFallback(ctx, invoker, invocation, b)
+	} else {
+		defer interfaceEntry.Exit()
 	}
```
justxuewei, line 131: `else` here could be removed.

```suggestion
	}
	defer interfaceEntry.Exit()
```

filter/sentinel/filter.go
```diff
@@ -139,13 +137,19 @@ func (d *sentinelProviderFilter) Invoke(ctx context.Context, invoker protocol.In
 	if b != nil {
 		// method blocked
 		return sentinelDubboProviderFallback(ctx, invoker, invocation, b)
+	} else {
+		defer methodEntry.Exit()
 	}
```
justxuewei, line 142: Ditto.

filter/sentinel/filter.go
```diff
@@ -177,22 +187,27 @@ func (d *sentinelConsumerFilter) Invoke(ctx context.Context, invoker protocol.In
 	if b != nil {
 		// interface blocked
 		return sentinelDubboConsumerFallback(ctx, invoker, invocation, b)
+	} else {
+		defer interfaceEntry.Exit()
 	}
```
justxuewei, line 192: Ditto.

filter/sentinel/filter.go
```diff
@@ -177,22 +187,27 @@ func (d *sentinelConsumerFilter) Invoke(ctx context.Context, invoker protocol.In
...
-	ctx = context.WithValue(ctx, InterfaceEntryKey, interfaceEntry)
 
 	methodEntry, b = sentinel.Entry(methodResourceName, sentinel.WithResourceType(base.ResTypeRPC),
 		sentinel.WithTrafficType(base.Outbound), sentinel.WithArgs(invocation.Arguments()...))
 	if b != nil {
 		// method blocked
 		return sentinelDubboConsumerFallback(ctx, invoker, invocation, b)
+	} else {
+		defer methodEntry.Exit()
 	}
```
justxuewei, line 201: Ditto.

justxuewei (commented):
Thanks! A few comments about code style.

In addition, I have another concern: `sentinel.TraceError` is called by `OnResponse()` previously, but now is called by `Invoke()`. **You changes the behavior of the filter.**

IMHO, you concern is that the context isn't updated with the latest data, right? So a possible fix could be:

```go
 func (fi *FilterInvoker) Invoke(ctx context.Context, invocation protocol.Invocation) protocol.Result { 
 	result := fi.filter.Invoke(ctx, fi.next, invocation)
 	// copy the context from the latest one
 	return fi.filter.OnResponse(ctx, result, fi.invoker, invocation) 
 }

r27153733:
> Thanks! A few comments about code style.
> 
> In addition, I have another concern: `sentinel.TraceError` is called by `OnResponse()` previously, but now is called by `Invoke()`. **THE SIDE EFFECT: You changes the behavior of the filter.**
> 
> IMHO, your concern is that the context isn't updated with the latest data, right? So a possible fix could be:
> 
> ```go
>  func (fi *FilterInvoker) Invoke(ctx context.Context, invocation protocol.Invocation) protocol.Result { 
>  	result := fi.filter.Invoke(ctx, fi.next, invocation)
>  	// do something to copy the context from the latest one
>  	return fi.filter.OnResponse(ctx, result, fi.invoker, invocation) 
>  } 
> ```
> 
> Wdyt?
> 
> UPDATE 1: Maybe the above way is not easy to implement. But the main idea is to share the context between `fi.filter.Invoke()` and `fi.filter.OnResponse()`.

我也想在fi.filter.Invoke() 和 fi.filter.OnResponse()中共享上下文，很遗憾的是修改FilterInvoker是不够的，必须修改Filter接口的定义。
我曾经想从java中找到灵感，但是...

justxuewei (approved):
This looks good to me overall. For what I mentioned eariler, I found [a post](https://medium.com/@bserdar_67914/gos-context-how-can-i-copy-the-values-stored-in-a-context-cf17036ba1ea) introducing a wrapped context, which could be worth considering.

Issue #2641, opened 2024-03-22
client can't get attachments which is set from server


2456868764:
### Environment

- Server: Dubbo-go, master
- Client: Dubbo-go, master
- Protocol: Triple
- Registry: N/A

### Issue description

server code 

```
func (srv *GreetTripleServer) Greet(ctx context.Context, req *greet.GreetRequest) (*greet.GreetResponse, error) {
	resp := &greet.GreetResponse{Greeting: req.Name}
	ctx = context.WithValue(ctx, constant.AttachmentKey, map[string]interface{}{"server_info": "dubbo httpbin"})
	return resp, nil
}
```

client code 

```
        ctx := context.Background()
	resp, err := svc.Greet(ctx, &greet.GreetRequest{Name: "hello world"})
	if err != nil {
		c.JSON(http.StatusBadGateway, err.Error())
		return
	}
	attachments := make(map[string]interface{})
	if ctx.Value(constant.AttachmentKey) != nil {
		attachments = ctx.Value(constant.AttachmentKey).(map[string]interface{})
	}
```

and found that client attachments is nil.
then use curl to simulate as http which return is follow:

````
(base) ➜  dubbo-go git:(feat-security) ✗ curl -v -H "content-type:application/json" http://127.0.0.1:8000/greet.GreetService/Greet -d '{"name":"hello"}'
*   Trying 127.0.0.1:8000...
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> POST /greet.GreetService/Greet HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/8.1.2
> Accept: */*
> content-type:application/json
> Content-Length: 16
>
< HTTP/1.1 200 OK
< Accept-Encoding: gzip
< Content-Type: application/json
< Date: Fri, 22 Mar 2024 03:05:43 GMT
< Content-Length: 20
<
* Connection #0 to host 127.0.0.1 left intact
{"greeting":"hello"}%

no attachments return

````

### Logs

<details><summary>Click me to check logs</summary>

```
Copy logs to here.
```

</details>

YarBor:
plz assign to me

Yan0613:
thx, I will fix it as soon as possible.

Yan0613:
> Hi @Yan0613 I was working on this issue and don't think two people need to work on this. : ) 
> I have been discussing this with mentors for some time. 
> However, if you have a solution ready, please let me know. Thanks!
> 

[...]

PR #2648, opened 2024-03-31
fix: Server return with Attachment

YarBor:
fix: Server return with Attachment (#2641)

> > in this pr, every call will return all attachment-data , whether built-in data or user-attached data. in triple unray call 
> > 
> > but in this issue 
> > i have no idea to fix it in an elegant way
> > 
> > I tried adding new fields like (userAttachmentKey or whatever) but it's not very nice to separate the built-in data from the user-attached data. I also thought about modifying the key of user-attached data before transfer and restoring it after transfer, but that doesn't seem feasible in this question (called in curl mode).
> > 
> > there may need some good idea？
> > 
> > but in this issue 
> > i have no idea to fix it in an elegant way
> > 
> > I tried adding new fields like (userAttachmentKey or whatever) but it's not very nice to separate the built-in data from the user-attached data. I also thought about modifying the key of user-attached data before transfer and restoring it after transfer, but that doesn't seem feasible in this question (called in curl mode).
> > 
> > there may need some good idea？
> > 
> > 
> 
> add new api
> 
> One possible point to discuss is that in the additional data received by both parties, the first letter of the key will be changed to a capital letter.

client-side-call
```go
func rpcCaller(){
        header := http.Header{"testKey1": []string{"testVal1"}, "testKey2": []string{"testVal2"}}
        // to store outgoing data ,and reserve the location for the receive field.
        // header will be copy , and header's key will change to be lowwer. 
	ctx, _ := triple_protocol.NewOutgoingContext(context.Background(), header )

        res, err := cli.Greet(ctx, &greet.GreetRequest{Name: "triple"})
	// stream, err := cli.GreetServerStream(ctx, &greet.GreetServerStreamRequest{Name: "triple"})
        // stream, err := cli.GreetStream(ctx)
        // stream, err := cli.GreetClientStream(ctx)

        /* 
             do something and close connection.
        */ 

        // to get return data , and key is lowwer , like grpc.
      	data, err := triple_protocol.FromIncomingContext(ctx)
}
```
server-side 
```go
func (impl *Impl) ****call(ctx context.Context, *** api.Req) (*** , errno){
	data, _ := triple.FromIncomingContext(ctx)

        // do something

        triple.SetOutgoingData(ctx , http.Header{"OutgoingDataKey1" , []string{"OutgoingDataVal1"}})
	/* OR */ 
        triple.AppendToOutgoingContext(ctx, "OutgoingContextKey2", "OutgoingDataVal2")
}
```

new api applies to all 4 calling methods.


protocol/triple/triple_protocol/header.go
```diff
@@ -181,3 +231,11 @@ func SendHeader(ctx context.Context, header http.Header) error {
 	mergeHeaders(conn.RequestHeader(), header)
 	return conn.Send(nil)
 }
+
+func outGoingKeyCheck(key string) bool {
```
DMwangnima, line 235: Where is this function used?

client/client.go
```diff
@@ -59,7 +59,8 @@ func (conn *Connection) call(ctx context.Context, reqs []interface{}, resp inter
 	if err != nil {
 		return nil, err
 	}
-	return conn.refOpts.invoker.Invoke(ctx, inv), nil
+	res := conn.refOpts.invoker.Invoke(ctx, inv)
```
DMwangnima, line 62: It seems there is no need to change statement style.

protocol/triple/triple_invoker.go
```diff
@@ -136,36 +143,50 @@ func (ti *TripleInvoker) Invoke(ctx context.Context, invocation protocol.Invocat
 	return &result
 }
 
+func mergeAttachmentToOutgoing(ctx context.Context, attachments map[string]interface{}) (context.Context, error) {
```
DMwangnima, line 146: Since you extract attachments merge logic from ```parseInvocation```. Why do not extract entire logics about processing attachments?

protocol/triple/triple_protocol/header.go
```diff
@@ -88,61 +90,109 @@ func addHeaderCanonical(h http.Header, key, value string) {
...
+	return context.WithValue(ctx, extraDataKey{}, extraData)
 }
 
 // NewOutgoingContext sets headers entirely. If there are existing headers, they would be replaced.
 // It is used for passing headers to server-side.
 // It is like grpc.NewOutgoingContext.
 // Please refer to https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#sending-metadata.
-func NewOutgoingContext(ctx context.Context, header http.Header) context.Context {
-	return context.WithValue(ctx, headerOutgoingKey{}, header)
+func NewOutgoingContext(ctx context.Context, data interface{}) (context.Context, error) {
```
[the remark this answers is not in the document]
YarBor: I will change it back.

protocol/triple/triple_protocol/header.go
```diff
@@ -88,61 +90,109 @@ func addHeaderCanonical(h http.Header, key, value string) {
...
+	extraData[headerOutgoingKey] = header
+	return context.WithValue(ctx, extraDataKey{}, extraData), nil
 }
 
 // AppendToOutgoingContext merges kv pairs from user and existing headers.
 // It is used for passing headers to server-side.
 // It is like grpc.AppendToOutgoingContext.
 // Please refer to https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#sending-metadata.
-func AppendToOutgoingContext(ctx context.Context, kv ...string) context.Context {
+func AppendToOutgoingContext(ctx context.Context, kv ...string) (context.Context, error) {
```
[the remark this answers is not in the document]
YarBor: I think this needs to be discussed. Should there be constraints on the attachment key? When transmitting and returning within the framework, the first letter of the attachment key will be changed to uppercase. On the receiving side, we cannot restore the original key.

protocol/triple/triple_protocol/header.go
```diff
@@ -88,61 +90,109 @@ func addHeaderCanonical(h http.Header, key, value string) {
...
 		header.Add(kv[i], kv[i+1])
 	}
-	return context.WithValue(ctx, headerOutgoingKey{}, header)
+	return ctx, nil
 }
 
-func ExtractFromOutgoingContext(ctx context.Context) http.Header {
-	headerRaw := ctx.Value(headerOutgoingKey{})
-	if headerRaw == nil {
+func ExtractFromOutgoingContext(ctx context.Context) map[string][]string {
```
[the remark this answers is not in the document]
YarBor: i will change it back.

protocol/triple/triple_protocol/header.go
```diff
@@ -88,61 +90,109 @@ func addHeaderCanonical(h http.Header, key, value string) {
...
-	// since headerOutgoingKey is only used in triple_protocol package, we need not verify the type
-	return headerRaw.(http.Header)
 }
 
 // FromIncomingContext retrieves headers passed by client-side. It is like grpc.FromIncomingContext.
+// it must call after append/setOutgoingContext to return current value
 // Please refer to https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#receiving-metadata-1.
-func FromIncomingContext(ctx context.Context) (http.Header, bool) {
-	header, ok := ctx.Value(headerIncomingKey{}).(http.Header)
+func FromIncomingContext(ctx context.Context) (map[string][]string, bool) {
```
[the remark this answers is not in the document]
YarBor: i will change it back.

protocol/triple/triple_protocol/header.go
YarBor: ditto.

YarBor: i think it should be check.

protocol/triple/triple_invoker.go
YarBor: i think u are right.

client/client.go
YarBor: Yes, the change here is a mistake. This was a change while testing, I will change it back to how it was.

chickenlj (approved):
LGTM.

PR #2669, opened 2024-05-10
feat : add script routing functionality

YarBor:
### JS Standard
  - The expect to recv js-script data is like:
```js
(function route(invokers, invocation, context) {
    var result = [];
    for (var i = 0; i < invokers.length; i++) {
        if ("127.0.0.1" === invokers[i].GetURL().Ip) {
            if (invokers[i].GetURL().Port !== "20000") {
                invokers[i].GetURL().Ip = "anotherIP"
                result.push(invokers[i]);
            }
        }
    }
    return result;
}(invokers, invocation, context));
```
  - The expected way to return value is like:

```js
    var result = [];
    result.push(invokers[i]);
    return result;
```
---
  - Supports method calling.
    Parameter methods are mapped by the passed in type.
    The called method first letter is capitalized (does not comply with js specifications).

e.g. `invokers[i].GetURL().SetParam("testKey","testValue")`

---
  - Like the Go language, it supports direct access to exportable variables within parameters.

e.g. `invokers[i].GetURL().Port`

--- 
  - publish config like :
```go
scriptRouteConfig := data, err := os.ReadFile(pathToConfigfile)
conf, _ := config.GetRootConfig().ConfigCenter.GetDynamicConfiguration()
_ = conf.PublishConfig("dubbo.io"+constant.ScriptRouterRuleSuffix, "dubbo", scriptRouteConfig)

```
  - config file like :
```yaml
configVersion: v3.0
key: dubbo.io
type: javascript
enabled: true
script: |
  (function route(invokers,invocation,context) {
  	var result = [];
  	for (var i = 0; i < invokers.length; i++) {
  	    if ("10.30.0.218" === invokers[i].GetURL().Ip) {
  			if (invokers[i].GetURL().Port === "20000"){
  				invokers[i].GetURL().Port = "20001" 
  				result.push(invokers[i]);
  			}
  	    }
  	}
  	return result;
  }(invokers,invocation,context));
```
---
#### Important
In the js-support.

1. Input args are `invokers, invocation, context`,
  The following methods are not allowed.
    - `invokers.Invoke(...)` 
    - `invokers.Destroy()`
    - `invokers.IsAvailable()` will always return true
2. Wrong input is program acceptable, it will not return an error or go panic , it will make an undefined error.
for example:
  ```js
  (function route(invokers, invocation, context) {
      var result = [];
      for (var i = 0; i < invokers.length; i++) {
          if ("127.0.0.1" === invokers[i].GetURL().Ip) {
              if (invokers[i].GetURL(" Err Here ").Port !== "20000") {
  //---------------------------------^  here wont go err , it will trans to ()
                  invokers[i].GetURL().Ip = "anotherIP"
                  result.push(invokers[i]);
              }
          }
      }
      return result;
  }(invokers, invocation, context));
  ```
  ```js
  (function route(invokers, invocation, context) {
      var result = [];
      for (var i = 0; i < invokers.length; i++) {
          if ("127.0.0.1" === invokers[i].GetURL().Ip) {
              if (invokers[i].GetURL(" Err Here ").Port !== "20000") {
                  invokers[i].GetURL().AddParam( null , "key string", "value string")
  //---------------------------------------------^  here wont make err , it will trans to ("" , "key string" , "value string")
                  result.push(invokers[i]);
              }
          }
      }
      return result;
  }(invokers, invocation, context));
  ```
  ```js
  (function route(invokers, invocation, context) {
      var result = [];
      for (var i = 0; i < invokers.length; i++) {
          if ("127.0.0.1" === invokers[i].GetURL().Ip) {
              if (invokers[i].GetURL(" Err Here ").Port !== "20000") {
                  invokers[i].GetURL().AddParam( invocation , "key string", "value string")
  //---------------------------------------------^  here wont make err , it will trans to ("[object Object]" , "key string" , "value string")
                  result.push(invokers[i]);
              }
          }
      }
      return result;
  }(invokers, invocation, context));
  ```
3. Wrong methods-call will go error.
for example:
  ```js
  (function route(invokers, invocation, context) {
      var result = [];
      for (var i = 0; i < invokers.length; i++) {
          if ("127.0.0.1" === invokers[i].GetURL().Ip) {
              if (invokers[i].GetURLS(" Err Here ").Port !== "20000") {
  //--------------------------^ // here will make err 
                  invokers[i].GetURLS().Ip = "anotherIP"
                  result.push(invokers[i]);
              }
          }
      }
      return result;
  }(invokers, invocation, context));
  ```


cluster/router/script/instance/instances_pool.go
```diff
@@ -0,0 +1,109 @@
...
+		return invokers, nil
+	}
+	matcher := i.insPool.Get().(*jsInstance)
+	matcher.initCallArgs(invokers, invocation)
+	matcher.initReplyVar()
+	scriptRes, err := matcher.runScript(i.program.Load())
+	if err != nil {
+		return nil, err
+	}
+	result := make([]protocol.Invoker, 0)
```
baerwang, line 59: map add cap

cluster/router/script/instance/instances_pool.go
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package instance
+
+import (
+	"dubbo.apache.org/dubbo-go/v3/protocol"
```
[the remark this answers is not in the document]
YarBor: I don't understand what the problem is, I ran go fmt and it doesn't work, looks the same.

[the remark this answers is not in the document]
EndlessSeeker: > I don't understand what the problem is, I ran go fmt and it doesn't work, looks the same.

Go fmt will not change the order of the import, it will only format locks and spaces. You can use goimports instead,https://pkg.go.dev/golang.org/x/tools/cmd/goimports
Generally speaking, the standard import format is:
import (
go standard packages (sorted alphabetically)
(space)
Three-party packages (sorted alphabetically)
)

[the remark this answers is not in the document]
YarBor: :+1:

chickenlj:
1. 监听的规则 key 应该是 provider 应用名，而不是当前进程自身应用名。可参考 tagrouter 的 notify 方法
2. 第1个逻辑改了后，Script Instance 实例的对应关系也要改一下，当前是全局共享的
3. 要确认下 dynamicconfuguration.AddListener() 的逻辑对不对（zookeeper/nacos），因为有多个 ScriptRouter 实例可能对应到一个 key (provider应用名)，dynamicconfuguration.AddListener()应该支持调用多次且记录多个listener --- zookeeper实现看起来没问题，nacos实现好像有问题。
4. instance.Compile 并发调用有没有问题

cluster/router/script/instance/js_instance.go
```diff
@@ -0,0 +1,203 @@
...
+		i.pgLock.Lock()
+		// double check to avoid race
+		if pg, ok = i.program[rawScript]; ok {
+			pg.addCount(1)
+		} else {
+			i.program[rawScript] = newProgram(newPg)
+		}
+		i.pgLock.Unlock()
+		return nil
+	}
```
chickenlj, line 143: The lock used here is a bit strange, try the following steps:
1. check without lock
2. if exists, return
3. if not exists, lock
   3.1 check again
   3.2 if exists return
   3.3 if not exists compile

YarBor: i will rewite it.

chickenlj: I see. I thought the following lock was not necessary, it seems to be necessary in this case.

```
i.pgLock.RLock()
pg, ok = i.program[rawScript]
i.pgLock.RUnlock()
```

chickenlj (approved):
LGTM.

Issue #2684, opened 2024-06-04
to support conditional routing with multiple destinations, customize conditional routing priorities and operation in route fail


YarBor:
**What would you like to be added**:
Support conditional routing with multiple destinations, and customize conditional routing priorities.

**Why is this needed**:
To cope with more complex routing scenarios.

Now [condition-rule](https://cn.dubbo.apache.org/en/overview/core-features/traffic/condition-rule/#condition) is
![image](https://github.com/apache/dubbo-go/assets/110281261/48590f9a-0e44-4738-b821-35ebf7692898)
Specifically, the matching condition of dubbo traffic(now) is only one match, and the incoming consumerURL is matched against all matching conditions, and the URLs of all callable objects and filtering conditions are matched.

in complex routing: (for example, i think what need to support in condition-routing)

all active service providers go match `region=$region & env=$env` first,
Match success, call them in loadbalance.
Match fail , then go match `env=$env` or any other ... 
until get service providers i want or nothing. 

Also, it is necessary to customize the behavior when a single conditional route fails to match. (match next or return null directly)

PR #2685, opened 2024-06-04
support conditional routing with multiple destinations, customize conditional routing priorities and operation in route fail

YarBor:
to support #2684 

new condition-rule config like this
```yaml
configVersion: v3.1
scope: service
force: true
runtime: true
enabled: true
key: org.apache.dubbo.samples.CommentService
conditions:
  - rule: method=getComment & env=gray & region=beijing => region=beijing & env=gray
    force: fasle
    priority: 10
  - rule: method=getComment => region=Hangzhou
    force: false
    priority: 10
  - rule: method=echo =>
    force: true
    priority: 15
```
| Condition Field | Type   | Required | DefaultValue |
|-----------------|--------|---------|---------|
| Rule            | String |   &#10003;    |         |
| Force           | Bool   |         | False   |
| Priority        | Int64  |         | 0       |

through `configVersion` to choose parse and load logic. 
set `configVersion: v3.1` to use new feature.

### DIFF
--- a/cluster/router/condition/dynamic_router.go
+++ b/cluster/router/condition/dynamic_router.go
@@ -18,9 +18,7 @@
 package condition
 
 import (
-	"dubbo.apache.org/dubbo-go/v3/cluster/utils"
 	"fmt"
-	"gopkg.in/yaml.v2"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


chickenlj (approved):
LGTM.

YarBor:
#2686

PR #2687, opened 2024-06-13
Compatibility/java interop

SHA-4096:
# What's the purpose of the change
## 1. handler mismatch
When interoperating with dubbo-java, there may be an issue when client from java side try to invoke an rpc function whose first-letter is in lower case, since dubbo-go now only provide handler whose method name is always with upper-cased first letter, due to the language feature of golang. It would cause failure in rpc function invocation, as mentioned in https://github.com/apache/dubbo-go-samples/pull/771 .

To tackle this problem, a straightforward idea is to register another handler with a different first-letter case, and that's the change made by this pull request.

## 2. content-type codec mismatch
When using hessian2 for serialization, some version of java-side client would set a wrong content-type in request, causing marshal and unmarshal error. In order to tackle the problem, a backupCodec is introduced to handle the message in the server-side serialization type. 

# Validity of the solution
## solution for problem 1
After adding a breakpoint at `dubbo-go/server/server.go` at line 254, which is the section that exports service , we can see the registered handlers on `service->serviceOpts->rpsService->s->triServer->handlers`,  and  `/org.apache.dubbo.tri.hessian2.api.GreetingsService/greet` is made possible by changes in this PR. 
image

## solution for problem 2
A warn message would be logged when request content-type and server serialization mismatches, and the server would try to marshal/unmarshal message in server serialization codec. 
image


chickenlj:
Please run `make fmt` to format the code

common/dubboutil/case_invert.go
```diff
@@ -0,0 +1,18 @@
+package dubboutil
```
chickenlj, line 1: Add apache license header for this file

common/rpc_service.go
```diff
@@ -19,6 +19,7 @@ package common
 
 import (
 	"context"
+	"dubbo.apache.org/dubbo-go/v3/common/dubboutil"
```
FoghostCn, line 22: pls format imports

server/server.go
```diff
@@ -180,6 +181,18 @@ func (s *Server) exportServices() (err error) {
 			err = svcOpts.ExportWithoutInfo()
 		} else {
 			info := infoRaw.(*ServiceInfo)
+			//Add a method with a name of a differtent first-letter case
+			//to achieve interoperability with java
+			var additionalMethods []MethodInfo
+			for _, method := range info.Methods {
+				newMethod := method
+				newMethod.Name = dubboutil.SwapCaseFirstRune(method.Name)
```
chickenlj, line 189: Do we need a deep copy of `method`?

SHA-4096: I'll have a look at other solutions

common/dubboutil/case_invert.go
SHA-4096: Added

common/rpc_service.go
SHA-4096: formatted

chickenlj (approved):
LGTM.

PR #2688, opened 2024-06-15
patch: update sort-func in generating condition-ruleV3.1

YarBor:
Increase the priority of traffic-disable option in `condition` sorting.


YarBor:
#2684

chickenlj (approved):
LGTM.

PR #2691, opened 2024-06-28
Fix nacos not support empty serverConfigs

AlbumenJ:
When nacos use address server mode, should support empty `serverConfigs`.


chickenlj (approved):
LGTM.

PR #2693, opened 2024-07-01
Update UseAsMetaReport and UseAsConfigCenter as String to prevent value override

AlbumenJ:
- Fix unable to set UseAsMetaReport and UseAsConfigCenter as `false` by fluent API or YAML Config


chickenlj (approved):
LGTM.

PR #2698, opened 2024-07-08
bug fixes and updates in ut

SHA-4096:
# What's the purpose of this PR
Code merged in #2687 canot pass unit test related to methods and codec, due to additional methods added for compatibility. The following is done:

- [x] update ut related to methods
- [x] fixed a possible segmentation fault caused by a nil backupCodec


FoghostCn (changes requested):
pls fix ci

chickenlj (approved):
LGTM. The integration testing failure is caused by other reasons.

PR #2715, opened 2024-07-24
fix(registry/nacos): Fixed the bug that NACOS service discovery could…

Allen442:
… not obtain all service names, and added the ability to filter interface-level service IDs


Allen442:
@chickenlj PTAL.

chickenlj (approved):
LGTM.

PR #2723, opened 2024-08-01
Add protobuf based MetadataService support

chickenlj:


chickenlj:
Failure in CI points to a method mapping issue caused by previous commits.

Issue #2701, opened 2024-07-11
 release the allocated memory of the ‘url’ field when destroy invoker


No-SilverBullet:
**What would you like to be added**:

 https://github.com/apache/dubbo-go/blob/faf6f2a27922f0259fa2b2779d238e8865779fe7/protocol/invoker.go#L92-L96

here is my question, since the invoker is destroyed ,so why do we  not set the url field in invoker to nil ,like below

func (bi *BaseInvoker) Destroy() {
	  logger.Infof("Destroy invoker: %s", bi.GetURL()) 
	  **bi.url = nil //add this**
	  bi.destroyed.Store(true)
	  bi.available.Store(false)
}

**Why is this needed**:
in this line, the invoker will be add to a global var dubboProtocol
https://github.com/apache/dubbo-go/blob/faf6f2a27922f0259fa2b2779d238e8865779fe7/protocol/dubbo/dubbo_protocol.go#L98

https://github.com/apache/dubbo-go/blob/faf6f2a27922f0259fa2b2779d238e8865779fe7/protocol/dubbo/dubbo_protocol.go#L59

so when destroy the invoker, the 'url' is useless and still it cannot be gc（unless destroy the global var）,which i think may cause memory leaks

mark4z:
#2223

Issue #2339, opened 2023-06-12
common url OOM


mengccmumu:
### Environment

- Server:
- Client: 
- Protocol: 
- Registry: 

### Issue description

OOM

### Logs

<details><summary>Click me to check logs</summary>

```
Copy logs to here.
![image](https://github.com/apache/dubbo-go/assets/10061880/9838c496-6501-4ca6-95db-188c1ae41d95)
![image](https://github.com/apache/dubbo-go/assets/10061880/9ef197cd-0596-4f23-88de-ed7d77e87653)

```

</details>

justxuewei:
Hi @mengccmumu, there was a pull request to fix this problem, see https://github.com/apache/dubbo-go/pull/2223. Could you help us verify this PR on your environment?

mengccmumu:
> 
already verify this PR,  but OOM still occurs

justxuewei:
From the picture you provided, I don't see any leaks about the `MergeURL`.

mengccmumu:
> From the picture you provided, I don't see any leaks about the `MergeURL`.

you are right .  
 I will analyze the issue with OOM again .

licon:
我们在生产环境也遇到这个问题，希望能尽快解决
pprof list：
```
(pprof) list MergeURL
Total: 489.08MB
ROUTINE ======================== dubbo.apache.org/dubbo-go/v3/common.MergeURL in dubbo.apache.org/dubbo-go/v3@v3.0.2/common/url.go
  271.67MB   324.67MB (flat, cum) 66.38% of Total
         .          .    681:// the params of result will be (a1->v1, b1->v2, a2->v3).
         .          .    682:// You should notice that the value of b1 is v2, not v4.
         .          .    683:// due to URL is not thread-safe, so this method is not thread-safe
         .          .    684:func MergeURL(serviceURL *URL, referenceURL *URL) *URL {
         .          .    685:	// After Clone, it is a new URL that there is no thread safe issue.
         .       52MB    686:	mergedURL := serviceURL.Clone()
         .          .    687:	params := mergedURL.GetParams()
         .          .    688:	// iterator the referenceURL if serviceURL not have the key ,merge in
         .          .    689:	// referenceURL usually will not changed. so change RangeParams to GetParams to avoid the string value copy.// Group get group
         .          .    690:	for key, value := range referenceURL.GetParams() {
         .          .    691:		if v := mergedURL.GetParam(key, ""); len(v) == 0 {
[...]

PR #2726, opened 2024-08-09
fix: set baseinvoker's url to nil to address memory leak issues

No-SilverBullet:
Bssed on PR #2223 ,the main purpose of this PR is to fix the memory leak problem caused by the url pointer in baseinvoker not being released correctly. You can refer to the following issue：

FIXES: #2701 #2339 #2031

### DIFF
--- a/protocol/invoker.go
+++ b/protocol/invoker.go
@@ -88,7 +88,7 @@ func (bi *BaseInvoker) Invoke(context context.Context, invocation Invocation) Re
 	return &RPCResult{}
 }
 
-// Destroy changes available and destroyed flag
+// Destroy changes available and destroyed flag and release the url's allocated memory
 func (bi *BaseInvoker) Destroy() {
 	logger.Infof("Destroy invoker: %s", bi.GetURL())
 	bi.destroyed.Store(true)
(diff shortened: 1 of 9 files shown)


protocol/invoker.go
```diff
@@ -88,11 +88,12 @@ func (bi *BaseInvoker) Invoke(context context.Context, invocation Invocation) Re
 	return &RPCResult{}
 }
 
-// Destroy changes available and destroyed flag
+// Destroy changes available and destroyed flag and release the url's allocated memory
 func (bi *BaseInvoker) Destroy() {
 	logger.Infof("Destroy invoker: %s", bi.GetURL())
 	bi.destroyed.Store(true)
 	bi.available.Store(false)
+	bi.url = nil
```
binbin0325, line 96: If set nil, is there any risk of a null pointer elsewhere

No-SilverBullet: As u can see in other changed files, in other places where`url` are used, I changed the order, that is, destroyed them in the end to avoid null pointer issues; But still requires careful review;

georgehao (approved):

No-SilverBullet:
@chickenlj plz take the time to review this PR, thanks

FinalT (approved):

PR #2732, opened 2024-09-18
fix: CallClientStream result has error

linfeip:
triple rpc server端对CallClientStream的handle处理, 当service return error的时候, 会result的Result()值会是nil, 这样会导致res.Result().(*tri.Response)出现panic


FinalT (approved):
LGTM.

Issue #2743, opened 2024-11-09
dubbo logger outputs inappropriate file names, function names, and line numbers


marsevilspirit:
For example：
![image](https://github.com/user-attachments/assets/cc0d27bd-084e-407b-8d1c-c480afd61a92)

PR #2744, opened 2024-11-09
Fix the log that outputs inappropriate file names, function names, and line

marsevilspirit:
#2743 
Because the logger was encapsulated once in gost, and then encapsulated again in dubbo, it needs to be skiped twice.

such as
image


FoghostCn:
seems right, @zhaoyunxing92

FoghostCn (approved):

Issue #2762, opened 2024-12-15
Code bug: should return a error but return a nil value error


alingse:
### Environment

- Server: 
- Client: 
- Protocol: 
- Registry: 

### Issue description

code 1 
https://github.com/apache/dubbo-go/blob/main/config/tls_config.go#L88
```go
	caBytes, err := os.ReadFile(opt.CACertFile)
	if err != nil {
		return nil, err
	}
	if ok := ca.AppendCertsFromPEM(caBytes); !ok {
		return nil, err
	}
```
the err is nil in line 88, here should define an error to return

code 2

https://github.com/apache/dubbo-go/blob/main/protocol/triple/internal/client/common/client.go#L68

https://github.com/apache/dubbo-go/blob/main/protocol/triple/internal/client/common/client.go#L92

```go
	if sendErr := stream.Send(&greet.GreetStreamRequest{Name: "triple"}); sendErr != nil {
		return err
	}
```
should return `sendErr` but return a nil value `err`

I create a linter to detect code that returns a non-relevant nilness error bug. I checked the top 1000 GitHub Go repositories and found this, all result listed in https://github.com/alingse/sundrylint/issues/4

### Logs

<details><summary>Click me to check logs</summary>

```
Copy logs to here.
```

</details>

No-SilverBullet:
> e that returns a non-relevant nilness error bug. I checked the top 1000 GitHub Go repositories and found this, all result liste

i will submit a PR to fix it

PR #2763, opened 2024-12-17
fix/return err

No-SilverBullet:
based on issue https://github.com/apache/dubbo-go/issues/2762, fix some err return bugs.


FoghostCn (changes requested):
https://github.com/apache/dubbo-go/blob/37bc5c31f1a417819c47f435443c93a4a146cb84/config/tls_config.go#L54-L59

same problem;

No-SilverBullet:
> https://github.com/apache/dubbo-go/blob/37bc5c31f1a417819c47f435443c93a4a146cb84/config/tls_config.go#L54-L59
> 
> same problem;

got it, already fixed

FoghostCn (approved):

Issue #2302, opened 2023-04-14
restful protocol transport dial cause io timeout


578157900:
### Environment

- Client: Dubbo-go, v1.5.9
- Protocol: rest
- Registry: zookeeper

### Issue description

https://github.com/apache/dubbo-go/blob/main/protocol/rest/client/client_impl/resty_client.go#L59
`
DialContext: func(_ context.Context, network, addr string) (net.Conn, error) {
				c, err := net.DialTimeout(network, addr, restOption.ConnectTimeout)
				if err != nil {
					return nil, err
				}
				err = c.SetDeadline(time.Now().Add(restOption.RequestTimeout))
				if err != nil {
					return nil, err
				}
				return c, nil
},
`
http连接池复用tcp连接时，因为设置了deadline，resp未返回直接抛出io timeout。

578157900:
`
func NewRestyClient(restOption *client.RestOptions) client.RestClient {
	client := resty.New()
	client.SetTransport(
		&http.Transport{
			DialContext: (&net.Dialer{
				Timeout: restOption.ConnectTimeout,
			}).DialContext,
			IdleConnTimeout: 90 * time.Second,
		}).SetTimeout(restOption.RequestTimeout)
	return &RestyClient{
		client: client,
	}
}
`
1. 通过配置IdleConnTimeout超时来回收空闲链接。
2. 通过client.SetTimeout来设置每次请求的超时时间。

otis95:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #2785, opened 2025-03-02
Fix/remove setdeadline

No-SilverBullet:
What this PR does

https://pkg.go.dev/net#Conn.SetDeadline

> *// SetDeadline sets the read and write deadlines associated*
>
> *// with the connection. It is equivalent to calling both*
>
> // SetReadDeadline and SetWriteDeadline.

Therefore, `SetDeadline` will cause the connection to be closed before the server returns completely, so `SetDeadline` is removed and `SetTimeout` is used instead. In addition, `IdleConnTimeout` is added to close idle connections after timeout to avoid invalid reuse.

Which issue(s) this PR fixes

https://github.com/apache/dubbo-go/issues/2302

### DIFF
--- a/protocol/rest/client/client_impl/resty_client.go
+++ b/protocol/rest/client/client_impl/resty_client.go
@@ -22,13 +22,18 @@ import (
 	"net"
 	"net/http"
 	"path"
+)
 
+import (
 	"github.com/go-resty/resty/v2"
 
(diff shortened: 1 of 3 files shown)


marsevilspirit (approved):
LGTM!

FoghostCn (approved):

FoghostCn:
LGTM

Issue #2779, opened 2025-02-21
jsonRpc Protocol in dubbogo:  WARN	jsonrpc/server.go:122	[ReadRequest] error: EOF


StupidManYI:
### Environment

- Server: Dubbo-go, v3.2.0-rc1
- Client: Dubbo-go, v3.2.0-rc1
- Protocol: jsonRpc
- Registry: none
- Reproducible Repository: https://github.com/StupidManYI/dubbo-go-jsonRpc-demo  (Q1 -> demo1 ,  Q2 -> demo2)
#### There isn't much documentation about jsonRpc on the official website, so I wrote the code according to that of simples. I encountered the following problems.

Q1. Everytime JsonRpc Client call server, server printed `jsonrpc/server.go:122	[ReadRequest] error: EOF`  (demo1)
Q2. After I used the filter, it took effect on the server side, but didn't work on the client side.

### Logs

<details><summary>Click me to check logs</summary>

### server:
```

2025-02-21 11:05:03	INFO	v3@v3.2.0-rc1/options.go:90	[Config Center] Config center doesn't start
2025-02-21 11:05:03	INFO	jsonrpc/jsonrpc_protocol.go:72	[JSONRPC protocol] Export service: jsonrpc://:20000/GreetProvider?accesslog=&app.version=&application=dubbo.io&auth=&bean.name=&cluster=failover&config.tracing=&environment=&execute.limit=&execute.limit.rejected.handler=&export=true&interface=GreetProvider&loadbalance=random&max-server-recv-msg-size=4mib&max-server-send-msg-size=&metadata-type=local&module=sample&name=dubbo.io&organization=dubbo-go&owner=dubbo-go&param.sign=&pid=13472&registry.role=3&release=dubbo-golang-3.2.0&retries=&serialization=&service.filter=echo%2Ctoken%2Caccesslog%2Ctps%2Cgeneric_service%2Cexecute%2Cpshutdown&side=provider&timestamp=1740107103&tps.limit.interval=&tps.limit.rate=&tps.limit.rejected.handler=&tps.limit.strategy=&tps.limiter=&warmup=
2025-02-21 11:05:03	INFO	jsonrpc/server.go:235	rpc server start to listen on [::]:20000
2025-02-21 11:05:03	WARN	common/rpc_service.go:373	method Resume of mtype func(*health.HealthTripleServer) has wrong number of in out parameters 0; needs exactly 1/2
[...]

PR #2790, opened 2025-03-08
Fix (Server): Avoid error reporting warning logs when detecting EOF via Peek to properly close the connection

1kasa:
- Fix: https://github.com/apache/dubbo-go/issues/2779
- describe: The client explicitly requested to close the connection after sending data, but the server did not handle it correctly, resulting in incorrect log printing.


No-SilverBullet (approved):
LGTM

PR #2791, opened 2025-03-08
fix: Fix problematic error returns

cuishuang:
Here  `sendErr` should be returned instead of `err`.


No-SilverBullet (approved):
LGTM, thanks for your contribution :)

Issue #2683, opened 2024-06-01
Does anyone know about this bug? memory leak


xiaominger535:
report link：https://github.com/apache/dubbo-go-hessian2/issues/373

PR #2792, opened 2025-03-09
fix issue 2683: defeat memory leak

LQyt2012:
There is a memory leak when using the dubbo-go-hessian2 protocol(https://github.com/apache/dubbo-go/issues/2683). Since the dubbo-go-hessian2 already fixed in (https://github.com/apache/dubbo-go-hessian2/pull/374) and released a new version with tag v1.12.5. So update the version in dubbo-go.


No-SilverBullet (approved):
LGTM

PR #2829, opened 2025-04-06
fix common.URL memory leak

LQyt2012:
Add the urlPool to manage the memory allocation of common.URL

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -59,10 +59,14 @@ var (
 	DubboNodes          = [...]string{"consumers", "configurators", "routers", "providers"} // Dubbo service node
 	DubboRole           = [...]string{"consumer", "", "routers", "provider"}                // Dubbo service role
 	compareURLEqualFunc CompareURLEqualFunc                                                 // function to compare two URL is equal
+	urlPool             sync.Pool
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


marsevilspirit:
check ci error.

marsevilspirit:
We need all CI to pass to ensure code quality.

No-SilverBullet:
There are two questions. The first is why memory leaks occur here. The other question is that pooling can improve memory utilization but it does not seem to fundamentally guarantee that there will be no memory leaks.

Linuxdazhao:
why  ReleaseURL func not call, is use sync pool auto remove ?

LQyt2012:
> 这 reset 函数怎么没有一个地方调用？

The user should call the Reset method when retrieving a URL structure from the pool.

LQyt2012:
> > > 这 reset 函数怎么没有一个地方调用？
> > 
> > 
> > The user should call the Reset method when retrieving a URL structure from the pool.
> 
> You mean that users should handle this manually? This is not a good idea, we should resolve this issue within dubbo-go itself. This approach would only increase the learning curve for users.

You can offer your solution, and then we can discuss it.

PR #2843, opened 2025-04-20
Fix: the issue where adding weight to the nacos option fails

1kasa:
Fix: the issue where adding weight to the nacos option fails

Error description：
When adding weights to the options option, the processing method was hard coded 1

The added constants follow：
image

### DIFF
--- a/registry/nacos/service_discovery.go
+++ b/registry/nacos/service_discovery.go
@@ -315,7 +315,6 @@ func (n *nacosServiceDiscovery) toRegisterInstance(instance registry.ServiceInst
 		metadata = make(map[string]string, 1)
 	}
 
-	fmt.Printf("n.register instance %+v\n", n.registryURL)
 	weightStr := n.registryURL.GetParam(constant.RegistryKey+"."+constant.WeightKey, "1.0")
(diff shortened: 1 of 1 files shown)


registry/nacos/registry.go
```diff
@@ -99,11 +99,22 @@ func createRegisterParam(url *common.URL, serviceName string, groupName string)
 	params[constant.MethodsKey] = strings.Join(url.Methods, ",")
 	common.HandleRegisterIPAndPort(url)
 	port, _ := strconv.Atoi(url.Port)
+
+	weightStr := url.GetParam(constant.WeightKey, "1.0")
+	weight, err := strconv.ParseFloat(weightStr, 64)
+	if err != nil || weight <= constant.MinNacosWeight {
+		logger.Warnf("Invalid weight value %q, using default 1.0. err: %v", weightStr, err)
+		weight = constant.DefaultNacosWeight
+	} else if weight > constant.MaxNacosWeight {
```
No-SilverBullet, line 108: can we extract the logic of processing weight into a separate function? it seems this logic has both been used in registry.go and service_discovery.go

FoghostCn (approved):
LGTM

Gucvii:
LGTM

Issue #2094, opened 2022-10-25
[Enhancement] multi registry center address verification


binbin0325:
**What would you like to be added**:

multi registry center address verification

当配置多个注册中心时，要对地址进行校验，地址不可相同。

**Why is this needed**:
如果相同导致不能获取正确的Registry
registryUrl.Location=注册中心地址
```go
func (proto *registryProtocol) getRegistry(registryUrl *common.URL) registry.Registry {
	var err error
	reg, loaded := proto.registries.Load(registryUrl.Location)
	if !loaded {
		reg, err = extension.GetRegistry(registryUrl.Protocol, registryUrl)
		if err != nil {
			logger.Errorf("Registry can not connect success, program is going to panic.Error message is %s", err.Error())
			panic(err)
		}
		proto.registries.Store(registryUrl.Location, reg)
	}
	return reg.(registry.Registry)
}
```

PR #2862, opened 2025-05-07
Multi registry verification and Use LoadOrStore instead

1kasa:
fix：https://github.com/apache/dubbo-go/issues/2094 Multi registry Address verification 

Rewriting with LoadOrStore here can avoid the problem of duplicate creation under multi-threaded concurrency.

### DIFF
--- a/config/root_config.go
+++ b/config/root_config.go
@@ -18,6 +18,7 @@
 package config
 
 import (
+	"fmt"
 	"sync"
 )
 
@@ -164,6 +165,10 @@ func (rc *RootConfig) Init() error {
 		if err := reg.Init(); err != nil {
 			return err
 		}
+	}
+
+	if err := validateRegistryAddresses(rc.Registries); err != nil {
+		return err
 	}
 
 	if err := rc.MetadataReport.Init(rc); err != nil {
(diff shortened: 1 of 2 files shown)


config/root_config.go
```diff
@@ -367,3 +372,27 @@ func (rc *RootConfig) Process(event *config_center.ConfigChangeEvent) {
...
+		namespace := reg.Namespace
+
+		cacheKey := address
+		if namespace != "" {
+			cacheKey = cacheKey + "?" + constant.NacosNamespaceID + "=" + namespace
+		}
+
+		if existingID, exists := cacheKeyMap[cacheKey]; exists {
+			err := fmt.Errorf("duplicate registry address: [%s] used by both [%s] and [%s]", cacheKey, existingID, id)
+			logger.Error(err)
```
marsevilspirit, line 390: 为什么不直接用logger.Errorf()?
这会导致多引入"fmt" package

config/root_config.go
```diff
@@ -367,3 +372,27 @@ func (rc *RootConfig) Process(event *config_center.ConfigChangeEvent) {
 	// dynamically update metric
 	rc.Metrics.DynamicUpdateProperties(updateRootConfig.Metrics)
 }
+
+func validateRegistryAddresses(registries map[string]*RegistryConfig) error {
+	cacheKeyMap := make(map[string]string)
```
marsevilspirit, line 377: 可以提前分配空间
cacheKeyMap := make(map[string]string, len(registries))

No-SilverBullet:
code looks good to me, have u ever validate this `registry verification` in your local environment?

marsevilspirit:
> > > It's like this. For example, multiple goroutines access a shared cache (such as proto.registries) at the same time, and these goroutines need to obtain the registry object based on the same cacheKey. If we use Load and Store, when multiple goroutines access it at the same time, it may happen that the `if !loaded {}` condition is executed, resulting in two identical Registry objects being stored in the cache, causing duplicate creation. After using LoadOrStore (which itself guarantees concurrent safety), no matter how many goroutines access it at the same time, as long as one goroutine successfully creates and stores the registry object, other goroutines will directly load the object from the cache, and duplicate creation will not occur.
> > 
> > 
> > Oh, I discovered a fatal problem. The config package is a package we are about to discard. We should implement this effect in global package.
> 
> If this is the case,we may need to move the impact of this code to the global package module later.

I think so too.

marsevilspirit (approved):
LGTM!

PR #2865, opened 2025-05-11
delete conf parser type

marsevilspirit:
The parsers of properties create a top-level dependency on config, and config is a package that needs to be removed, so we are trying to eliminate it.


marsevilspirit:
> ```
> // checkFileSuffix check file suffix
> func checkFileSuffix(suffix string) error {
> 	for _, g := range []string{"json", "toml", "yaml", "yml", "properties"} {
> 		if g == suffix {
> 			return nil
> 		}
> 	}
> 	return errors.Errorf("no support file suffix: %s", suffix)
> }
> ```
> 
> 这段也感觉也没有用处了可以一并改了

fixed

marsevilspirit:
This PR will remove the top-level loader's dependency on config, and we have taken another step forward in the direction of removing config.

marsevilspirit:
Why is this PR a delete instead of moving to global?

Because we discussed in WeChat:  
TOML and properties configuration files are rare, so we just deleted them directly.

FoghostCn (approved):

PR #2887, opened 2025-05-20
fix: AddCallerSkip fix for default zap logger

brushax:
Wrong AddCallerSkip arg for default logger after the logger refactorization, which will cause incorrect file level.


marsevilspirit (approved):
LGTM

PR #2896, opened 2025-05-24
feat:  improve URL generation logic

Similarityoung:
While addressing the  #2044 , I discovered that Dubbo-go already supports both interface-level and service-level service discovery for providers. The original problem described in the issue no longer exists as users can simply set `registry-type: all` in their configuration to enable both modes simultaneously,like this:

```
nacos:
  protocol: nacos
  address: 127.0.0.1:8848
  username: nacos
  password: nacos
  namespace: my-ns
  group: dubbo
  registry-type: all  # Enable both discovery modes
  params:
    nacos.logLevel: warn
```

So I just optimized the code related to this, improving its readability.

### DIFF
--- a/config/registry_config.go
+++ b/config/registry_config.go
@@ -59,7 +59,7 @@ type RegistryConfig struct {
 }
 
 // Prefix dubbo.registries
-func (RegistryConfig) Prefix() string {
+func (*RegistryConfig) Prefix() string {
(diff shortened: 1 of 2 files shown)


marsevilspirit:
The config package has been deprecated, please make changes in the global package.

PR #2908, opened 2025-06-01
Optimize: remove deplicate judge available

marsevilspirit:
The logic for determining whether the invoker is available is repeated.

### DIFF
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -97,10 +97,8 @@ func (di *DubboInvoker) Invoke(ctx context.Context, ivc base.Invocation) result.
 		return &res
 	}
 
-	di.clientGuard.RLock()
-	defer di.clientGuard.RUnlock()
-
-	if di.client == nil {
+	client := di.getClient()
+	if client == nil {
 		res.SetError(base.ErrClientClosed)
 		logger.Debugf("result.Err: %v", res.Error())
 		return &res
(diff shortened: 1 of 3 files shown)


protocol/dubbo/dubbo_invoker.go
```diff
@@ -93,24 +93,16 @@ func (di *DubboInvoker) Invoke(ctx context.Context, ivc base.Invocation) result.
...
 	defer di.clientGuard.RUnlock()
 
 	if di.client == nil {
-		res.Err = base.ErrClientClosed
-		logger.Debugf("result.Err: %v", res.Err)
-		return &res
-	}
-
-	if !di.BaseInvoker.IsAvailable() {
-		// Generally, the case will not happen, because the invoker has been removed
```
1kasa, line 110: Can you explain why you want to delete it? If you delete it, will there be a certain risk of causing race conditions in high concurrency scenarios? I hope to get an answer

marsevilspirit: You need to take a look at the complete Invoke method of dubbo Invoker. This function checks available twice, and the Invoke function is called very frequently, so it shouldn't spend extra time checking twice. The triple invoker only checks once, and the dubbo invoker should be synchronized with the triple invoker.

protocol/dubbo/dubbo_invoker.go
```diff
@@ -93,24 +93,16 @@ func (di *DubboInvoker) Invoke(ctx context.Context, ivc base.Invocation) result.
...
 		logger.Warnf("this dubboInvoker is destroyed")
-		res.Err = base.ErrDestroyedInvoker
+		res.SetError(base.ErrDestroyedInvoker)
 		return &res
 	}
 
 	di.clientGuard.RLock()
 	defer di.clientGuard.RUnlock()
 
 	if di.client == nil {
```
No-SilverBullet: i think we can remove this lock  too
use di.getclient  to replace di.client

marsevilspirit: done

No-SilverBullet (commented):
LGTM

PR #2910, opened 2025-06-03
chore: rename invocation

marsevilspirit:
Make the code naming more in line with Go's conventions.

### DIFF
--- a/filter/auth/default_authenticator.go
+++ b/filter/auth/default_authenticator.go
@@ -56,35 +56,35 @@ func newDefaultAuthenticator() filter.Authenticator {
 }
 
 // Sign adds the signature to the invocation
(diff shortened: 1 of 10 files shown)


filter/auth/default_authenticator.go
```diff
@@ -56,16 +56,16 @@ func newDefaultAuthenticator() filter.Authenticator {
 }
 
 // Sign adds the signature to the invocation
-func (authenticator *defaultAuthenticator) Sign(invocation base.Invocation, url *common.URL) error {
+func (authenticator *defaultAuthenticator) Sign(invo base.Invocation, url *common.URL) error {
```
AlexStocks: image

多问下 gpt，改为 inv 吧

marsevilspirit: fixed.

PR #2917, opened 2025-06-19
fix: Fix incorrect index usage in blacklist refresh logic

liushiqi1001:
The TryRefreshBlackList function had a critical bug where it would check the availability of invoker at index j but incorrectly remove the invoker at index i, leading to wrong invokers being removed from the blacklist. This caused frequent and incorrect blacklist operations, resulting in excessive INFO logging and unstable service behavior.

Changes:
- Fix incorrect index usage: RemoveInvokerUnhealthyStatus(ivks[i]) -> RemoveInvokerUnhealthyStatus(ivks[j])
- Optimize import structure for better readability

This ensures the blacklist refresh mechanism works correctly by removing the actual invoker that was checked and found to be available.

Fixes: #XXXX

### DIFF
--- a/protocol/rpc_status.go
+++ b/protocol/rpc_status.go
@@ -21,17 +21,12 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
-)
 
-import (
 	"github.com/dubbogo/gost/log/logger"
 
-	uberAtomic "go.uber.org/atomic"
-)
-
-import (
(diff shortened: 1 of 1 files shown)


marsevilspirit:
Please don't change the order of imports.

marsevilspirit:
The main branch is not our development branch, please make changes on the develop branch and submit a PR.

PR #2920, opened 2025-06-23
feat: Delete all polaris related ability

Alanxtl:
dubbo-go no longer supports polaris(limit/registry/router)


Alanxtl:
the ci fails because dubbo-go-samples still have integration test for polaris, ci need to be retriggered after https://github.com/apache/dubbo-go-samples/pull/860 merge

SkyeBeFreeman:
@Alanxtl Why remove polaris registry？Are there any maintenance difficulties here?

marsevilspirit:
@SkyeBeFreeman We are very sorry. The initial decision to remove it
was because we found that Polaris had not been updated for six months,
and we suspected that no one was maintaining Polaris. However, we
greatly appreciate your prompt communication and your attention to
dubbo-go. We have decided to support Polaris again and give it the
same attention and respect.

We also have some questions about Polaris. If it's convenient, could
you add me on WeChat?  


WeChat: marsevilspirit

PR #2943, opened 2025-07-06
feat(triple): add timeout for graceful stop

marsevilspirit:
## Description

Currently, the `GracefulStop` process for the Triple protocol server lacks a timeout mechanism. This can cause the server shutdown to hang indefinitely if an underlying issue prevents it from completing, which impacts service reliability and automated deployment pipelines.

This PR introduces a timeout for the `GracefulStop` function to ensure that the shutdown process completes within a predictable timeframe.

### Key Changes

- **Added `DefaultGracefulShutdownTimeout` Constant**: A new constant set to 10 seconds has been added in `common/constant/default.go` to provide a centralized and configurable timeout value.
- **Applied Timeout to `GracefulStop`**: The `GracefulStop` method in `protocol/triple/server.go` now uses `context.WithTimeout` to enforce the 10-second shutdown deadline.
- **Added Error Logging**: Implemented error logging to capture and report any potential errors that occur during the graceful shutdown process, improving observability.

### DIFF
--- a/common/constant/default.go
+++ b/common/constant/default.go
@@ -108,7 +108,7 @@ const (
 	DefaultMaxCallRecvMsgSize = 1024 * 1024 * 4
 	DefaultMaxCallSendMsgSize = math.MaxInt32
 
-	// keeplive
+	// keepalive
 	DefaultKeepAliveInterval = "10s"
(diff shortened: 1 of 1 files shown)


No-SilverBullet (approved):
LGTM

PR #2946, opened 2025-07-08
feat(client): add WithInterface option

marsevilspirit:
## description

This pull request introduces a new functional option, `WithInterface`, for the client-side reference configuration.

### Motivation

To provide a clear and idiomatic way for users to configure the remote service they intend to call, this PR adds `WithInterface` following the common functional options pattern. This approach makes the client setup more readable and consistent with modern Go practices.

This configuration is essential for the service discovery mechanism to correctly locate and route requests to the appropriate provider.

### Changes

  - Adds a new exported function `WithInterface(interfaceName string) ReferenceOption`.
  - This function allows the target interface name to be passed directly into the client constructor.

### Usage Example

```go
  svc, err := greet.NewGreetService(
      cli,
      client.WithInterface("com.your.company.GreetService"),
  )
```

### DIFF
--- a/client/action.go
+++ b/client/action.go
@@ -107,7 +107,9 @@ func (refOpts *ReferenceOptions) refer(srv common.RPCService, info *ClientInfo) 
 
 	var methods []string
 	if info != nil {
-		ref.InterfaceName = info.InterfaceName
+		if ref.InterfaceName == "" {
+			ref.InterfaceName = info.InterfaceName
+		}
 		methods = info.MethodNames
 		refOpts.id = info.InterfaceName
 		refOpts.info = info
--- a/client/options.go
+++ b/client/options.go
@@ -166,6 +166,26 @@ func WithFilter(filter string) ReferenceOption {
 	}
 }
 
+// WithInterface sets the interface name for the service reference.
+//
+// As a functional option, it is passed to a client constructor
+// (e.g., NewGreetService) to configure which remote service to connect to.
+//
+// The interfaceName is a crucial identifier for service discovery and routing,
+// and it must exactly match the name registered by the service provider.
+//
+// Usage:
+//
+//	svc, err := greet.NewGreetService(
+//	    cli,
+//	    client.WithInterface("com.your.company.GreetService"),
+//	)
+func WithInterface(interfaceName string) ReferenceOption {
+	return func(opts *ReferenceOptions) {
+		opts.Reference.InterfaceName = interfaceName
+	}
+}
+
 func WithRegistryIDs(registryIDs ...string) ReferenceOption {
 	return func(opts *ReferenceOptions) {
 		if len(registryIDs) > 0 {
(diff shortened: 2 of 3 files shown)


Alanxtl (approved):

client/action.go
```diff
@@ -107,7 +107,9 @@ func (refOpts *ReferenceOptions) refer(srv common.RPCService, info *ClientInfo)
 
 	var methods []string
 	if info != nil {
-		ref.InterfaceName = info.InterfaceName
+		if ref.InterfaceName == "" {
+			ref.InterfaceName = info.InterfaceName
+		}
```
Alanxtl, line 112: the default value of `ref.InterfaceName` sets to `package.service` in IDL generated file, instead of "".

for example in `dubbo-go-samples/helloworld/proto/greet.triple.go`
```go
// NewGreetService constructs a client for the greet.GreetService service.
func NewGreetService(cli *client.Client, opts ...client.ReferenceOption) (GreetService, error) {
	conn, err := cli.DialWithInfo("greet.GreetService", &GreetService_ClientInfo, opts...)
	if err != nil {
		return nil, err
	}
	return &GreetServiceImpl{
		conn: conn,
	}, nil
}
```

the first value of `cli.DialWithInfo` passed in as `ref.InterfaceName`, so `ref.InterfaceName` is always equals to  `info.InterfaceName` if not using your `WithInterface`, and if use `WithInterface` `ref.InterfaceName` is always overwritten by  `info.InterfaceName`, makes `WithInterface` always useless

So what about just delete this line

marsevilspirit: Let me clarify the execution logic here.

As shown in the image above, opts have the highest priority, so WithInterface has the highest priority, so it is **always** useful.

The code execution flow is as follows:

- Inside the cli.DialWithInfo function, the ref configuration object is initially initialized with the default interface name you pointed out. At this moment, the value of ref.InterfaceName is indeed "greet.GreetService".

- Next, the program iterates over the passed opts ...client.ReferenceOption. If the user used WithInterface("custom-interface") when calling, then this Option function will be executed at this time.

- Executing WithInterface will overwrite the value of ref.InterfaceName to "custom-interface".

I do not want to delete this line of code, because this line of code does not only serve the new triple, the old triple may have a situation of "", unless you prove to me that the old triple will not have this problem.

And I have tested the code, this function works very well.

Alanxtl: I've figured out why I failed to pass interface, it seems my code logic problem. Excellent work 👍

client/action.go
```diff
@@ -107,7 +107,9 @@ func (refOpts *ReferenceOptions) refer(srv common.RPCService, info *ClientInfo)
...
 		methods = info.MethodNames
 		refOpts.id = info.InterfaceName
```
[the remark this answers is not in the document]
marsevilspirit: useless.

PR #2985, opened 2025-08-25
fix(apollo): delete the generation of configuration files

marsevilspirit:
fix the generation of configuration files when running unit tests in Apollo.

### DIFF
--- a/config_center/apollo/impl_test.go
+++ b/config_center/apollo/impl_test.go
@@ -24,16 +24,22 @@ import (
 	"strings"
 	"sync"
 	"testing"
+)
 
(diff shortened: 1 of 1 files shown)


No-SilverBullet:
Is this to solve the problem of integration testing or what

marsevilspirit:
> Is this to solve the problem of integration testing or what

When running make test, Apollo's unit test will
Generate redundant configuration files. I added parameters that do not generate redundant configuration files to prevent contamination of the code repository.

1kasa:
After my verification, the issue has been correctly resolved here. However, why wasn't ConfigBackupConfigKey used instead of hardcoding "config-center.isBackupConfig"

Issue #3006, opened 2025-09-06
[BUG] The values ​​of rpcStatus.maxElapsed, rpcStatus.succeededMaxElapsed, and rpcStatus.failedMaxElapsed are not updated strictly.


manzhizhen:
### ✅ 验证清单

- [x] 🔍 我已经搜索过 [现有 Issues](https://github.com/apache/dubbo-go/issues)，确信这不是重复问题

### 🚀 Go 版本

1.23.0

### 📦 Dubbo-go 版本

v3.3.0

### 💾 操作系统

🪟 Windows

### 📝 Bug 描述

The values ​​of rpcStatus.maxElapsed, rpcStatus.succeededMaxElapsed, and rpcStatus.failedMaxElapsed are not updated strictly.

`	
if rpcStatus.maxElapsed < elapsed {
	atomic.StoreInt64(&rpcStatus.maxElapsed, elapsed)
}
`

`		
if rpcStatus.succeededMaxElapsed < elapsed {
	atomic.StoreInt64(&rpcStatus.succeededMaxElapsed, elapsed)
}
`

`
if rpcStatus.failedMaxElapsed < elapsed {
	atomic.StoreInt64(&rpcStatus.failedMaxElapsed, elapsed)
}
`

### 🔄 重现步骤

if you use leastactive loadbalance maybe get this problem

### ✅ 预期行为

rpcStatus.maxElapsed, rpcStatus.succeededMaxElapsed, and rpcStatus.failedMaxElapsed are updated strictly.

### ❌ 实际行为

rpcStatus.maxElapsed, rpcStatus.succeededMaxElapsed, and rpcStatus.failedMaxElapsed are not updated strictly.

### 💡 可能的解决方案

The more rigorous approach is to use CAS. I will add a PR to fix this issue.

AlexStocks:
Your PR weill be welcome. Please attention that the target branch of your PR should be 'develop'. Thank you.

manzhizhen:
> Your PR weill be welcome. Please attention that the target branch of your PR should be 'develop'. Thank you.

ok， I do it ：  https://github.com/apache/dubbo-go/pull/3010

PR #3010, opened 2025-09-06
fix: Ensure that some rpcStatus fields are correctly updated. 

manzhizhen:
Ensure that rpcStatus.maxElapsed, rpcStatus.succeededMaxElapsed, and rpcStatus.failedMaxElapsed are correctly replaced through cas.

Fixes: https://github.com/apache/dubbo-go/issues/3006

### DIFF
--- a/protocol/base/rpc_status.go
+++ b/protocol/base/rpc_status.go
@@ -159,21 +159,42 @@ func endCount0(rpcStatus *RPCStatus, elapsed int64, succeeded bool) {
 	atomic.AddInt32(&rpcStatus.total, 1)
 	atomic.AddInt64(&rpcStatus.totalElapsed, elapsed)
 
-	if rpcStatus.maxElapsed < elapsed {
-		atomic.StoreInt64(&rpcStatus.maxElapsed, elapsed)
-	}
-	if succeeded {
-		if rpcStatus.succeededMaxElapsed < elapsed {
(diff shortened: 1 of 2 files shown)


protocol/base/rpc_status.go
```diff
@@ -159,21 +159,42 @@ func endCount0(rpcStatus *RPCStatus, elapsed int64, succeeded bool) {
 	atomic.AddInt32(&rpcStatus.total, 1)
 	atomic.AddInt64(&rpcStatus.totalElapsed, elapsed)
 
-	if rpcStatus.maxElapsed < elapsed {
-		atomic.StoreInt64(&rpcStatus.maxElapsed, elapsed)
+	for {
```
AlexStocks: 这三段代码类似，何不封装成一个 func？

manzhizhen: > 这三段代码类似，何不封装成一个 func？

有道理，我重新搞搞

AlexStocks:
maybe you can add an ut for your PR.

manzhizhen:
> maybe you can add an ut for your PR.

I try it

marsevilspirit:
I'm afraid this will reduce the performance of dubbo-go RPC calls.

manzhizhen:
> I'm afraid this will reduce the performance of dubbo-go RPC calls.

I understand that 99% of requests will break here:
`oldValue := atomic.LoadInt64(curValue)
if oldValue >= newMaxValue {
       break
}`

PR #3014, opened 2025-09-06
fix: correct log rotation path to prevent file loss

MrSibe:
- use full path (accessLog) instead of fileInfo.Name() in os.Rename
- avoids rename failure when cwd differs from log dir
- prevents log rotation breakage and disk fill-up

Origin issue: https://github.com/apache/dubbo-go/issues/3008


mrsibe:
@AlexStocks PTAL

AlexStocks:
Submit a ut for this PR.

mrsibe:
> Submit a ut for this PR.

No Problem. Thanks for merging.

PR #3015, opened 2025-09-06
fix(auth,triple): replace panics with proper error returns to avoid crash

MrSibe:
- auth: return fmt.Errorf instead of panic when accessKeyStorages missing
- triple: return CodeInternal error instead of panic on nil/invalid resp
- keeps server alive on configuration or runtime misbehavior

Origin issue: https://github.com/apache/dubbo-go/issues/3008


AlexStocks:
Please fix the copilot's comment

mrsibe:
> Please fix the copilot's comment请修复助手的评论

OK, I will fix the function signature and return value.

PR #3018, opened 2025-09-07
fix: reduce config warnings

CAICAIIs:
## Description

This PR fixes RPC method signature validation warnings for health server methods and improves server configuration handling.

## Problem

The following warnings were being generated due to incorrect RPC method signatures:
```
2025-09-04 15:25:16 WARN common/rpc_service.go:382 method Resume of mtype func(health.HealthTripleServer) has wrong number of in out parameters 0; needs exactly 1/2
2025-09-04 15:25:16 WARN common/rpc_service.go:382 method SetServingStatus of mtype func(health.HealthTripleServer, string, triple_health.HealthCheckResponse_ServingStatus) has wrong number of in out parameters 0; needs exactly 1/2
2025-09-04 15:25:16 WARN common/rpc_service.go:382 method Shutdown of mtype func(health.HealthTripleServer) has wrong number of in out parameters 0; needs exactly 1/2
```
These warnings occur because the RPC service validation in `common/rpc_service.go` requires methods to have 1 or 2 return values (with the last one being `error`), but the health server methods had no return values.

Additionally, unnecessary warnings about missing service configurations were being shown when using the new server API without config files.

Related to apache/dubbo-go-samples#910

### DIFF
--- a/common/rpc_service.go
+++ b/common/rpc_service.go
@@ -376,8 +376,8 @@ func suiteMethod(method reflect.Method) *MethodType {
 	// Health helper methods are not RPCs and should be ignored.
 	// They should not to be checked.
 	if mname == "Reference" || mname == "SetGRPCServer" || strings.HasPrefix(mname, "XXX") ||
-		(method.Type.In(0).String() == "*health.HealthTripleServer" && 
-		 (mname == "Resume" || mname == "SetServingStatus" || mname == "Shutdown")) {
+		(method.Type.In(0).String() == "*health.HealthTripleServer" &&
+			(mname == "Resume" || mname == "SetServingStatus" || mname == "Shutdown")) {
 		return nil
 	}
 


protocol/triple/health/healthServer.go
```diff
@@ -117,15 +117,16 @@ func (srv *HealthTripleServer) Watch(ctx context.Context, request *triple_health
 
 // SetServingStatus is called when need to reset the serving status of a service
 // or insert a new service entry into the statusMap.
-func (srv *HealthTripleServer) SetServingStatus(service string, servingStatus triple_health.HealthCheckResponse_ServingStatus) {
+func (srv *HealthTripleServer) SetServingStatus(service string, servingStatus triple_health.HealthCheckResponse_ServingStatus) error {
```
No-SilverBullet: seems error always is nil, why add error return

CAICAIIs: I thought this might be more standard. If this change isn't appropriate, I can revert it and remove the 'error'.

No-SilverBullet: > I thought this might be more standard. If this change isn't appropriate, I can revert it and remove the 'error'.

If the error is always empty, I don't think there is any need to add a return. unless you are implementing an interface

marsevilspirit: Agree.

CAICAIIs: got it

common/rpc_service.go
```diff
@@ -373,8 +373,10 @@ func suiteMethod(method reflect.Method) *MethodType {
 
 	// Reference is used to define service reference, and method with prefix 'XXX' is generated by triple pb tool.
 	// SetGRPCServer is used for pb reflection.
+	// Health helper methods are not RPCs and should be ignored.
 	// They should not to be checked.
-	if mname == "Reference" || mname == "SetGRPCServer" || strings.HasPrefix(mname, "XXX") {
+	if mname == "Reference" || mname == "SetGRPCServer" || strings.HasPrefix(mname, "XXX") ||
+		mname == "Resume" || mname == "Shutdown" || mname == "SetServingStatus" {
```
1kasa: Besides the several methods of health.HealthTripleServer that require special handling, are there any other similar methods? If not, I think I can write something more accurate like this: (method.Type.In(0).String() == "*health.HealthTripleServer") &&
(mname == "Restore" || mname == "SetServingStatus" || mname == "Shutdown")

CAICAIIs: > Besides the several methods of health.HealthTripleServer that require special handling, are there any other similar methods? If not, I think I can write something more accurate like this: (method.Type.In(0).String() == "*health.HealthTripleServer") && (mname == "Restore" || mname == "SetServingStatus" || mname == "Shutdown")

I haven't found any other methods that require special handling besides this one. Your suggestion is good, and I will change it to what you provided,thx.

1kasa (approved):
LGTM

PR #3021, opened 2025-09-10
fix: replace panic with error handling for cluster extension failure

CAICAIIs:
Related to #3008

### DIFF
--- a/registry/protocol/protocol.go
+++ b/registry/protocol/protocol.go
@@ -170,7 +170,9 @@ func (proto *registryProtocol) Refer(url *common.URL) base.Invoker {
 	clusterKey := serviceUrl.GetParam(constant.ClusterKey, constant.DefaultCluster)
 	cluster, err := extension.GetCluster(clusterKey)
 	if err != nil {
-		panic(err)
+		logger.Errorf("consumer service %v get cluster %s error, error message is %s, will return nil invoker!",
+			serviceUrl.String(), clusterKey, err.Error())
+		return nil
 	}
 	invoker := cluster.Join(dic)
 	return invoker


CAICAIIs:
> Because no error is returned here, if it is adjusted to return nil, is there a risk of panic in other referenced places? If so, including unit tests, they also need to be adjusted.

got it. I'll look for other places that reference it.

registry/directory/directory.go
```diff
@@ -367,7 +367,13 @@ func (dir *RegistryDirectory) toGroupInvokers() []protocolbase.Invoker {
 			clusterKey := dir.GetURL().SubURL.GetParam(constant.ClusterKey, constant.DefaultCluster)
 			cluster, err := extension.GetCluster(clusterKey)
 			if err != nil {
-				panic(err)
+				logger.Errorf("directory get cluster %s error, error message is %s, will skip this group",
```
AlexStocks: "error message is %s" 改为 " error message is %w"，酱紫后面参数只用写 err 即可

registry/directory/directory.go
```diff
@@ -367,7 +367,7 @@ func (dir *RegistryDirectory) toGroupInvokers() []protocolbase.Invoker {
 			clusterKey := dir.GetURL().SubURL.GetParam(constant.ClusterKey, constant.DefaultCluster)
 			cluster, err := extension.GetCluster(clusterKey)
 			if err != nil {
-				logger.Errorf("directory get cluster %s error, error message is %s, will skip this group",
+				logger.Errorf("directory get cluster %s error, error message is %w, will skip this group",
```
AlexStocks, line 370: 改为 %w 后，err.Error() 后面的 Error() 可以不用要了

marsevilspirit:
I prefer to use panic to deal with these serious and irreverable problems.

PR #3022, opened 2025-09-10
fix: URL parameter loss issue in SetParams method

CAICAIIs:


marsevilspirit:
Please add more descriptions.

CAICAIIs:
> Please add more descriptions.

got it

CAICAIIs:
> 1. According to the comments, this method is not thread-safe. Is it possible to solve this problem too?
> 2. Why use strings.Join(vs, ",") to concatenate multiple values, can't we assign the value directly? The comma splicing strategy is a little bit strange.
>    for k, values := range m {
>    c.params[k] = values  //like this
>    }

got it, thanks for your code review!

PR #3023, opened 2025-09-11
fix: resolve memory leaks in goroutine management, file handles, and extension caches

WangzJi:
This PR addresses critical memory leaks, goroutine deadlocks, and file handle leaks that could lead to resource exhaustion in production environments. The fixes include comprehensive architectural improvements to the access log filter, extension cache management, and Nacos service discovery components.

Key Changes

Access Log Filter Refactoring (filter/accesslog/)

  - Fix goroutine deadlock: Move context management to Filter struct to reduce global dependencies
  - Add timeout protection: Implement timeout mechanisms for writeLogToFile to prevent deadlocks
  - Improve architecture: Extract goroutine logic to dedicated start/processLogs/drainLogs methods
  - Enhanced error handling: Add panic recovery in processLogs goroutine
  - Clean up globals: Remove unused global variables (shutdownOnce, shutdownCtx, shutdownCancel)

Extension Cache Management (common/extension/)

  - Add unregister methods: Implement proper cleanup methods for filter, loadbalance, and protocol caches
  - Prevent memory accumulation: Enable proper resource cleanup when extensions are no longer needed

Nacos Service Discovery (registry/nacos/)

  - Enhanced cleanup: Improve service discovery cleanup mechanisms to prevent resource leaks
  - Fix mock implementation: Correct Nacos mockClient.Unsubscribe implementation for better testing

Testing Improvements

  - Memory leak validation: Add comprehensive memory leak tests for all affected components
  - Fix variable shadowing: Resolve variable shadow issues in access log filter tests
  - Improve test isolation: Create resetGlobalState helper to reduce test code duplication

Closes: #3025

### DIFF
--- a/filter/accesslog/filter.go
+++ b/filter/accesslog/filter.go
@@ -102,7 +102,7 @@ func newFilter() filter.Filter {
 				ctx:       ctx,
 				cancel:    cancel,
 			}
-			accessLogFilter.start()
+			go accessLogFilter.processLogs()
 		})
 	}
 	return accessLogFilter
@@ -187,11 +187,6 @@ func (f *Filter) buildAccessLogData(_ base.Invoker, invocation base.Invocation) 
 // OnResponse do nothing
 func (f *Filter) OnResponse(_ context.Context, result result.Result, _ base.Invoker, _ base.Invocation) result.Result {
 	return result
-}
-
-// start initializes and starts the background goroutine for processing logs
-func (f *Filter) start() {
-	go f.processLogs()
 }
 
 // processLogs runs in a background goroutine to process log data
--- a/filter/accesslog/filter.go
+++ b/filter/accesslog/filter.go
@@ -61,6 +61,9 @@ const (
 var (
 	once            sync.Once
 	accessLogFilter *Filter
+	shutdownOnce    sync.Once
+	shutdownCtx     context.Context
+	shutdownCancel  context.CancelFunc
 )
 
 func init() {
@@ -84,16 +87,45 @@ func init() {
  * AccessLogFilter is designed to be singleton
  */
 type Filter struct {
-	logChan chan Data
+	logChan   chan Data
+	fileCache map[string]*os.File
+	fileLock  sync.RWMutex
 }
 
 func newFilter() filter.Filter {
 	if accessLogFilter == nil {
 		once.Do(func() {
-			accessLogFilter = &Filter{logChan: make(chan Data, LogMaxBuffer)}
+			shutdownCtx, shutdownCancel = context.WithCancel(context.Background())
+			accessLogFilter = &Filter{
+				logChan:   make(chan Data, LogMaxBuffer),
+				fileCache: make(map[string]*os.File),
+			}
 			go func() {
-				for accessLogData := range accessLogFilter.logChan {
-					accessLogFilter.writeLogToFile(accessLogData)
+				defer func() {
+					// Drain remaining log data on shutdown
+					for {
+						select {
+						case accessLogData, ok := <-accessLogFilter.logChan:
+							if !ok {
+								return
+							}
+							accessLogFilter.writeLogToFile(accessLogData)
+						default:
+							return
+						}
+					}
+				}()
+
+				for {
+					select {
+					case accessLogData, ok := <-accessLogFilter.logChan:
+						if !ok {
+							return
+						}
+						accessLogFilter.writeLogToFile(accessLogData)
+					case <-shutdownCtx.Done():
+						return
+					}
 				}
 			}()
 		})
@@ -190,7 +222,7 @@ func (f *Filter) writeLogToFile(data Data) {
 		return
 	}
 
-	logFile, err := f.openLogFile(accessLog)
+	logFile, err := f.getOrOpenLogFile(accessLog)
 	if err != nil {
 		logger.Warnf("Can not open the access log file: %s, %v", accessLog, err)
 		return
@@ -204,12 +236,52 @@ func (f *Filter) writeLogToFile(data Data) {
 	}
 }
 
+// getOrOpenLogFile gets or opens the log file with proper caching and handle management
+func (f *Filter) getOrOpenLogFile(accessLog string) (*os.File, error) {
+	f.fileLock.RLock()
+	if logFile, exists := f.fileCache[accessLog]; exists {
+		// Check if we need to rotate the log
+		now := time.Now().Format(FileDateFormat)
+		if fileInfo, err := logFile.Stat(); err == nil {
+			last := fileInfo.ModTime().Format(FileDateFormat)
+			if now == last {
+				f.fileLock.RUnlock()
+				return logFile, nil
+			}
+		}
+	}
+	f.fileLock.RUnlock()
+
+	// Need to open new file or rotate existing one
+	f.fileLock.Lock()
+	defer f.fileLock.Unlock()
+
+	// Double-check after acquiring write lock
+	if logFile, exists := f.fileCache[accessLog]; exists {
+		now := time.Now().Format(FileDateFormat)
+		if fileInfo, err := logFile.Stat(); err == nil {
+			last := fileInfo.ModTime().Format(FileDateFormat)
+			if now == last {
+				return logFile, nil
+			}
+		}
+		// Close the old file before rotation
+		logFile.Close()
+		delete(f.fileCache, accessLog)
+	}
+
+	logFile, err := f.openLogFile(accessLog)
+	if err != nil {
+		return nil, err
+	}
+
+	f.fileCache[accessLog] = logFile
+	return logFile, nil
+}
+
 // openLogFile will open the log file with append mode.
 // If the file is not found, it will create the file.
 // Actually, the accessLog is the filename
-// You may find out that, once we want to write access log into log file,
-// we open the file again and again.
-// It needs to be optimized.
 func (f *Filter) openLogFile(accessLog string) (*os.File, error) {
 	logFile, err := os.OpenFile(accessLog, os.O_CREATE|os.O_APPEND|os.O_RDWR, LogFileMode)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


marsevilspirit:
Please create an issue for the pr you wrote, and record in detail what problems you have solved and the ideas for solving the problem.

filter/accesslog/filter.go
```diff
@@ -84,18 +84,25 @@ func init() {
  * AccessLogFilter is designed to be singleton
  */
 type Filter struct {
-	logChan chan Data
+	logChan      chan Data
+	fileCache    map[string]*os.File
+	fileLock     sync.RWMutex
```
No-SilverBullet: put file-lock on fileCache to identify the protected object

filter/accesslog/filter.go
```diff
@@ -182,6 +189,68 @@ func (f *Filter) OnResponse(_ context.Context, result result.Result, _ base.Invo
...
+func (f *Filter) writeLogToFileWithTimeout(data Data, timeout time.Duration) {
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		f.writeLogToFile(data)
+	}()
+
+	select {
+	case <-done:
+		// Success
```
No-SilverBullet: maybe we can add a success log here

filter/accesslog/filter.go
```diff
@@ -204,12 +273,52 @@ func (f *Filter) writeLogToFile(data Data) {
...
+	f.fileLock.RLock()
+	if logFile, exists := f.fileCache[accessLog]; exists {
+		// Check if we need to rotate the log
+		now := time.Now().Format(FileDateFormat)
+		if fileInfo, err := logFile.Stat(); err == nil {
+			last := fileInfo.ModTime().Format(FileDateFormat)
+			if now == last {
+				f.fileLock.RUnlock()
+				return logFile, nil
+			}
```
No-SilverBullet: this logic can be extract to a func

filter/accesslog/filter.go
```diff
@@ -204,12 +273,52 @@ func (f *Filter) writeLogToFile(data Data) {
...
+	if logFile, exists := f.fileCache[accessLog]; exists {
+		now := time.Now().Format(FileDateFormat)
+		if fileInfo, err := logFile.Stat(); err == nil {
+			last := fileInfo.ModTime().Format(FileDateFormat)
+			if now == last {
+				return logFile, nil
+			}
+		}
+		// Close the old file before rotation
+		logFile.Close()
```
No-SilverBullet: check error

filter/accesslog/filter.go
```diff
@@ -204,12 +273,52 @@ func (f *Filter) writeLogToFile(data Data) {
 	}
 }
 
+// getOrOpenLogFile gets or opens the log file with proper caching and handle management
+func (f *Filter) getOrOpenLogFile(accessLog string) (*os.File, error) {
```
No-SilverBullet, line 282: Is the reason for judging twice here to reduce the time of writing lock?

WangzJi: Yes, exactly! This is the Double-Checked Locking pattern to minimize write lock contention.

Why two checks:
  1. First check (read lock): Fast path - most requests find existing valid files and return immediately
  2. Second check (write lock): Prevents race conditions - ensures another goroutine didn't already create/rotate the file while we were acquiring the write lock

filter/accesslog/filter.go
WangzJi: Hi, @No-SilverBullet.
Good suggestion! However, I have concerns about adding an Info-level success log here due to potential log volume.

Analysis:
- This method is called for every RPC request that has access logging enabled
- In high-traffic scenarios, this would generate massive amounts of success logs
- Could lead to log file bloat and I/O pressure

Would you prefer adding a Debug-level log instead?

No-SilverBullet: > Hi, @No-SilverBullet. Good suggestion! However, I have concerns about adding an Info-level success log here due to potential log volume.
> 
> Analysis:
> 
> * This method is called for every RPC request that has access logging enabled
> * In high-traffic scenarios, this would generate massive amounts of success logs
> * Could lead to log file bloat and I/O pressure
> 
> Would you prefer adding a Debug-level log instead?

Your idea is very thoughtful, but it also reminds me that it seems that it would be a good feature to introduce the configuration of log sampling rate and batch writing to reduce io in the future.

filter/accesslog/filter.go
No-SilverBullet: > Yes, exactly! This is the Double-Checked Locking pattern to minimize write lock contention.
> 
> Why two checks:
> 
> 1. First check (read lock): Fast path - most requests find existing valid files and return immediately
> 2. Second check (write lock): Prevents race conditions - ensures another goroutine didn't already create/rotate the file while we were acquiring the write lock

Great idea, I think DCLP should have a significant performance improvement when used here.  Excellent work, thanks for your contribution :)

filter/accesslog/filter.go
```diff
@@ -214,6 +189,68 @@ func (f *Filter) OnResponse(_ context.Context, result result.Result, _ base.Invo
 	return result
 }
 
+// start initializes and starts the background goroutine for processing logs
+func (f *Filter) start() {
```
AlexStocks: 把原来的 go func() 改为 start() + processLogs()，为何不是把所有代码放进 start() 同一个函数里面？

filter/accesslog/filter.go
```diff
@@ -363,23 +400,32 @@ func (d *Data) toLogMessage() string {
...
+		// Close all cached file handles
+		f.fileLock.Lock()
+		for path, file := range f.fileCache {
+			if err := file.Close(); err != nil {
+				logger.Warnf("Error closing access log file %s: %v", path, err)
 			}
-			accessLogFilter.fileLock.Unlock()
+			delete(f.fileCache, path)
 		}
+		f.fileLock.Unlock()
```
AlexStocks: 可以改为 defer

filter/accesslog/filter.go
WangzJi: @AlexStocks 感谢review，分离主要考虑一下因素：
  1. 职责分离：`start()` 负责启动，`processLogs()` 负责核心逻辑，单一职责更清晰
  2. 可测试性：可以独立测试 `processLogs()` 的业务逻辑，无需启动`goroutine`
  3. 可读性：函数名语义明确，比单个大函数更易理解

filter/accesslog/filter.go
WangzJi: Done in [96a574d](https://github.com/apache/dubbo-go/pull/3023/commits/96a574da30d0a002f2da7b66011d51e30538b414)

filter/accesslog/filter.go
AlexStocks: 你可以只保留 processLogs，然后在调用的地方直接调用 processLogs，这也不影响写 ut

WangzJi: @AlexStocks Done in [911a47e](https://github.com/apache/dubbo-go/pull/3023/commits/911a47ec976ba16489c531449ebb788152df5f2a).

marsevilspirit:
I see that this pr allows extension to unregister, but extension should be registered once later, and will not be created frequently, which will not lead to memory leakage. Can you answer my question?

filter/accesslog/filter.go
```diff
@@ -182,6 +189,63 @@ func (f *Filter) OnResponse(_ context.Context, result result.Result, _ base.Invo
...
+		f.drainLogs()
+	}()
+
+	for {
+		select {
+		case accessLogData, ok := <-f.logChan:
+			if !ok {
+				return
+			}
+			f.writeLogToFileWithTimeout(accessLogData, 5*time.Second)
```
LQyt2012, line 207: Could you make the timeout parameter configurable or a member of the filter struct if a system always takes more than 5 seconds to write logs to the file, there will be no logs written to the file.

registry/nacos/service_discovery.go
```diff
@@ -86,6 +86,23 @@ func (n *nacosServiceDiscovery) Destroy() error {
...
+			ServiceName: serviceName,
+			GroupName:   n.group,
+		})
+		if err != nil {
+			logger.Warnf("Failed to unsubscribe from service %s: %v", serviceName, err)
+		}
+	}
+	// Clear the listener map
+	n.instanceListenerMap = make(map[string]*gxset.HashSet)
+	n.listenerLock.Unlock()
```
LQyt2012: Use defer to unlock to make sure the lock always be released.

filter/accesslog/filter.go
WangzJi: Hi, @LQyt2012. Thank you for the valuable feedback!
I agree that making the timeout configurable would be useful in edge cases with higher I/O latency.
However, I suggest handling this as a separate enhancement PR, so that this PR stays focused:
- The current 5s timeout is sufficient for most production scenarios (writes usually complete within milliseconds).
- Making it configurable would require broader changes (filter struct, config parsing, URL params, etc.).

registry/nacos/service_discovery.go
WangzJi: @LQyt2012 PTAL.
Thanks for the suggestion! I’ve updated the code in [b3d0846](https://github.com/apache/dubbo-go/pull/3023/commits/b3d08468a4a8e70a3d62be1e9f8efd65098414e4).

WangzJi:
> I see that this pr allows extension to unregister, but extension should be registered once later, and will not be created frequently, which will not lead to memory leakage. Can you answer my question?

Hi @marsevilspirit, Sorry for the late response.
You make a very valid point, in typical usage, extensions are registered once and won't be created frequently, so they shouldn't cause memory leaks.

image

Added the unregistration functionality mainly to address the issue mentioned in #3008 : global extension cache lacks expiration cleanup mechanism. While the impact is limited under current usage patterns, this is a preventive fix that provides foundational support for potential dynamic extension scenarios.


The implementation cost is very low, and we've also added `memory_leak_test.go` to verify the functionality works correctly.

Issue #2741, opened 2024-11-04
Package dependency optimization


FoghostCn:
1. Remove all dependencies on the config package 解除所有对 config 包的依赖关系
2. Remove all dependencies on the client/server package 解除所有对 client/server 包的依赖关系

FoghostCn:
@solisamicus @lqyt2012

FoghostCn:
有些依赖可能是不必要的，甚至是错误的，需要先梳理一下模块的依赖关系再决定要怎么去解决，有些可以用接口解决，有些需要挪动一些代码

marsevilspirit:
我认为这个工作非常有意义，应该继续推进下去

marsevilspirit:
当前依赖config package的模块:
- [x] dubbo.apache.org/dubbo-go/v3 eazy
- [x] dubbo.apache.org/dubbo-go/v3/cluster/router/affinity zeruige doing
- [x] dubbo.apache.org/dubbo-go/v3/cluster/router/polaris zeruige doing
- [x] dubbo.apache.org/dubbo-go/v3/cluster/router/tag zeruige doing
- [x] dubbo.apache.org/dubbo-go/v3/common/extension eazy
- [x] dubbo.apache.org/dubbo-go/v3/filter/graceful_shutdown 雪涛 doing
- [x] dubbo.apache.org/dubbo-go/v3/filter/polaris/limit 雪涛
- [x] dubbo.apache.org/dubbo-go/v3/protocol/dubbo @marsevilspirit 凤瑞 doing
- [x] dubbo.apache.org/dubbo-go/v3/protocol/dubbo3 @marsevilspirit 凤瑞 doing
- [x] dubbo.apache.org/dubbo-go/v3/protocol/grpc @marsevilspirit 凤瑞 doing
- [x] dubbo.apache.org/dubbo-go/v3/protocol/jsonrpc @marsevilspirit 凤瑞 doing
- [x] dubbo.apache.org/dubbo-go/v3/protocol/rest/config/reader @marsevilspirit 凤瑞 doing
- [x] dubbo.apache.org/dubbo-go/v3/protocol/triple @marsevilspirit 凤瑞 doing
- [x] dubbo.apache.org/dubbo-go/v3/registry/directory siruige doing
- [x] dubbo.apache.org/dubbo-go/v3/registry/protocol siruige doing
- [x] dubbo.apache.org/dubbo-go/v3/remoting/getty xiaoyuge doing
[...]

PR #3054, opened 2025-10-17
Using global to replace config in protocol

nanjiek:
@https://github.com/apache/dubbo-go/issues/2741   
Replace config with global while ensuring compatibility in protocol.go
protocol_test.go still have some problems about TestExportWithApplicationConfig and TestGetProviderUrlWithHideKey

### DIFF
--- a/registry/protocol/protocol.go
+++ b/registry/protocol/protocol.go
@@ -582,7 +582,7 @@ func newProviderConfigurationListener(overrideListeners *sync.Map, url *common.U
 	listener := &providerConfigurationListener{}
 	listener.overrideListeners = overrideListeners
 
-	// Fallback to config package (old API)
+	//TODO: Temporary compatibility with old APIs, can be removed later
 	if rootConfig := config.GetRootConfig(); rootConfig != nil && rootConfig.Application != nil {
 		listener.InitWith(
 			rootConfig.Application.Name+constant.ConfiguratorSuffix,


nanjiek:
### 1. Where the error occurs

The error happens in `registry/base_configuration_listener.go:65`, inside the method `BaseConfigurationListener.InitWith()`:

```go
if rawConfig, err := bcl.dynamicConfiguration.GetRule(key,
    config_center.WithGroup(constant.Dubbo)); err != nil {
    bcl.configurators = []config_center.Configurator{}
    return
} else if len(rawConfig) > 0 {
    if err := bcl.genConfiguratorFromRawRule(rawConfig); err != nil {
        logger.Error("bcl.genConfiguratorFromRawRule(rawConfig:%v) = error:%v", rawConfig, err)
    }
}
```

---

### 2. Root cause — configuration format mismatch

The **core issue** is a **format mismatch**.

In `config_center/mock_dynamic_config.go` (lines 59–82), the `MockDynamicConfiguration` is initialized with content in **properties format**:

```go
dynamicConfiguration.content = `
    dubbo.consumer.request_timeout=5s
    dubbo.consumer.connect_timeout=5s
    dubbo.application.organization=ikurento.com
    ...
`
```

However, in `config_center/parser/configuration_parser.go`, the `ParseToUrls()` method (line 118) expects the content to be in **YAML format** for the `ConfiguratorConfig` structure:

```go
func (parser *DefaultConfigurationParser) ParseToUrls(content string) ([]*common.URL, error) {
    config := ConfiguratorConfig{}
    if err := yaml.Unmarshal([]byte(content), &config); err != nil {
        return nil, err
    }
    // ...
}
```

---

### 3. Error trigger flow

Here’s how the error happens step by step:

1. **Initialization phase**:
   The test calls `exporterNormal()` → `regProtocol.Export()` → which initializes `BaseConfigurationListener`.

2. **Configuration retrieval**:
   `BaseConfigurationListener.InitWith()` calls `GetRule()` to fetch the existing configuration from the mock config center.

3. **Incorrect format returned**:
   `MockDynamicConfiguration.GetRule()` returns the **properties-format** content.

4. **Parsing failure**:
   Inside `genConfiguratorFromRawRule()`, it calls `ParseToUrls()`, which tries to parse the properties string as YAML — causing a parsing error.

---

### 4. Why the test still passes

Even though an error is logged, the test **passes** successfully for several reasons:

1. **Error is gracefully handled**
   The failure in `genConfiguratorFromRawRule()` is only logged using `logger.Error`, not returned or propagated, so the test execution continues.

2. **Subsequent correct configurations**
   The test later calls `MockServiceConfigEvent()` and `MockApplicationConfigEvent()`, which send properly formatted **YAML configurations**:

   ```go
   func (c *MockDynamicConfiguration) MockServiceConfigEvent() {
       config := &parser.ConfiguratorConfig{
           ConfigVersion: "2.7.1",
           Scope:         parser.GeneralType,
           Key:           mockServiceName,
           Enabled:       true,
           Configs: []parser.ConfigItem{...},
       }
       value, _ := yaml.Marshal(config)  // Correct YAML format
       // ...
   }
   ```

3. **Final configuration takes effect**
   Through the configuration event mechanism, the correctly formatted YAML content overrides the faulty one from initialization.
   As a result, the test assertions still succeed even though the earlier parsing error occurred.

---

registry/protocol/protocol.go
```diff
@@ -269,7 +291,34 @@ func registerServiceMap(invoker base.Invoker) error {
 	// such as dubbo://:20000/org.apache.dubbo.UserProvider?bean.name=UserProvider&cluster=failfast...
 	id := providerUrl.GetParam(constant.BeanNameKey, "")
 
-	serviceConfig := config.GetProviderConfig().Services[id]
+	if providerConfRaw, ok := providerUrl.GetAttribute(constant.ProviderConfigPrefix); ok {
```
KamToHung, line 315: This part of the logic if it is better to first check the config and then the global

registry/protocol/protocol.go
```diff
@@ -185,12 +186,33 @@ func (proto *registryProtocol) Refer(url *common.URL) base.Invoker {
...
 func (proto *registryProtocol) Export(originInvoker base.Invoker) base.Exporter {
-	proto.once.Do(func() {
-		proto.initConfigurationListeners()
-	})
 	registryUrl := getRegistryUrl(originInvoker)
 	providerUrl := getProviderUrl(originInvoker)
 
+	if _, ok := registryUrl.GetAttribute(constant.ShutdownConfigPrefix); !ok {
+		// Only set default global config when config package doesn't have one
+		if config.GetShutDown() == nil {
```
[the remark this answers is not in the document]
nanjiek: The logic of advice is wrong.

registry/protocol/protocol.go
nanjiek: I have solved it.

registry/protocol/protocol.go
```diff
@@ -518,14 +578,29 @@ type providerConfigurationListener struct {
...
+func newProviderConfigurationListener(overrideListeners *sync.Map, url *common.URL) *providerConfigurationListener {
 	listener := &providerConfigurationListener{}
 	listener.overrideListeners = overrideListeners
-	listener.InitWith(
-		config.GetRootConfig().Application.Name+constant.ConfiguratorSuffix,
-		listener,
-		extension.GetDefaultConfiguratorFunc(),
-	)
+
+	// Fallback to config package (old API)
```
Alanxtl:
```suggestion
	//TODO: Temporary compatibility with old APIs, can be removed later
```

registry/protocol/protocol.go
```diff
@@ -269,26 +292,48 @@ func registerServiceMap(invoker base.Invoker) error {
...
-	serviceConfig := config.GetProviderConfig().Services[id]
-	if serviceConfig == nil {
-		s := "reExport can not get serviceConfig"
-		return perrors.New(s)
-	}
-	rpcService := config.GetProviderService(id)
-	if rpcService == nil {
-		s := "reExport can not get RPCService"
-		return perrors.New(s)
+	providerConfig := config.GetProviderConfig()
```
Alanxtl, line 295:
```suggestion
    //TODO: Temporary compatibility with old APIs, can be removed later
	providerConfig := config.GetProviderConfig()
```

registry/protocol/protocol.go
```diff
@@ -424,9 +469,24 @@ func (proto *registryProtocol) Destroy() {
...
-			exporter.UnExport()
-			proto.bounds.Delete(key)
+			if configShutdown := config.GetShutDown(); configShutdown != nil {
+				<-time.After(configShutdown.GetStepTimeout() + configShutdown.GetConsumerUpdateWaitTime())
+				exporter.UnExport()
+				proto.bounds.Delete(key)
+				return
+			}
+
+			if shutdownConfRaw, ok := exporter.registerUrl.GetAttribute(constant.ShutdownConfigPrefix); ok {
```
Alanxtl, line 479: need to make sure is ShutdownConfig already set into url or not

nanjiek:
```
	if _, ok := registryUrl.GetAttribute(constant.ShutdownConfigPrefix); !ok {
		// Only set default global config when config package doesn't have one
		if config.GetShutDown() == nil {
			registryUrl.SetAttribute(constant.ShutdownConfigPrefix, global.DefaultShutdownConfig())
		}
	}   
```   In the front ， it has judgement.

Alanxtl:
why closed

nanjiek:
I will move the change to the fork of KamToHung

PR #3062, opened 2025-10-25
fix(test): relax EOF assertion for concurrent triple stream tests

zbchi:
Some stream tests occasionally hit EOF due to concurrent stream close.
When the server closes the stream right after receiving HTTP headers and invoking CumSum,
the client may still be sending and thus read EOF. This is expected behavior.

The related test assertions have been relaxed to tolerate expected EOFs. All tests pass stably now.
Related to #3058


marsevilspirit:
It is recommended to add comments to the code added first.

Issue #2890, opened 2025-05-22
采用nacos注册中心，通过go泛化调用dubbo接口，无法通过标签路由选择服务地址


alan2016llg:
### Environment

- Server: dubbo 
- Client: Dubbo-go ,v3.3.0
- Protocol: dubbo
- Registry: Nacos, v2.3.2

### Issue description

go端关键代码：
```go
ctx := context.WithValue(context.Background(), constant.AttachmentKey, map[string]interface{}{"dubbo.tag": "p6666"})
	resp, err := refConf.GetRPCService().(*generic.GenericService).Invoke(
		//context.TODO(),
		ctx,
		methodName,
		types,
		args,
	)
```

### Logs

<details><summary>Click me to check logs</summary>

```
Copy logs to here.
```

</details>

No-SilverBullet:
我们正在跟进这个 @marsevilspirit

Alanxtl:
one of the causes is solved by https://github.com/apache/dubbo-go/pull/3069

Alanxtl:
wait for router new api

Alanxtl:
Hi @alan2016llg  ,
I think this issue has been solved by https://github.com/apache/dubbo-go/pull/3185. If there are any problems, please let us know.

PR #3069, opened 2025-11-06
feat: inject dubbo.tag into service metadata

zbchi:
when users configured
``` go
	dubbo.WithTag("newtag"),
	...,
)
```
Tag was not written into metadata.
As a result, registries could not see the tag field, and tag router rules could not match services by tag.

## What’s Changed

This PR adds a built-in tagCustomizer under register/servicediscovery/customizer/,
which automatically injects dubbo.tag into the instance metadata during service export
``` go
if instance.GetTag() != "" {
    instance.GetMetadata()[constant.Tagkey] = instance.GetTag()
}
```

This ensures the tag value appears in the registry metadata, for example:

Before
```
dubbo.metadata-service.url-params={"group":"dubbo_registry_nacos_server","port":"43775","protocol":"dubbo"}
dubbo.metadata.storage-type=local
id=172.20.10.3:20000
meta-v=1.0.0
```
After
```
dubbo.metadata-service.url-params={"group":"dubbo_registry_nacos_server","port":"43775","protocol":"dubbo"}
dubbo.metadata.storage-type=local
id=172.20.10.3:20000
meta-v=1.0.0
dubbo.tag=newtag
```
related to #2890

### DIFF
--- a/registry/servicediscovery/customizer/service_instance_tag_customizer_test.go
+++ b/registry/servicediscovery/customizer/service_instance_tag_customizer_test.go
@@ -22,10 +22,12 @@ import (
 )
 
 import (
+	"github.com/stretchr/testify/assert"
+)
+
+import (
 	"dubbo.apache.org/dubbo-go/v3/common/constant"
 	"dubbo.apache.org/dubbo-go/v3/registry"
-
-	"github.com/stretchr/testify/assert"
 )
 
 func TestTagCustomizer(t *testing.T) {


registry/servicediscovery/customizer/service_instance_tag_customizer.go
```diff
@@ -0,0 +1,42 @@
...
+func (t *tagCustomizer) GetPriority() int {
+	return 2
+}
+
+// Customize injects the tag value into instance metadata
+func (t *tagCustomizer) Customize(instance registry.ServiceInstance) {
+	if instance.GetTag() != "" {
+		instance.GetMetadata()[constant.Tagkey] = instance.GetTag()
+	}
+}
```
Alanxtl, line 44:
```suggestion
func (t *tagCustomizer) Customize(instance registry.ServiceInstance) {
        tag := instance.GetTag()
	if tag == "" {
		return
	}
	instance.GetMetadata()[constant.Tagkey] = tag
}
```

zbchi:
> cool solution.
> 
> 1. need to write a unit test file for this
> 2. need to recheck the problem discussed in the issue has been solved

ok,
the pr fixed one of the causes , the missing tag in the registry.
The issue itself hasn’t been fully resolved and is still under investigation.

PR #3080, opened 2025-11-17
feat: Allow using group/version wildcards when the interface matches exactly

Akashisang:
xref #3074

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -28,15 +28,22 @@ import (
 	"strings"
 	"sync"
 	"time"
+)
 
+import (
 	cm "github.com/Workiva/go-datastructures/common"
 
 	gxset "github.com/dubbogo/gost/container/set"
+
 	"github.com/google/uuid"
+
 	"github.com/jinzhu/copier"
 
-	"dubbo.apache.org/dubbo-go/v3/common/constant"
 	perrors "github.com/pkg/errors"
+)
+
+import (
+	"dubbo.apache.org/dubbo-go/v3/common/constant"
 )
 
 // dubbo role type constant
--- a/common/url_test.go
+++ b/common/url_test.go
@@ -21,9 +21,14 @@ import (
 	"encoding/base64"
 	"net/url"
 	"testing"
+)
 
-	"dubbo.apache.org/dubbo-go/v3/common/constant"
+import (
 	"github.com/stretchr/testify/assert"
+)
+
+import (
+	"dubbo.apache.org/dubbo-go/v3/common/constant"
 )
 
 const (


common/url.go
```diff
@@ -477,8 +477,13 @@ func ParseServiceKey(serviceKey string) (string, string, string) {
...
-	return intf == constant.AnyValue && (group == constant.AnyValue ||
-		group == serviceURL.Group()) && (version == constant.AnyValue || version == serviceURL.Version())
+	matchCondition := func(pattern, actual string) bool {
+		return pattern == constant.AnyValue || pattern == actual
+	}
+
+	return matchCondition(intf, serviceURL.Service()) &&
+		matchCondition(group, serviceURL.Group()) &&
+		matchCondition(version, serviceURL.Version())
 }
```
Alanxtl, line 487: are you sure java judges in this logic? Can you paste java code here, so that we can compare it.

Akashisang: I think it should be partly correspond to `UrlUtils.isMatch`, for this function takes on more responsibilities in java dubbo. I think it is an equal edit in mentioned calling context
```
public static boolean isMatch(URL consumerUrl, URL providerUrl) {
        String consumerInterface = consumerUrl.getServiceInterface();
        String providerInterface = providerUrl.getServiceInterface();

        // FIXME accept providerUrl with '*' as interface name, after carefully thought about all possible scenarios I
        // think it's ok to add this condition.

        // Return false if the consumer interface is not equals the provider interface,
        // except one of the interface configurations is equals '*' (i.e. any value).
        if (!(ANY_VALUE.equals(consumerInterface)
                || ANY_VALUE.equals(providerInterface)
                || StringUtils.isEquals(consumerInterface, providerInterface))) {
            return false;
        }

        // If the category of provider URL does not match the category of consumer URL.
        // Usually, the provider URL's category is empty, and the default category ('providers') is present.
        // Hence, the category of the provider URL is 'providers'.
        // Through observing of debugging process, I found that the category of the consumer URL is
        // 'providers,configurators,routers'.
        if (!isMatchCategory(providerUrl.getCategory(DEFAULT_CATEGORY), consumerUrl.getCategory(DEFAULT_CATEGORY))) {
            return false;
        }

        // If the provider is not enabled, return false.
        if (!providerUrl.getParameter(ENABLED_KEY, true) && !ANY_VALUE.equals(consumerUrl.getParameter(ENABLED_KEY))) {
            return false;
        }

        // Obtain consumer's group, version and classifier.
        String consumerGroup = consumerUrl.getGroup();
        String consumerVersion = consumerUrl.getVersion();
        String consumerClassifier = consumerUrl.getParameter(CLASSIFIER_KEY, ANY_VALUE);

        // Obtain provider's group, version and classifier.
        String providerGroup = providerUrl.getGroup();
        String providerVersion = providerUrl.getVersion();
        String providerClassifier = providerUrl.getParameter(CLASSIFIER_KEY, ANY_VALUE);

        // If Group, Version, Classifier all matches, return true.
        boolean groupMatches = ANY_VALUE.equals(consumerGroup)
                || StringUtils.isEquals(consumerGroup, providerGroup)
                || StringUtils.isContains(consumerGroup, providerGroup);
        boolean versionMatches =
                ANY_VALUE.equals(consumerVersion) || StringUtils.isEquals(consumerVersion, providerVersion);
        boolean classifierMatches = consumerClassifier == null
                || ANY_VALUE.equals(consumerClassifier)
                || StringUtils.isEquals(consumerClassifier, providerClassifier);

        return groupMatches && versionMatches && classifierMatches;
}
```
calling context in dubbo-go `registry/zookeeper/listener.go`
```
func (l *RegistryDataListener) DataChange(event remoting.Event) bool {
	...
	for serviceKey, listener := range l.subscribed {
		intf, group, version := common.ParseServiceKey(serviceKey)
		if serviceURL.ServiceKey() == serviceKey || common.IsAnyCondition(intf, group, version, serviceURL) {
			listener.Process(
				&config_center.ConfigChangeEvent{
					Key:        event.Path,
					Value:      serviceURL.Clone(),
					ConfigType: event.Action,
				},
			)
			match = true
		}
	}
	return match
}
```
same scenario in java in `RegistryProtocol.java`
```
private List<URL> getMatchedUrls(List<URL> configuratorUrls, URL currentSubscribe) {
            List<URL> result = new ArrayList<>();
            for (URL url : configuratorUrls) {
                URL overrideUrl = url;
                // Compatible with the old version
                if (url.getCategory() == null && OVERRIDE_PROTOCOL.equals(url.getProtocol())) {
                    overrideUrl = url.addParameter(CATEGORY_KEY, CONFIGURATORS_CATEGORY);
                }

                // Check whether url is to be applied to the current service
                if (UrlUtils.isMatch(currentSubscribe, overrideUrl)) {
                    result.add(url);
                }
            }
            return result;
}
```

PR #3098, opened 2025-12-02
fix: optimization of Serve in server/server.go (#3042)

y138g:
* fix: resolved the issue of the Serve() method causing the program to hang indefinitely through graceful_shutdown [#3042](https://github.com/apache/dubbo-go/issues/3042)

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -19,9 +19,12 @@
 package server
 
 import (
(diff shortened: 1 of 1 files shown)


CAICAIIs:
please commit into apache:develop

PR #3141, opened 2025-12-22
test: add unit test for protocol/result 

Akashisang:


marsevilspirit:
The protocol/result.go file is only about 100 lines long, but your unit test code is 8 times longer than that of result.go?

PR #3184, opened 2026-01-28
fix(protocol): fix panic caused by race condition during invoker destruction

HGD-coder:
### Description
Fixes #3153

#### 1. The Issue (Race Condition)
As described in #3153, a panic (`runtime error: invalid memory address`) occurs under high concurrency when `ReferenceConfig.Destroy()` is triggered:
1. `Destroy()` sets `bi.url = nil` to release resources.
2. Concurrent calls to `GetURL()` return `nil`.
3. Upper layers (Router, LoadBalance, Cluster, Filter) access methods on the returned `nil` URL (e.g., `url.GetParam()`), causing the panic.

#### 2. Solution: Null Object Pattern
This PR introduces a **Null Object Pattern** to fix the panic safely:
- **Global Sentinel**: Adds a thread-safe `emptyURL` initialized via `common.NewURLWithOptions`.
- **Safe Access**: Modifies `GetURL()` to return `emptyURL` instead of `nil` when `bi.url` is destroyed.

#### 3. Why we chose this solution (Comparison of Alternatives)
We evaluated three potential solutions. Here is why the Null Object Pattern is the best choice:

**Option A: Remove `bi.url = nil` in Destroy()**
- **Idea**: Keep the old URL even after destruction.
- **Problem**: This creates **"Zombie Invokers"**. Upper layers would still see a valid Protocol/Host/Port and continue routing traffic to a closed client, causing `use of closed network connection` errors. It also prevents the URL (and its large Map) from being GC'd, causing potential **memory leaks**.

**Option B: Add `if url != nil` checks at every call site**
- **Idea**: Modify all callers (Router, Cluster, LoadBalance, etc.) to check for nil.
- **Problem (Unfeasible)**:
    - **"Whac-A-Mole" Risk**: `GetURL()` is a fundamental method used in **50+ files**. Missing just one spot means the panic persists.
    - **Maintenance Nightmare**: It forces every future developer to remember "Check for nil after GetURL", increasing cognitive load.

**Option C: Null Object Pattern (This PR)**
- **Result**: `GetURL()` never returns nil.
- **Benefit**: Callers don't need to change. The `emptyURL` returns empty parameters, so upper layers (like Router) will naturally filter out this destroyed invoker (e.g., matching params fails gracefully).
- **Cost**: Negligible memory (**~429 Bytes** singleton, verified by benchmark).

#### 4. Verification
| Approach | Bug #3153 Reproduction | Result |
| :--- | :--- | :--- |
[...]

PR #3230, opened 2026-03-02
fix(filter):solve side effects caused by 3184 refactor

HGD-coder:
### Description
This PR fixes a side effect (active count leak) introduced by the `BaseInvoker` refactor in #3184.

**Background & Issue:**
In #3184, `BaseInvoker.GetURL()` was modified to return an `emptyURL` instead of `nil` when the invoker is concurrently destroyed. However, in `filter/active/filter.go`, if the destruction happens during an ongoing invocation, `BeginCount` and `EndCount` will operate on different URL objects (the original URL vs. the `emptyURL`). This mismatch causes the active count to leak, as the original URL's count is never decremented.

**Solution:**
To ensure consistency, this PR caches the original URL pointer in the `Invocation` attachments during the `Invoke` phase, so `OnResponse` can retrieve and use the exact same URL object for `EndCount`.

### Checklist


Alanxtl (approved):
LGTM

PR #3245, opened 2026-03-13
Update dependencies and restore etcd API

### LINKED ISSUE #3243: [UPDATE] Restore the previously removed API for etcd/grpc compatibility/update
### Description

The API / behavior was originally removed during the changes in:
- apache/dubbo-go#3046

Later, we found this caused downstream compatibility problems for Dubbo-Go:
- apache/dubbo-go#3070

In particular, removing the `grpc.WithBlock()`-related behavior in `gost` affected the initialization semantics along this dependency chain:

`dubbo-go` -> `gost` -> `etcd/clientv3` -> `grpc`

As discussed in #3070, this broke the expected behavior in the etcd registry path and caused unit test failures in Dubbo-Go.

Now the upstream etcd side has already updated the related implementation / discussion:
- etcd-io/etcd#20910
- https://github.com/etcd-io/etcd/issues/20910#issuecomment-4048775213

***So I think it is time to add this API / behavior back in `gost`.***

### Motivation

The original removal was understandable at that time, but it introduced real downstream compatibility issues.

Since the upstream dependency has now moved forward, restoring this API / behavior would help downstream projects such as Dubbo-Go remove temporary compatibility workarounds and recover the expected client initialization semantics.

### References

- Original removal: apache/dubbo-go#3046
- Compatibility regression discussion: apache/dubbo-go#3070
- Upstream etcd update: etcd-io/etcd#20910

utafrali:
### Description
Fixes #3243

Bumped the dependencies to newer versions. The main thing is restoring the etcd API in the client that was previously removed - it's needed for grpc compatibility.

### Checklist


remoting/etcdv3/client.go
```diff
@@ -36,7 +40,7 @@ func ValidateClient(container clientFacade, opts ...gxetcd.Option) error {
 
 	// new Client
 	if container.Client() == nil {
-		newClient, err := gxetcd.NewClient(options.Name, options.Endpoints, options.Timeout, options.Heartbeat)
+		newClient, err := gxetcd.NewClientWithOptions(context.Background(), options)
```
AlexStocks, line 43: All three `NewClientWithOptions` call sites use `context.Background()`, which creates an uncancellable context. `ValidateClient` is called under `container.ClientLock()`, so if etcd is unreachable and the dial blocks, the lock is held indefinitely — freezing the entire application.

`NewClientWithOptions` internally calls `context.WithCancel(ctx)`, so cancellation from the caller side is fully supported but unused here.

**Suggestion**: At minimum, use `context.WithTimeout(context.Background(), options.Timeout)` to bound the blocking time:
```go
ctx, cancel := context.WithTimeout(context.Background(), options.Timeout)
defer cancel()
newClient, err := gxetcd.NewClientWithOptions(ctx, options)
```

Ideally, accept a `context.Context` parameter in `ValidateClient` and `NewServiceDiscoveryClient` to give callers full control over cancellation.

AlexStocks:
`metadata/report/etcd/report.go:120` — This PR migrates `remoting/etcdv3/client.go` from `NewClient` to `NewClientWithOptions` (which supports TLS, username, password), but `metadata/report/etcd/report.go` still uses the old `NewClient` API:

```go
client, err := gxetcd.NewClient(gxetcd.MetadataETCDV3Client, addresses, timeout, 1)
```

This inconsistency means metadata report will not benefit from TLS/auth support, and if a future gost version removes `NewClient`, this file breaks.

**Suggestion**: Update to `NewClientWithOptions` for consistency:
```go
client, err := gxetcd.NewClientWithOptions(context.Background(), &gxetcd.Options{
    Name:      gxetcd.MetadataETCDV3Client,
    Endpoints: addresses,
    Timeout:   timeout,
    Heartbeat: 1,
})
```

AlexStocks:
The PR checklist states "I confirm the target branch is develop", but this PR actually targets `main`. Please confirm the intended target branch — if the project convention is to merge into `develop` first, this PR should be retargeted.

AlexStocks:
> The PR checklist states "I confirm the target branch is develop", but this PR actually targets `main`. Please confirm the intended target branch — if the project convention is to merge into `develop` first, this PR should be retargeted.

I have changed the target branch to develop.

Alanxtl:
@CAICAIIs check this

utafrali:
Done, pushed the fix.

Alanxtl:
actually, what you should do is to modify https://github.com/dubbogo/gost instead of dubbo-go repo.
The api we deleted in https://github.com/dubbogo/gost/pull/131 and added back in https://github.com/dubbogo/gost/pull/134.

What you should do is to do extally the same as https://github.com/dubbogo/gost/pull/131

remoting/etcdv3/client.go
AlexStocks, line 43: 这里硬编码了 `context.Background()`。建议使用 `context.TODO()` 表明这是临时的，或者考虑添加一个 context 参数让调用者控制。

remoting/etcdv3/client.go
```diff
@@ -47,7 +51,7 @@ func ValidateClient(container clientFacade, opts ...gxetcd.Option) error {
 
 	// Client lose connection with etcd server
 	if container.Client().GetRawClient() == nil {
-		newClient, err := gxetcd.NewClient(options.Name, options.Endpoints, options.Timeout, options.Heartbeat)
+		newClient, err := gxetcd.NewClientWithOptions(context.Background(), options)
```
AlexStocks, line 54: 同上，硬编码 `context.Background()`。

remoting/etcdv3/client.go
```diff
@@ -68,7 +72,7 @@ func NewServiceDiscoveryClient(opts ...gxetcd.Option) *gxetcd.Client {
 		opt(options)
 	}
 
-	newClient, err := gxetcd.NewClient(options.Name, options.Endpoints, options.Timeout, options.Heartbeat)
+	newClient, err := gxetcd.NewClientWithOptions(context.Background(), options)
```
AlexStocks, line 75: 同上，硬编码 `context.Background()`。

CAICAIIs:
Please resolve the conflict when you have time~

utafrali:
@Alanxtl Thanks for pointing out that these changes should go in the [dubbogo/gost](https://github.com/dubbogo/gost) repo instead. You're right, that's the proper place for this fix. Closing this PR. Apologies for the noise.

PR #3258, opened 2026-03-16
migrate the ValidateRegistryIDs to internal

nanjiek:
### Description
Fixes # (issue)
Delete the validateRegistryIDs in server/options and client/options. Move the function into internal to avoid dry problem.

### Checklist

### DIFF
--- a/client/options.go
+++ b/client/options.go
@@ -18,7 +18,6 @@
 package client
 
 import (
-	"fmt"
 	"strconv"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


internal/config.go
```diff
@@ -206,3 +206,12 @@ func ValidateMethodConfig(method *global.MethodConfig) error {
 
 	return nil
 }
+
+func ValidateRegistryIDs(ids []string, regs map[string]*global.RegistryConfig) error {
+	for _, id := range ids {
+		if _, ok := regs[id]; !ok {
+			return fmt.Errorf("registry id %q not found", id)
```
AlexStocks, line 213: 从 client 和 server 两边各删了一份一模一样的 `validateRegistryIDs`，收到 internal 包里统一维护，消除了重复代码。函数逻辑没变，测试也覆盖到了 valid 和 missing 两种 case，没问题。

另外 client/server 那边删掉函数后，多余的 `fmt` import 也一并清理了，细节到位。

PR #3282, opened 2026-03-27
Fix/3246 url memory leaks

### LINKED ISSUE #3246: [BUG] URL struct memory leaks: attributes never deleted, listeners never cleaned, blacklist key mismatch
### Verification Checklist

- [x] I have searched the [existing Issues](https://github.com/apache/dubbo-go/issues) and believe this is not a duplicate

### Go Version

1.24

### Dubbo-go Version

v3.3.1

### OS

Linux / All platforms

### Bug Description

After a thorough code audit of the `common.URL` struct and its lifecycle management across the codebase, I identified **10 memory leak points** (1 P0, 5 P1, 4 P2). These leaks cause unbounded memory growth in long-running services, especially in environments with frequent service registration/deregistration or configuration changes.

https://github.com/apache/dubbo-go/discussions/2764

---

## P0: `registryProtocol` singleton leaks `overrideListeners` and `serviceConfigurationListeners`

**File**: `registry/protocol/protocol.go:59-70, 457-496`

```go
type registryProtocol struct {
    overrideListeners             *sync.Map  // leaked
    serviceConfigurationListeners *sync.Map  // leaked
}
```

`Export()` (L218, L221) stores listeners into these maps, but `Destroy()` (L457-496) only cleans `bounds` and `registries` — **never** cleans `overrideListeners` or `serviceConfigurationListeners`. Each `overrideSubscribeListener` holds a reference to the full `originInvoker` (including URL and connection resources). Since `registryProtocol` is a package-level singleton (L50 `regProtocol`), these references are never GC'd.

**Impact**: Every exported service leaks its listener + invoker + URL + connections. In microservice environments with frequent service lifecycle changes, memory grows proportionally to total services ever exported.

**Fix**: Add `Range + Delete` cleanup for both maps in `Destroy()`. Also `Delete` old listeners in `reExport`.

---

## P1-1: URL `attributes` are write-only — no `DeleteAttribute` method exists

**File**: `common/url.go:583-590`

```go
func (c *URL) SetAttribute(key string, value any) {
    c.attributes[key] = value  // write-only, no delete API
}
```

The entire codebase has zero calls to delete attributes. Values stored include large objects: `*global.ApplicationConfig`, `map[string]*global.RegistryConfig`, `ServiceInfo`, `RPCService` implementations. The `filter/adaptivesvc/filter.go:96` writes an updater on **every RPC call**.

**Fix**: Add `DeleteAttribute(key string)`. Clean attributes on invoker/exporter Destroy.

---

## P1-2: `RegistryDirectory.Destroy()` doesn't clean `cacheInvokersMap`

**File**: `registry/directory/directory.go:549-573`

```go
func (dir *RegistryDirectory) Destroy() {
    dir.DoDestroy(func() {
        invokers := dir.cacheInvokers
        dir.cacheInvokers = []protocolbase.Invoker{}
        for _, ivk := range invokers {
            ivk.Destroy()
        }
        // cacheInvokersMap (sync.Map) is NEVER cleaned!
    })
}
```

The `sync.Map` still holds references to all destroyed invokers + their URLs. If `RegistryDirectory` is held by the singleton `registryProtocol`, these references leak permanently.

**Fix**: Add `dir.cacheInvokersMap.Range(func(k, _ any) bool { dir.cacheInvokersMap.Delete(k); return true })`.

---

## P1-3: `invokerBlackList` key mismatch prevents cleanup

**File**: `protocol/base/rpc_status.go:40, 213`

```go
// Store uses URL.Key()
func SetInvokerUnhealthyStatus(invoker Invoker) {
    invokerBlackList.Store(invoker.GetURL().Key(), invoker)
}

// Delete uses GetCacheInvokerMapKey() — different format!
func RemoveUrlKeyUnhealthyStatus(key string) {
    invokerBlackList.Delete(key)
}
```

`SetInvokerUnhealthyStatus` uses `URL.Key()` as the map key, but `RemoveUrlKeyUnhealthyStatus` receives keys from `GetCacheInvokerMapKey()` which has a **different format**. The Delete never matches, so blacklisted invokers (holding full invoker + URL + connections) are **never removed**.

**Fix**: Unify key generation. Or store only the key string, not the full invoker object.

---

## P1-4: `GetParams()` returns internal map reference without lock or copy

**File**: `common/url.go:657-659`

```go
func (c *URL) GetParams() url.Values {
    return c.params  // no lock, no copy
}
```

Callers hold the internal `params` reference indefinitely, preventing URL GC. Callers also modify it directly (e.g., `base_configuration_listener.go:111` does `delete(override, constant.AnyhostKey)` on the returned map), which is a concurrent mutation without any lock.

**Fix**: Return a copy, or deprecate in favor of `CopyParams()`.

---

## P1-5: `configurators` slice only grows, never shrinks

**File**: `registry/directory/directory.go:379`

```go
dir.configurators = append(dir.configurators, extension.GetDefaultConfigurator(ret))
```

Every override/configurator event appends a new `Configurator` (holding a `*common.URL`). Never replaced or trimmed. In long-running systems with frequent config pushes, this slice grows unbounded.

**Fix**: Replace entire slice on new config (like `BaseConfigurationListener.Process` does at L88), not append.

---

## P2-1: `MergeURL` reads `anotherUrl.attributes` without lock

**File**: `common/url.go:879-891`

```go
for attrK, attrV := range anotherUrl.attributes {  // no attributesLock!
    if _, ok := mergedURL.GetAttribute(attrK); !ok {
        mergedURL.attributes[attrK] = attrV
    }
}
```

This is both a data race and a reference leak (shallow copy extends object lifetimes).

---

## P2-2: `GetCacheInvokerMapKey` creates a temporary URL on every call

**File**: `common/url.go:418-428`

```go
func (c *URL) GetCacheInvokerMapKey() string {
    urlNew, _ := NewURL(c.PrimitiveURL)  // allocates a full URL object
    // just to extract timestamp
}
```

Called on every `ServiceEvent.Key()` in hot paths. Not a leak, but unnecessary GC pressure.

---

## P2-3: `AddParam` vs `SetParam` semantic confusion

**File**: `common/url.go:537-543`, `registry/protocol/protocol.go:400`

```go
// isMatched() calls this repeatedly, accumulating values under same key
providerUrl.AddParam(constant.CategoryKey, constant.ConfiguratorsCategory)
```

`url.Values.Add` appends (not overwrites). Each `isMatched` call adds another CategoryKey value, causing `params[CategoryKey]` to grow linearly.

**Fix**: Use `SetParam` instead of `AddParam` in `isMatched`.

---

## P2-4: `registryProtocol.Destroy()` missing else branch in async cleanup

**File**: `registry/protocol/protocol.go:457-496`

```go
go func() {
    if configShutdown := config.GetShutDown(); configShutdown != nil {
        // cleanup path 1
    }
    if shutdownConfRaw, ok := ...; ok {
        // cleanup path 2
    }
    // NO else: if both conditions fail, goroutine exits without UnExport or Delete
}()
```

**Fix**: Add else branch that unconditionally UnExport + Delete.

---

### Summary

| # | Issue | Severity | Type |
|---|-------|----------|------|
| 1 | overrideListeners/serviceConfigurationListeners never cleaned | P0 | Leak |
| 2 | attributes write-only, no delete API | P1 | Design |
| 3 | cacheInvokersMap not cleaned on Destroy | P1 | Leak |
| 4 | invokerBlackList key mismatch | P1 | Leak |
| 5 | GetParams() returns internal reference | P1 | Leak + Race |
| 6 | configurators only appended, never replaced | P1 | Leak |
| 7 | MergeURL attributes access without lock | P2 | Race + Leak |
| 8 | GetCacheInvokerMapKey allocates temp URL | P2 | GC pressure |
| 9 | AddParam/SetParam confusion | P2 | Accumulation |
| 10 | Destroy async cleanup missing else | P2 | Conditional leak |

### Suggested Fix Priority

1. Fix P0 first (listener maps cleanup in `registryProtocol.Destroy`)
2. Add `DeleteAttribute` API and clean attributes on Destroy
3. Unify blacklist key generation
4. Fix `GetParams()` to return copy
5. Replace `append` with full replacement for configurators

nagisa-kunhah:
### Description
Fixes #3246

### Checklist


Alanxtl:
our project use `import-formatter` to format import blocks, that's the reason why ur CI fails. For you, u should 

1. run `go install github.com/dubbogo/tools/cmd/imports-formatter@latest`
2. cd to the root dir of `dubbo-go`
3. run `imports-formatter`

common/url.go
```diff
@@ -102,6 +96,7 @@ type URL struct {
 	Port     string
 
 	PrimitiveURL string
+	primitiveTS  string
```
Alanxtl: what does TS stand for?

common/url.go
```diff
@@ -596,6 +590,20 @@ func (c *URL) GetAttribute(key string) (any, bool) {
...
+	defer c.attributesLock.Unlock()
+	if c.attributes != nil {
+		delete(c.attributes, key)
+	}
+}
+
+func (c *URL) ClearAttributes() {
+	c.attributesLock.Lock()
+	defer c.attributesLock.Unlock()
+	c.attributes = nil
```
Alanxtl: isn't it better to set `c.attributes` into a empty map instead of `nil` ?

common/url.go
```diff
@@ -653,9 +661,29 @@ func (c *URL) GetNonDefaultParam(s string) (string, bool) {
...
+
+	params := make(url.Values, len(c.params))
+	for k, vs := range c.params {
+		copied := make([]string, len(vs))
+		copy(copied, vs)
+		params[k] = copied
+	}
+
+	return params
 }
```
Alanxtl, line 694: every time I want to `GetParams()`, CopyParams() will generate a new map, it is consuming

common/url.go
nagisa-kunhah: ok, fix it

common/url.go
nagisa-kunhah: primitiveTS just caches the original timestamp from c.PrimitiveURL. Previously, GetCacheInvokerMapKey() reparsed PrimitiveURL and used that value, not the current mutable params. This keeps the same behavior and only removes the extra allocation.
https://github.com/apache/dubbo-go/blob/fb8eff2f090851261e2d1acfc55414ede91ca9c0/common/url.go#L418-L428

common/url.go
nagisa-kunhah: Since the issue points out that callers may retain the internal params map and mutate it without locking, returning it directly is unsafe. So GetParams() now returns a copy; if a caller only needs a single value, it should use GetParam() instead.

nagisa-kunhah:
here are some confusement about this issue
1. For P1–5, my understanding is that the original behavior may have been to append the new config and apply it together with the existing one. However, based on BaseConfigurationListener.Process, it seems the behavior there is to replace the existing config entirely. Which behavior is actually expected here?
2. For P2-2, the original behavior in isMatched is to append. Replacing it with SetParam may change that behavior. Is that expected?
@Alanxtl @AlexStocks

common/url.go
Alanxtl: cool

registry/directory/directory.go
```diff
@@ -250,34 +250,49 @@ func (dir *RegistryDirectory) refreshInvokers(event *registry.ServiceEvent) {
 // not in the incoming list can be removed.  The Action of serviceEvent should be EventTypeUpdate or EventTypeAdd.
 func (dir *RegistryDirectory) refreshAllInvokers(events []*registry.ServiceEvent, callback func()) {
```
Alanxtl, line 300: In `refreshAllInvokers`, a batch containing only configurator URLs now returns early after `replaceConfigurators(...)`. This updates `dir.configurators`, but seems to skip rebuilding/reapplying overrides to the existing cached invokers.
Could this delay configurator-only updates until the next provider event? It may be worth adding a test that starts with cached provider invokers, sends a configurator-only `NotifyAll`, and verifies the effective invoker URLs are refreshed immediately.

Alanxtl:
> here are some confusement about this issue
> 
> 1. For P1–5, my understanding is that the original behavior may have been to append the new config and apply it together with the existing one. However, based on BaseConfigurationListener.Process, it seems the behavior there is to replace the existing config entirely. Which behavior is actually expected here?
> 2. For P2-2, the original behavior in isMatched is to append. Replacing it with SetParam may change that behavior. Is that expected?
>    @Alanxtl @AlexStocks

as for 2, do you mean P2-3? 
If the goal is only to avoid duplicate growth while preserving append semantics, another option would be to append only when configurators is not already present.

AlexStocks (commented):
## Code Review: fix/3246 url memory leaks

### 评价

内存泄漏修复方案合理，通过引入 `primitiveTS` 字段避免重复解析 URL，并通过 `CopyParams()` 返回深拷贝解决并发访问问题。

### [P1] 规范级问题

**1. `common/url.go:307-320` - deprecated 方法设计问题**

`GetParams()` 标记为 deprecated 但实现已改为返回深拷贝，这可能导致：
- 调用方误以为返回的是原始 map（修改不会影响原数据）
- 现有代码无需修改即可安全使用

**建议**：考虑直接移除 deprecated 标记，因为当前实现是安全的。或者明确文档说明：
```go
// GetParams returns a deep copy of params.
// Deprecated: Use CopyParams() for clarity. This method now also returns a copy.
func (c *URL) GetParams() url.Values {
    return c.CopyParams()
}
```

**2. `registry/directory/directory.go:372-388` - 重复代码**

`refreshAllInvokers` 中的 `configuratorURLs` 处理逻辑与 `convertUrl` 存在重复，建议抽取公共方法：
```go
func (dir *RegistryDirectory) processServiceEvent(event *registry.ServiceEvent) (*common.URL, bool) {
    // 返回 URL 和是否为 configurator 标志
}
```

**3. `common/url_test.go` - 并发测试验证**

新增的 `TestMergeURLAttributeAccessRace` 测试验证了并发访问，但仅运行 50ms，建议延长测试时间或增加迭代次数。

---

总体评价：修复方向正确，代码质量良好，建议优化 deprecated 方法的文档说明。

common/url.go
```diff
@@ -102,6 +102,7 @@ type URL struct {
 	Port     string
 
 	PrimitiveURL string
+	primitiveTS  string
```
No-SilverBullet: Add a comment here, "primitiveTS caches......"

No-SilverBullet (approved):
LGTM :)

AlexStocks:
文件都冲突了

Alanxtl:
pls resolve file confilct

AlexStocks:
please fix the file confliction.

nagisa-kunhah:
> please fix the file confliction.

Thanks for the reminder. I’ve been a bit busy these past few days, and I’ll resolve the file conflict shortly.

common/url.go
nagisa-kunhah: fix it

registry/directory/directory.go
nagisa-kunhah: I think there are two possible behaviors for configurator-only `NotifyAll`, and I’d like to confirm which one we want before changing the rebuild logic.

Example:

- original provider URL:
  `dubbo://10.0.0.8:20880/com.foo.UserService?group=g1&version=1.0.0&serialization=fastjson&warmup=100`
- consumer reference URL:
  `dubbo://127.0.0.1/com.foo.UserService?group=g1&version=1.0.0&timeout=3000&cluster=failover`

Then suppose we receive one configurator-only batch:

- first batch:
  `override://0.0.0.0/com.foo.UserService?timeout=2s&serialization=hessian2`

and later another configurator-only batch:

- next batch:
  `override://0.0.0.0/com.foo.UserService?cluster=mock3`

If we rebuild from the original provider URL each time, the final effective URL would become:

`dubbo://10.0.0.8:20880/com.foo.UserService?group=g1&version=1.0.0&serialization=fastjson&warmup=100&timeout=3000&cluster=mock3`

If we instead re-apply the new configurator batch on top of the current effective invoker URL, the final URL would become:

`dubbo://10.0.0.8:20880/com.foo.UserService?group=g1&version=1.0.0&serialization=hessian2&warmup=100&timeout=2s&cluster=mock3`

Since this patch changed configurator handling from append to replace-by-batch, I’m leaning toward the first behavior, but I’d like to confirm the expected semantics first.

registry/directory/directory.go
```diff
@@ -271,34 +271,49 @@ func (dir *RegistryDirectory) refreshInvokers(event *registry.ServiceEvent) {
...
+			continue
+		}
+		providerEvents = append(providerEvents, event)
+	}
+	if configuratorURLs != nil {
+		dir.replaceConfigurators(configuratorURLs)
+	}
+	defer callback()
+	if len(providerEvents) == 0 && len(events) > 0 {
+		return
```
AlexStocks, line 295: [P1] 这里在 configurator-only 的 `NotifyAll` 场景下直接 `return`，只替换了 `dir.configurators`，但没有把新规则重新应用到已经缓存的 invoker。结果就是这批 override 要等下一次 provider 事件或配置中心事件触发后才会真正生效，Alan 之前提的那个语义问题还在。建议这里要么遍历 `cacheInvokersMap` 按原始 provider URL 重新 `overrideUrl + doCacheInvoker`，要么至少补一个覆盖现有 cached invoker 立即刷新行为的测试，确保纯 configurator 批次不会静默延迟。

Alanxtl:
pls fix conflict

registry/directory/directory.go
nagisa-kunhah: @Alanxtl Hi, could you please take a look at my question above when you have a chance? Thanks!

Alanxtl: For `NotifyAll`, I think we should treat the incoming configurator set as the authoritative current snapshot.

So for a configurator-only `NotifyAll`, my expectation is:
- rebuild from the original provider URL
- then apply the current configurator snapshot
- not re-apply on top of the current effective invoker URL

Otherwise rules that disappeared from the latest snapshot would keep leaking into the effective URL, which makes `replace-by-batch` non-idempotent.

So in your example, I would expect the first behavior, not the layered one.

That said, I think this is separate from the original review point: even if we choose snapshot semantics, a configurator-only `NotifyAll` should still recompute the existing cached invokers immediately. Returning early after `replaceConfigurators(...)` means the new snapshot does not take effect until the next provider event, which seems incorrect for `NotifyAll`.

A test that starts with cached provider invokers, sends a configurator-only `NotifyAll`, and verifies the invokers are rebuilt from the original provider URLs against the new configurator snapshot would probably make the intended behavior explicit.

registry/directory/directory.go
```diff
@@ -372,7 +387,13 @@ func (dir *RegistryDirectory) cacheInvokerByEvent(event *registry.ServiceEvent)
...
+			if isConfiguratorURL(event.Service) {
+				dir.replaceConfigurators([]*common.URL{event.Service})
+			}
 			u := dir.convertUrl(event)
+			if u == nil && dir.cacheOriginUrl == nil {
+				return nil, nil
+			}
 			logger.Infof("[Registry Directory] selector add service url{%s}", event.Service)
 			if u != nil && constant.RouterProtocol == u.Protocol {
 				dir.configRouters()
```
Alanxtl, line 399: The single-event path now replaces `dir.configurators` with a slice containing only the latest configurator URL. If the registry sends override rules as incremental add/update events, each new rule wipes out the previously active ones, so only the last configurator remains effective. This breaks multi-rule override scenarios even though the new tests do not cover them.

registry/protocol/protocol.go
```diff
@@ -275,6 +275,13 @@ func (proto *registryProtocol) reExport(invoker base.Invoker, newUrl *common.URL
...
+
+		oldProviderURL := getProviderUrl(invoker)
+		oldOverrideURL := getSubscribedOverrideUrl(oldProviderURL)
+		oldOverrideKey := oldOverrideURL.String()
+		proto.overrideListeners.Delete(oldOverrideKey)
+		proto.serviceConfigurationListeners.Delete(oldProviderURL.ServiceKey())
+
 		// oldExporter UnExport function unRegister rpcService from the serviceMap, so need register it again as far as possible
 		if err := registerServiceMap(invoker); err != nil {
 			logger.Error(err.Error())
```
Alanxtl, line 286: `reExport` removes the old override/service listeners from local `sync.Maps`, but the old listener was already passed to `reg.Subscribe` and remains owned by the registry subscription goroutine. That means the listener and its captured invoker/protocol can still be retained, so the memory leak is only partially fixed; stale override events may also continue to hit the old listener. The old listener should be loaded before deletion and passed to `reg.UnSubscribe`(oldOverrideURL, oldListener) where the registry supports it; Destroy has the same cleanup gap for exporter.subscribeUrl.

common/url.go
```diff
@@ -653,9 +667,29 @@ func (c *URL) GetNonDefaultParam(s string) (string, bool) {
...
+
+	params := make(url.Values, len(c.params))
+	for k, vs := range c.params {
+		copied := make([]string, len(vs))
+		copy(copied, vs)
+		params[k] = copied
+	}
+
+	return params
 }
```
Alanxtl, line 693: This PR changes `GetParams()` to return a deep copy, but `ToConfigurators` still does `override := url.GetParams()`; `delete(override, constant.AnyhostKey)`. That delete no longer mutates the configurator URL. For override URLs containing anyhost plus real config params, anyhost will now remain on the configurator URL and can be written into provider/consumer URLs when Configure calls CloneExceptParams(...).SetParams(...). 

What about update `ToConfigurators` to clone/filter the URL explicitly before building the configurator?

AlexStocks (approved):

registry/protocol/protocol.go
nagisa-kunhah: ok, fixed by unsubscribing the old override listener during reExport and cleaning up the subscribed override listener during Destroy.

common/url.go
nagisa-kunhah: ok, updated ToConfigurators to clone the URL, remove anyhost from the clone, and build the configurator from that filtered URL

Alanxtl (approved):
lgtm, cool work :)

PR #3301, opened 2026-04-20
feat: support mounting existing HTTP handlers on the Triple listener

### LINKED ISSUE #2847: 支持 HTTP 服务迁移到 dubbo-go
希望支持现有的不依赖 IDL 定义的 HTTP 服务迁移到 dubbo-go。
**Why is this needed**:
对于新服务或者是原有的基于 IDL 的微服务迁移到 dubbo-go，都可以完整地利用 dubbo 的生态。
但是如果原有微服务是基于 HTTP 这种无协议文件的，我们也想能够统一生态，迁移到 dubbo。
请问是否可以支持？

zbchi:
## Description

 #2847.

   Add a minimal, transport-scoped way to host an existing root `http.Handler` on the Triple listener.

  This change does not try to turn plain HTTP into a first-class dubbo-go service model in one step. It focuses on the transport-layer integration needed to make incremental HTTP-to-Triple migration possible.

  With this change:

  - an existing HTTP service can be hosted on the same listener as Triple
  - both `mount-first` and `export-first` flows are supported
  - HTTP-only startup on a Triple listener is supported
  - Triple routes still take priority over mounted HTTP traffic
  - mounted HTTP traffic acts as a transport-level fallback after Triple route lookup

  ## Design

  ### 1. Add an optional protocol capability

  Mounting an existing HTTP handler is not a generic server configuration concern.
  It is only meaningful for protocols that are actually implemented on top of `net/http`.

  For that reason, this PR introduces an optional protocol capability instead of pushing an `http.Handler` into generic config structs:

  ```go
  type HTTPHandlerMountable interface {
  	MountHTTPHandler(url *common.URL, handler http.Handler) error
  }
  ```
  This keeps HTTP mounting scoped to protocols that can actually host HTTP
  traffic, rather than leaking transport-specific runtime state into generic
  server configuration.

  ### 2. Model mounting as a single root handler

  Mounting is intentionally modeled as a single root http.Handler.

  That decision is deliberate. dubbo-go owns one shared listener, while plain HTTP route composition should remain in user space. If a caller needs to expose multiple HTTP services, they can compose them behind their own http.ServeMux
  or router and mount that mux once.

  This keeps the boundary clear:

  dubbo-go owns listener lifecycle and Triple routing
  user code owns plain HTTP route composition

  ### 3. Reuse Triple as the transport host

  This pr makes mount-first and export-first converge on the same shared
  listener lifecycle.

  This allows:

  1. mounting an HTTP handler first, then exporting Triple services later
  2. exporting Triple services first, then mounting an HTTP fallback later
  3. running a mounted HTTP service on a Triple listener even when no Triple service is exported yet

  This avoids introducing a separate startup model just for mounted HTTP traffic.

  ### 4. Preserve Triple route priority

  Mounted HTTP traffic is intentionally implemented as a transport-level fallback,
  not as a competing routing layer.

  The route mux resolves requests in this order:

  1. exact Triple procedure match
  2. method-name case-adapted Triple fallback
  3. mounted HTTP fallback

  A simplified version of the serving logic is:

```go
  func (m *methodRouteMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
  	handler, pattern := m.Handler(r)
  	if pattern != "" {
  		handler.ServeHTTP(w, r)
  		return
  	}
  	if fallback := m.fallback; fallback != nil {
  		fallback.ServeHTTP(w, r)
  		return
  	}
  	http.NotFound(w, r)
  }
```
  This ordering is important: mounting an existing HTTP service must not preempt
  registered Triple procedures on the same listener.

  ### 5. Validate listener-fixed transport settings

  When the same Triple listener is reused, transport-level settings must remain
  consistent.

  This pr validates and rejects conflicts for listener-fixed settings such as:

  HTTP/2 vs HTTP/3 transport mode
  TLS configuration

  Handler-level Triple options may still vary per exported service.

## Usage
```go
  	mux := http.NewServeMux()
  	mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
  		_, _ = w.Write([]byte("ok"))
  	})
  	mux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
  		_, _ = w.Write([]byte("ready"))
  	})

  	srv, err := server.NewServer(
  		server.WithServerProtocol(
  			protocol.WithPort(20000),
  			protocol.WithTriple(),
  		),
  	)
  	if err != nil {
  		panic(err)
  	}

  	if err := srv.MountHTTPHandler(mux); err != nil {
  		panic(err)
  	}

  	// Triple services can still be registered as usual before Serve.

  	if err := srv.Serve(); err != nil {
  		panic(err)
  	}
```
## Checklist

  - [x] I confirm the target branch is develop

### DIFF
--- a/protocol/triple/server.go
+++ b/protocol/triple/server.go
@@ -54,13 +54,13 @@ import (
 // Server is TRIPLE adaptation layer representation. It makes use of tri.Server to
 // provide functionality.
 type Server struct {
-	triServer          *tri.Server
-	cfg                *global.TripleConfig
-	mu                 sync.RWMutex
-	services           map[string]grpc.ServiceInfo
-	transportStarted   bool
-	mountedHTTPHandler http.Handler
-	transportSettings  *transportSettings
+	triServer           *tri.Server
+	cfg                 *global.TripleConfig
+	mu                  sync.RWMutex
+	services            map[string]grpc.ServiceInfo
+	transportStarted    bool
+	attachedHTTPHandler http.Handler
+	transportSettings   *transportSettings
 }
 
 type transportSettings struct {
@@ -163,33 +163,38 @@ func (s *Server) ensureTriServer(settings *transportSettings) error {
 	s.mu.Unlock()
 
 	s.triServer = tri.NewServer(settings.location, settings.tripleConfig)
-	// MountHTTPHandler may initialize transport before any Triple service is
-	// exported, so the first tri.Server instance must inherit the root handler.
-	s.mu.RLock()
-	handler := s.mountedHTTPHandler
-	s.mu.RUnlock()
-	if handler != nil {
-		s.triServer.SetFallbackHTTPHandler(handler)
-	}
+	s.wireAttachedHTTPFallback()
 	internal.ReflectionRegister(s)
 	return nil
 }
 
-// MountHTTPHandler attaches a single root HTTP fallback to the Triple server.
-// The mounted handler follows the same single-root semantics as server.Server:
-// once a handler has been attached for one listener, later mounts must fail
-// fast instead of silently replacing the existing fallback.
-func (s *Server) MountHTTPHandler(handler http.Handler) error {
+// wireAttachedHTTPFallback keeps the tri.Server route mux as the main listener
+// handler and connects an attached external HTTP handler through the
+// transport-level fallback slot so Triple routes keep higher priority.
+func (s *Server) wireAttachedHTTPFallback() {
+	s.mu.RLock()
+	handler := s.attachedHTTPHandler
+	s.mu.RUnlock()
+	if handler != nil && s.triServer != nil {
+		s.triServer.SetFallbackHTTPHandler(handler)
+	}
+}
+
+// AttachHTTPHandler attaches a single root HTTP handler to the Triple server.
+// The attached handler follows the same single-root semantics as server.Server:
+// once a handler has been attached for one listener, later attaches must fail
+// fast instead of silently replacing the existing handler.
+func (s *Server) AttachHTTPHandler(handler http.Handler) error {
 	if handler == nil {
-		return fmt.Errorf("mounted HTTP handler must not be nil")
+		return fmt.Errorf("attached HTTP handler must not be nil")
 	}
 
 	s.mu.Lock()
-	if s.mountedHTTPHandler != nil {
+	if s.attachedHTTPHandler != nil {
 		s.mu.Unlock()
-		return fmt.Errorf("an HTTP handler has already been mounted")
+		return fmt.Errorf("an HTTP handler has already been attached")
 	}
-	s.mountedHTTPHandler = handler
+	s.attachedHTTPHandler = handler
 	triServer := s.triServer
 	s.mu.Unlock()
 
--- a/protocol/triple/server.go
+++ b/protocol/triple/server.go
@@ -25,10 +25,21 @@ import (
 	"reflect"
 	"strings"
 	"sync"
+)
 
+import (
 	hessian "github.com/apache/dubbo-go-hessian2"
+
 	"github.com/dubbogo/gost/log/logger"
 
+	grpc_go "github.com/dubbogo/grpc-go"
+
+	"github.com/dustin/go-humanize"
+
+	"google.golang.org/grpc"
+)
+
+import (
 	"dubbo.apache.org/dubbo-go/v3/common"
 	"dubbo.apache.org/dubbo-go/v3/common/constant"
 	"dubbo.apache.org/dubbo-go/v3/global"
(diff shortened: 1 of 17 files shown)


Alanxtl:
cool work

1. add docs in dubbo-website
2. add sample and integrate test in dubbo-go-samples

还需要考虑两个事情

1. 这个http server会不会影响优雅下线
2. 这个http server是否已经兼容目前的probe探针

server/http_mount.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	s.mountedHTTPHandler = handler
+	return nil
+}
+
+// mountHTTPHandlers boots HTTP-capable protocols that support hosting an
+// existing root handler on the same listener as framework-managed routes.
+// Today that means Triple running on an explicit, user-selected port.
+func (s *Server) mountHTTPHandlers() error {
```
Alanxtl, line 61: 这两个函数的名字也太confusing了

protocol/triple/triple.go
```diff
@@ -95,40 +96,42 @@ func (tp *TripleProtocol) Export(invoker base.Invoker) base.Exporter {
 	exporter := NewTripleExporter(serviceKey, invoker, tp.ExporterMap())
 	tp.SetExporterMap(serviceKey, exporter)
 	logger.Infof("[TRIPLE Protocol] Export service: %s", url.String())
-	tp.openServer(invoker, info)
-	internal.HealthSetServingStatusServing(serviceKey)
+	if err := tp.openServer(invoker, info); err != nil {
+		exporter.UnExport()
+		panic(err)
```
Alanxtl, line 101: panic之前log一下

protocol/triple/server.go
```diff
@@ -68,64 +79,168 @@ func NewServer(cfg *global.TripleConfig) *Server {
...
+	// MountHTTPHandler may initialize transport before any Triple service is
+	// exported, so the first tri.Server instance must inherit the root handler.
+	s.mu.RLock()
+	handler := s.mountedHTTPHandler
+	s.mu.RUnlock()
+	if handler != nil {
+		s.triServer.SetFallbackHTTPHandler(handler)
+	}
+	internal.ReflectionRegister(s)
+	return nil
```
Alanxtl, line 175: 这里是不是主handler是s.mountedHTTPHandler
然后fallback handler也是s.mountedHTTPHandler呀

zbchi:
> cool work
> 
> 1. add docs in dubbo-website
> 2. add sample and integrate test in dubbo-go-samples
> 
> 还需要考虑两个事情
> 
> 1. 这个http server会不会影响优雅下线
> 2. 这个http server是否已经兼容目前的probe探针

因为共享了http server，已有的http请求会和triple请求一起被优雅drain，也不会影响现在的probe

protocol/triple/server.go
zbchi: 确实有歧义，主handler一直是tri.server的mux，用户传入的http handler只是被 attach到server上，然后在 transport层通过fallback接进这个mux

protocol/triple/triple.go
```diff
@@ -95,40 +96,43 @@ func (tp *TripleProtocol) Export(invoker base.Invoker) base.Exporter {
...
 	tp.SetExporterMap(serviceKey, exporter)
 	logger.Infof("[TRIPLE Protocol] Export service: %s", url.String())
-	tp.openServer(invoker, info)
-	internal.HealthSetServingStatusServing(serviceKey)
+	if err := tp.openServer(invoker, info); err != nil {
+		logger.Errorf("[TRIPLE Protocol] Export service failed: %s, err: %v", url.String(), err)
+		exporter.UnExport()
+		panic(err)
+	}
+	internal.HealthSetServingStatusServing(url.Service())
```
Alanxtl: `Export` used to mark a Triple service as SERVING with `url.ServiceKey()`, which matches both the exporter map and the shutdown path. This change switches the key to `url.Service()`, but `Check` still looks up the exact requested service string and shutdown still writes NOT_SERVING under `serviceKey`. In practice, health checks for versioned/grouped services can stay `unknown service` even after a successful export, and the SERVING/NOT_SERVING updates no longer target the same entry.

server/server.go
Alanxtl: `Serve` now calls `hostAttachedHTTPHandler()` before `exportServices()`. In the attach-first path that reaches `HostHTTPHandler -> StartHTTPTransport`, the Triple listener is started in a background goroutine immediately; if any later step in `exportServices`, `exportInternalServices`, or service-instance registration fails, `Serve()` returns an error but never tears that listener back down. Before this PR, startup errors happened before any listener was started, so this is a behavioral regression that can leave a partially started server still accepting HTTP traffic after `Serve()` reports failure.

Alanxtl:
pls fix confilcts

Alanxtl (approved):
lgtm

AlexStocks (approved):

PR #3321, opened 2026-05-11
fix(server): replace select{} in Serve() with graceful lifecycle management

### LINKED ISSUE #3042: [提案] server/server.go中Serve优化
### ⚠️ 验证

- [x] 我已经搜索过 [issues](https://github.com/apache/dubbo-go/issues)，确信这不是一个重复的提案。

### ✨ 提案摘要

Image

使用select{}阻塞后无法响应外部信号，关闭不优雅

### 🛠️ 实现方法

使用信号监听或者context等处理

### 📚 附加上下文

_No response_

cvictory:
## Bug
- Symptom: `Server.Serve()` uses `select {}` to block forever, making it impossible to shut down gracefully from code or respond to lifecycle signals.
- Root cause: hard spin (`select {}`) has no termination path.

## Fix
- Add `stopCh chan struct{}` to `Server` for lifecycle signaling.
- `NewServer` initializes `stopCh`.
- `Serve()` blocks on `<-stopCh` instead of `select{}`.
- Add `Stop()` method: safe, idempotent, and allows `Serve()` to be called again after shutdown.
- `Serve()` re-creates `stopCh` if previously closed, supporting restart.
- Add `serve_stop_test.go` with 4 lifecycle tests (all pass).

## Verification
- Regression tests: all 148 existing server tests pass.
- New tests: `TestStopBeforeServe`, `TestStopIdempotent`, `TestServeStopGracefulShutdown`, `TestServeAfterStop` all pass.

Fixes #3042

### DIFF
--- a/server/serve_stop_test.go
+++ b/server/serve_stop_test.go
@@ -1,0 +1,112 @@
+package server
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestStopBeforeServe verifies that Stop() is safe to call before Serve()
+// and does not panic.
+func TestStopBeforeServe(t *testing.T) {
+	srv, err := NewServer()
+	require.NoError(t, err)
+	// Should not panic when server has not been started
+	assert.NotPanics(t, func() {
+		srv.Stop()
+	})
+}
+
+// TestStopIdempotent verifies that calling Stop() multiple times is safe.
+func TestStopIdempotent(t *testing.T) {
+	srv, err := NewServer()
+	require.NoError(t, err)
+
+	// Start Serve in a goroutine; it will block on stopCh.
+	serveErrCh := make(chan error, 1)
+	go func() {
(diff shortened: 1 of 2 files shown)


Alanxtl (commented):
`server/server.go:379-386` only flips `serve=false` and closes `stopCh`. But `Serve()` has already exported services at `server/server.go:355-358` and registered the service instance at `server/server.go:361`; none of those resources are unexported/deregistered/destroyed. Existing teardown lives in paths like `ServiceOptions.Unexport()` (`server/action.go:310`), but `Stop()` never calls it. Result: `Serve()` returns, yet listeners and registry exposure can remain alive, so this is not a graceful shutdown and “Serve can be called again after shutdown” is misleading. The new tests miss this because they use an empty `NewServer()` with no exported service. Please add a real exported Triple/Dubbo service test: call `Serve()`, verify it accepts traffic, call `Stop()`, then verify the port/request is closed or fails and registry/protocol cleanup happened.

Alanxtl:
this issue have been solved in #3293

PR #3333, opened 2026-05-18
fix(registry): make base configurators thread-safe

### LINKED ISSUE #3332: [BUG] BaseConfigurationListener configurators can race with config updates
### What happened

BaseConfigurationListener keeps configurators in a plain slice. Config center callbacks can call Process and replace that slice while protocol or directory override paths call Configurators or OverrideUrl and read or iterate the same slice.

### Impact

Concurrent dynamic config updates and provider or consumer override handling can trigger a data race on the configurator slice. This is a remaining focused race case related to #3247.

### Expected behavior

BaseConfigurationListener should guard configurator updates and expose read-only snapshots to callers, so override paths never iterate a slice while it is being replaced.

### Reproduction idea

Run a race test that concurrently calls Process with update/delete events while other goroutines call Configurators and OverrideUrl.

fallintoplace:
## What changed

- Guard BaseConfigurationListener state with an RWMutex.
- Return copied configurator snapshots from Configurators.
- Make OverrideUrl iterate a snapshot instead of the live slice.
- Add tests for snapshot isolation and concurrent Process / Configurators / OverrideUrl access.

## Why

Config center callbacks can replace BaseConfigurationListener.configurators while provider or consumer override paths read or iterate it. That leaves the shared slice exposed to data races during dynamic config updates.

Fixes #3332

## Tests

- go test ./registry -run 'TestBaseConfigurationListener'
- go test -race ./registry -run 'TestBaseConfigurationListener'
- go test ./registry
- go test -race ./registry
- go test ./...

### DIFF
--- a/registry/base_configuration_listener.go
+++ b/registry/base_configuration_listener.go
@@ -18,6 +18,10 @@
 package registry
 
 import (
+	"sync"
+)
+
+import (
 	"github.com/dubbogo/gost/log/logger"
 
 	perrors "github.com/pkg/errors"
@@ -33,6 +37,7 @@ import (
 
 // BaseConfigurationListener provides shared logic for processing configuration center updates.
 type BaseConfigurationListener struct {
+	lock                    sync.RWMutex
 	configurators           []config_center.Configurator
 	dynamicConfiguration    config_center.DynamicConfiguration
 	defaultConfiguratorFunc func(url *common.URL) config_center.Configurator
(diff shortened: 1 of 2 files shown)


Alanxtl:
@CAICAIIs @nanjiek  check this

CAICAIIs (approved):
LGTM

Alanxtl (approved):
lgtm

AlexStocks (approved):

PR #3364, opened 2026-06-02
fix(metadata): improve error message when protocol is missing

### LINKED ISSUE #3359: [BUG] REST application-level service discovery requires importing Dubbo protocol to fetch metadata
## Title

REST application-level service discovery requires importing Dubbo protocol to fetch metadata

## Problem

A REST consumer using application-level service discovery fails unless it also imports the Dubbo protocol extension:

```go
_ "dubbo.apache.org/dubbo-go/v3/protocol/dubbo"
```

Without this import, the consumer panics:

```text
panic: protocol for [dubbo] is not existing
```

This is surprising because the user-facing service protocol is REST. The Dubbo protocol is only used internally by the service discovery path to call `org.apache.dubbo.metadata.MetadataService`.

## Minimal Reproduction

There is currently no ready-made REST service-discovery sample in the repository, so this was reproduced with a small local REST provider/consumer sample.

### 1. Shared REST mapping

Create a REST service with one method:

- interface: `org.apache.dubbo.samples.rest.GreetingService`
- method: `GetGreeting`
- REST path: `POST /api/v1/users/{userID}/greeting`
- args:
  - arg 0 -> path param `userID`
  - arg 1 -> query param `name`
  - arg 2 -> header `X-Trace-ID`
  - arg 3 -> JSON body

Both provider and consumer install the same REST method config through:

```go
restconfig.SetRestProviderServiceConfigMap(...)
restconfig.SetRestConsumerServiceConfigMap(...)
```

### 2. Provider setup

The provider is configured as a REST service and uses application-level registration:

```go
    dubbo.WithName("dubbo_rest_basic_server"),
    dubbo.WithProtocol(
        protocol.WithREST(),
        protocol.WithIp("127.0.0.1"),
        protocol.WithPort(20080),
    ),
    dubbo.WithRegistry(
        registry.WithZookeeper(),
        registry.WithAddress("127.0.0.1:2181"),
        registry.WithoutUseAsConfigCenter(),
        registry.WithRegisterService(),
    ),
)
```

Then register the REST service:

```go
srv, _ := ins.NewServer()
err = srv.RegisterService(
    &GreetingProvider{},
    dubboserver.WithInterface("org.apache.dubbo.samples.rest.GreetingService"),
)
err = srv.Serve()
```

Required provider imports include REST protocol, registry protocol, service discovery, metadata mapping/report, and ZooKeeper registry.

### 3. Consumer setup that fails

The consumer also uses application-level service discovery:

```go
    dubbo.WithName("dubbo_rest_basic_client"),
    dubbo.WithRegistry(
        registry.WithZookeeper(),
        registry.WithAddress("127.0.0.1:2181"),
        registry.WithoutUseAsConfigCenter(),
        registry.WithRegisterService(),
    ),
)

cli, err := ins.NewClient(client.WithClientNoCheck())
conn, err := cli.Dial(
    "org.apache.dubbo.samples.rest.GreetingService",
    client.WithProtocol(constant.RESTProtocol),
)
```

The consumer imports REST protocol and service discovery related packages, but intentionally does **not** import Dubbo protocol:

```go
_ "dubbo.apache.org/dubbo-go/v3/protocol/rest"
_ "dubbo.apache.org/dubbo-go/v3/registry/protocol"
_ "dubbo.apache.org/dubbo-go/v3/registry/servicediscovery"
_ "dubbo.apache.org/dubbo-go/v3/registry/directory"
_ "dubbo.apache.org/dubbo-go/v3/registry/zookeeper"
_ "dubbo.apache.org/dubbo-go/v3/metadata/mapping/metadata"
_ "dubbo.apache.org/dubbo-go/v3/metadata/report/zookeeper"
```

Run:

```bash
go run ./server
go run ./client
```

The consumer fails with:

```text
panic: protocol for [dubbo] is not existing
```

### 4. Workaround

Adding this import to the REST consumer fixes the issue:

```go
_ "dubbo.apache.org/dubbo-go/v3/protocol/dubbo"
```

After that, the REST call succeeds. The consumer discovers the provider application, fetches metadata through:

```text
dubbo://127.0.0.1:<metadata-port>/org.apache.dubbo.metadata.MetadataService
```

and reconstructs a REST provider URL like:

```text
rest://127.0.0.1:20080/org.apache.dubbo.samples.rest.GreetingService
```

## Expected Behavior

A REST consumer should not need to manually import Dubbo protocol just because it uses application-level service discovery.

If the metadata service internally uses Dubbo protocol, that dependency should be hidden by the service discovery / metadata module, or the error should clearly explain the required import.

## Actual Behavior

The REST consumer fails at runtime because service discovery tries to call the metadata service through `dubbo://...`, but the Dubbo protocol extension is not registered.

## Why This Matters

This leaks an internal metadata transport detail into user code. From the user perspective, the configured service protocol is REST, the registry is ZooKeeper/Nacos, and the registry type is application-level service discovery. Requiring a Dubbo protocol import is non-obvious and easy to miss.

## Suggested Fix

One of the following would improve the behavior:

1. service discovery automatically registers the protocol needed for metadata service calls;
2. provide a documented import bundle for application-level service discovery dependencies;
3. return a clearer error such as: `metadata service uses dubbo protocol, please import dubbo.apache.org/dubbo-go/v3/protocol/dubbo`.

wlgusqkr:
### Description
Fixes #3359  (issue)

Application-level service discovery fetches metadata through the metadata service URL. When that metadata service protocol is not registered, return an actionable error instead of panicking through MustGet.

For dubbo metadata service URLs, the error now explains that the consumer needs to import dubbo.apache.org/dubbo-go/v3/protocol/dubbo.

Add a metadata client test to verify the missing protocol path is handled without panic.

### Checklist

### DIFF
--- a/common/extension/protocol.go
+++ b/common/extension/protocol.go
@@ -18,10 +18,6 @@
 package extension
 
 import (
-	"fmt"
-)
-
-import (
 	"dubbo.apache.org/dubbo-go/v3/protocol/base"
 )
 
@@ -35,15 +31,6 @@ func SetProtocol(name string, v func() base.Protocol) {
 // GetProtocol finds the protocol extension with @name
 func GetProtocol(name string) base.Protocol {
 	return protocols.MustGet(name)()
-}
-
-// GetProtocolWithError finds the Protocol extension with @name, returns error if not found
-func GetProtocolWithError(name string) (base.Protocol, error) {
-	creator, ok := protocols.Get(name)
-	if !ok {
-		return nil, fmt.Errorf("protocol %s is not found", name)
-	}
-	return creator(), nil
 }
 
 // UnregisterProtocol removes the protocol extension with @name
--- a/common/extension/protocol.go
+++ b/common/extension/protocol.go
@@ -18,6 +18,10 @@
 package extension
 
 import (
+	"fmt"
+)
+
+import (
 	"dubbo.apache.org/dubbo-go/v3/protocol/base"
 )
 
@@ -31,6 +35,15 @@ func SetProtocol(name string, v func() base.Protocol) {
 // GetProtocol finds the protocol extension with @name
 func GetProtocol(name string) base.Protocol {
 	return protocols.MustGet(name)()
+}
+
+// GetProtocolWithError finds the Protocol extension with @name, returns error if not found
+func GetProtocolWithError(name string) (base.Protocol, error) {
+	creator, ok := protocols.Get(name)
+	if !ok {
+		return nil, fmt.Errorf("protocol %s is not found", name)
+	}
+	return creator(), nil
 }
 
 // UnregisterProtocol removes the protocol extension with @name
--- a/filter/README.md
+++ b/filter/README.md
@@ -2,7 +2,7 @@
 
 ## Getting Started
 
-Recommended Way: import what you needs, see also [dubbo-go/imports](https://github.com/dubbogo/imports).
+Recommended Way: import only the filters you need. See also [dubbo-go/imports](https://github.com/dubbogo/imports).
 
 ```go
 package demo
@@ -24,13 +24,16 @@ import _ "dubbo.apache.org/dubbo-go/v3/filter/filter_impl"
 
 - accesslog: Access Log Filter(https://github.com/apache/dubbo-go/pull/214)
 - active
+- adaptivesvc: Adaptive Service Filter
 - auth: Auth/Sign Filter(https://github.com/apache/dubbo-go/pull/323)
 - echo: Echo Health Check Filter
-- execlmt: Execute Limit Filter(https://github.com/apache/dubbo-go/pull/246)
+- exec_limit: Execute Limit Filter(https://github.com/apache/dubbo-go/pull/246)
 - generic: Generic Filter(https://github.com/apache/dubbo-go/pull/291)
-- gshutdown: Graceful Shutdown Filter
+- graceful_shutdown: Graceful Shutdown Filter
 - hystrix: Moved to [dubbo-go-extensions](https://github.com/apache/dubbo-go-extensions/tree/main/filter/hystrix)
 - metrics: Metrics Filter(https://github.com/apache/dubbo-go/pull/342)
+- otel/trace: OpenTelemetry Tracing Filter
+- polaris/limit: Polaris Rate Limit Filter
 - seata: Seata Filter
 - sentinel: Sentinel Filter
 - token: Token Filter(https://github.com/apache/dubbo-go/pull/202)
--- a/metadata/client.go
+++ b/metadata/client.go
@@ -53,15 +53,7 @@ func GetMetadataFromMetadataReport(revision string, instance registry.ServiceIns
 func GetMetadataFromRpc(revision string, instance registry.ServiceInstance) (*info.MetadataInfo, error) {
 	url := buildStandardMetadataServiceURL(instance)
 	url.SetParam(constant.TimeoutKey, defaultTimeout)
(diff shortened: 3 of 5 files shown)


Alanxtl:
our project use `import-formatter` to format import blocks, that's the reason why ur CI fails. For you, u should 

1. run `go install github.com/dubbogo/tools/cmd/imports-formatter@latest`
2. cd to the root dir of `dubbo-go`
3. run `imports-formatter`

Alanxtl (changes requested):
This improves the error message but does not fix the underlying REST service discovery issue. A REST consumer still needs to import dubbo.apache.org/dubbo-go/v3/protocol/dubbo because metadata fetching still depends on dubbo://... internally.

Alanxtl:
Thanks for working on this, but I think we should not fix this by improving the error message.

The issue is not that the error is unclear. The issue is that REST application-level service discovery leaks an internal metadata transport dependency to user code.

For a REST consumer, users configure and import REST-related components:

```go
_ "dubbo.apache.org/dubbo-go/v3/protocol/rest"
_ "dubbo.apache.org/dubbo-go/v3/registry/servicediscovery"
```

But application-level discovery internally fetches metadata through:

```text
dubbo://.../org.apache.dubbo.metadata.MetadataService
```

So today the consumer also has to manually import:

```go
_ "dubbo.apache.org/dubbo-go/v3/protocol/dubbo"
```

That should not be required. The Dubbo protocol here is an internal metadata-service transport detail.

I suggest changing the fix direction:

1. Revert the new `GetProtocolWithError` / metadata-specific import hint change.
2. Make application-level service discovery register its internal metadata transport dependency automatically.
3. Add a test proving that a REST consumer using service discovery works without importing `protocol/dubbo`.

A minimal fix may be to add a side-effect import in the service discovery package:

```go
import (
    _ "dubbo.apache.org/dubbo-go/v3/protocol/dubbo"
)
```

Probably `registry/servicediscovery` is a reasonable place, because that is the feature path that calls `metadata.GetMetadataFromRpc(...)` and requires the default Dubbo metadata protocol.

The expected test should cover:

- provider protocol: REST;
- registry type: application-level service discovery;
- consumer imports REST protocol and service discovery packages;
- consumer does **not** import `dubbo.apache.org/dubbo-go/v3/protocol/dubbo`;
- metadata is still fetched and the REST provider URL is reconstructed successfully.

So I would prefer fixing the dependency registration instead of returning a clearer error that still asks REST users to import Dubbo manually.

wlgusqkr:
@Alanxtl Thanks for the guidance! I've updated the PR to follow this direction.

I reverted the `GetProtocolWithError` / metadata import-hint change and added a side-effect import of `protocol/dubbo` in `registry/servicediscovery`, so application-level service discovery registers its internal metadata transport dependency automatically.

I kept the code change minimal and did not add a test in this revision. I considered adding an isolated import test under something like `registry/servicediscovery/importtest`, because an in-package test under `registry/servicediscovery` can be affected by existing tests that call `extension.SetProtocol("dubbo", ...)` with a mock protocol.

Do you think adding that isolated import test is worthwhile here, or would you prefer keeping this as the minimal side-effect import change?

Alanxtl:
Sorry, I took another look and realized this is not actually a bug in the way I originally described.

The metadata service protocol is recorded in the service instance metadata, and the registry itself does not need to understand or be compatible with the metadata RPC protocol. The consumer just needs the corresponding protocol capability when it calls the provider metadata service.

So this is more about sample/import setup than a framework bug. I’ll close this issue and the related PR. Thanks for looking into it and sorry for the confusion.

PR #3377, opened 2026-06-08
docs(config-center): explain optional governance keys

wyf027:
### Description

Refs #2151.

This PR documents the config-center behavior that caused confusion in the issue:

- enabling a config center subscribes to application-level and service-level governance rule keys
- those keys use the application or service key plus the `configurators` suffix
- missing governance rule keys are optional and should not be interpreted as application startup failure

### Checklist

- [x] I confirm the target branch is `main`
- [x] Code has passed local testing (`gofmt`, `git diff --check`)
- [x] Tests are not required because this is documentation-only

### DIFF
--- a/config_center/doc.go
+++ b/config_center/doc.go
@@ -15,5 +15,18 @@
  * limitations under the License.
  */
 
-// Package config_center provides Config Center definition and implementations for listening service governance rules.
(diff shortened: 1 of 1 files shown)


Alanxtl:
not planned

PR #3379, opened 2026-06-08
test(triple): allow invalid stream send close errors

### LINKED ISSUE #3142: [BUG] TestStreamForServer fails in CI with QUIC port announcement error (not reproducible locally)

wyf027:
## What changed

- Reworked the fix around the failing stream test expectation instead of changing Alt-Svc/HTTP3 negotiation behavior.
- In `TestStreamForServer/client-stream-conn`, an invalid server-side `stream.Conn().Send("not-proto")` still has to fail on the server side.
- The client-side `CloseAndReceive` path now accepts the observed `CodeUnknown` write-envelope error when the stream is already closed by that invalid send.

## Why

The original PR tried to suppress unrelated QUIC Alt-Svc header advertisement for `:0`. Maintainer feedback pointed out that #3142 is about the invalid stream send path, not HTTP3 negotiation. This update keeps framework behavior unchanged and makes the test assert the actual invalid-send behavior explicitly.

Refs #3142

## Validation

- `go test ./protocol/triple/triple_protocol -run 'TestStreamForServer/client-stream-conn|TestAltSvcHandlerNegotiation|TestNewAltSvcHandler'`
- `git diff --check`

### DIFF
--- a/protocol/triple/triple_protocol/triple_ext_test.go
+++ b/protocol/triple/triple_protocol/triple_ext_test.go
@@ -1681,7 +1681,10 @@ func TestStreamForServer(t *testing.T) {
 		assert.Nil(t, stream.Send(&pingv1.SumRequest{Number: 1}))
 		res := triple.NewResponse(&pingv1.SumResponse{})
 		err = stream.CloseAndReceive(res)
-		assert.Nil(t, err)
+		if err != nil {
+			assert.Equal(t, triple.CodeUnknown, triple.CodeOf(err))
+			assert.True(t, strings.Contains(err.Error(), "write envelope"))
+		}
 	})
 	t.Run("client-stream-send-msg", func(t *testing.T) {
 		t.Parallel()


Alanxtl:
Thanks for the patch, but I don't think this is the right direction.

The failure in #3142 is about `TestStreamForServer/client-stream-conn` receiving `unknown: write envelope: EOF` after the server intentionally calls `stream.Conn().Send("not-proto")`. That is a test scenario built around an invalid server-side send, and the current PR only suppresses an unrelated Alt-Svc warning for `:0`.

I don't think we should change framework behavior just to hide this warning. The fix should first revisit the test expectation: after an invalid `stream.Conn().Send("not-proto")`, it may not be valid to require the client-side `CloseAndReceive` path to complete with nil in all environments. If the goal is to verify invalid message handling, the test should assert the actual error behavior more explicitly instead of depending on this stream continuing cleanly.

So I think this PR should be reworked around the failing test / stream expectation, not around skipping Alt-Svc header advertisement.

Alanxtl:
please update **all of your prs**, your base branch should be develop and your target branch should be develop

wyf027:
Updated this PR based on the review. It no longer changes Alt-Svc/HTTP3 negotiation. The diff is now limited to the failing stream test: after the server intentionally sends a non-proto message and that send fails, the client CloseAndReceive path may return the observed CodeUnknown write-envelope error instead of always completing nil. Local validation: go test ./protocol/triple/triple_protocol -run 'TestStreamForServer/client-stream-conn|TestAltSvcHandlerNegotiation|TestNewAltSvcHandler' and git diff --check.

wyf027:
I checked the failed Integration Test run. The failure is in the integration sample `generic`, where the Go client cannot connect to `127.0.0.1:50052` after the sample server startup (`connect: connection refused`, followed by `Some generic call tests failed`). This PR now only changes `protocol/triple/triple_protocol/triple_ext_test.go` and does not touch the generic sample, integration scripts, or runtime code. I tried to rerun the failed job, but GitHub requires repository admin permission for reruns from my account. Could a maintainer rerun the failed Integration Test job?

AlexStocks (approved):

Issue #2582, opened 2024-01-16
老dubbo-go接入项目升级dubbo-go v3.x，发现GenericService2 的定义被移除，invoke函数调用无法获取“响应attchment”上下文结果


drama10096:
### Environment

- Server: 
- Client: 
- Protocol: 
- Registry: 

### Issue description

项目原本使用“zookeeper + dubbo-go 1.5.8”，像是遇到内存泄露问题，准备升级dubbo-go v3.x尝试解决，但是dubbo-go v3.x里没有 		
GenericService2 的定义，invoke函数调用无法获取“响应attchment”上下文

### Logs

<details><summary>Click me to check logs</summary>
![Snipaste_2024-01-16_15-46-23](https://github.com/apache/dubbo-go/assets/37733059/627fcbd9-0f24-48d1-889b-1d723415e794)

![Snipaste_2024-01-16_15-41-16](https://github.com/apache/dubbo-go/assets/37733059/31644de5-6a11-4683-9259-5ab7794162a7)

Copy logs to here.
```

</details>

drama10096:
老的dubbo-go 1.5.8就可以获取的

![Snipaste_2024-01-16_15-46-23](https://github.com/apache/dubbo-go/assets/37733059/f4017686-542e-4bfe-a43c-13948b5f16df)

No-SilverBullet:
@1kasa

Alanxtl:
refer to https://github.com/apache/dubbo-go-samples/tree/main/context

PR #3381, opened 2026-06-08
docs(generic): explain response attachment access

wyf027:
## What changed

- Documented how dubbo-go 3.x callers can read response attachments after a Triple generic invocation.
- Added a migration-oriented example for applications that previously relied on `GenericService2` in dubbo-go 1.5.x.
- Clarified that provider-side result attachments are propagated through Triple response trailers for unary calls.

## Why

Issue #2582 asks how to get response attachments after upgrading from the old `GenericService2` API. The current Triple path already propagates unary result attachments as response trailers, but the replacement usage pattern was not documented in this repository.

Refs #2582

## Validation

- `git diff --check`


Alanxtl (approved):
lgtm

Alanxtl:
can you help duplicate this doc to https://github.com/apache/dubbo-website/blob/master/content/en/overview/mannual/golang-sdk/tutorial/rpc/filter.md and https://github.com/apache/dubbo-go-samples/tree/main/generic ?

Issue #2410, opened 2023-09-01
Dubbo-go作为Provider时, RPC返回的结构体中如果包含指针字段, Java端无法正确解析


yukinomiu:
### Environment

- Server: Dubbo-go, v3.10
- Client: dubbo-spring-boot-starter, v3.2.5
- Protocol: dubbo
- Registry: Nacos, v2.2.2

### Issue description

Dubbo-go作为Provider, RPC接口返回结构体指针. 这个结构体内包含一个指针类型的字段.
Java Dubbo作为Consumer, 调用接口, 无法解析结果.

Provider端代码:

```
type TestService interface {
	Test(ctx context.Context, req *Req) (*Resp, error)
}

type Resp struct {
	Total *int64
}

```

Consumer端代码:

```
public interface TestService {
    Resp test(Req request);
}

public class Resp implements Serializable {

    private Long total;

    public Long getTotal() {
        return total;
    }

    public void setTotal(Long total) {
        this.total = total;
    }
}
```

### Logs

<details><summary>Click me to check logs</summary>

```
2023-09-01 14:49:26.344  WARN 44832 --- [nio-8090-exec-6] o.a.d.r.p.dubbo.DecodeableRpcResult      :  [DUBBO] Decode rpc result failed: com.github.yukinomiu.dubbo.api.Resp.total: com.alibaba.com.caucho.hessian.io.BasicDeserializer: unexpected object java.lang.String, dubbo version: 3.2.5, current host: 172.16.116.11, error code: 4-20. This may be caused by , go to https://dubbo.apache.org/faq/4/20 to find instructions. 

com.alibaba.com.caucho.hessian.io.HessianFieldException: com.github.yukinomiu.dubbo.api.Resp.total: com.alibaba.com.caucho.hessian.io.BasicDeserializer: unexpected object java.lang.String
	at com.alibaba.com.caucho.hessian.io.JavaDeserializer.logDeserializeError(JavaDeserializer.java:168) ~[hessian-lite-3.2.13.jar:na]
	at com.alibaba.com.caucho.hessian.io.JavaDeserializer$ObjectFieldDeserializer.deserialize(JavaDeserializer.java:415) ~[hessian-lite-3.2.13.jar:na]
	at com.alibaba.com.caucho.hessian.io.JavaDeserializer.readObject(JavaDeserializer.java:277) ~[hessian-lite-3.2.13.jar:na]
[...]

PR #3382, opened 2026-06-08
test(hessian2): cover pointer long response fields

wyf027:
## What changed

- Added a Dubbo Hessian response regression test for a POJO containing a `*int64` field.
- The test encodes a provider response, decodes it back, and verifies the pointer field remains a non-nil `*int64` with the expected value.

## Why

Issue #2410 reported Java consumers failing to deserialize a Go provider response when the returned struct contains a pointer `int64` field mapped to Java `Long`. The current dependency already uses `dubbo-go-hessian2 v1.12.5`, which includes the upstream fix mentioned in the issue thread. This test keeps that compatibility covered in dubbo-go.

Refs #2410

## Validation

- `go test ./protocol/dubbo/hessian2 -run 'TestResponseWithPointerLongField|TestResponse'`
- `git diff --check`

### DIFF
--- a/protocol/dubbo/hessian2/hessian_dubbo_test.go
+++ b/protocol/dubbo/hessian2/hessian_dubbo_test.go
@@ -48,8 +48,16 @@ type CaseB struct {
 	B CaseA
 }
 
+type PointerLongResp struct {
+	Total *int64
+}
+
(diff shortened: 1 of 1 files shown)


Alanxtl (approved):
lgtm

Issue #2084, opened 2022-10-18
panic when define a 'Echo'  service


weiyuanke:
服务端proto文件
![image](https://user-images.githubusercontent.com/1445434/196411721-3c742955-f86e-4e38-b2ef-aa98ef649b62.png)

客户端测试代码
![image](https://user-images.githubusercontent.com/1445434/196412004-3412c01b-6249-4a77-bf3b-bfa0d59f2f0e.png)

报错信息：
![image](https://user-images.githubusercontent.com/1445434/196412638-59e2369a-10c9-4ace-8c7d-2111ecf36466.png)

`2022-10-17T20:02:15.705+0800  INFO  config/root_config.go:150  [Config Center] Config center doesn't start
2022-10-17T20:02:15.706+0800  INFO  [dub](https://github.com/apache/dubbo-go/issues/new?assignees=&labels=kind%2Fbug&template=bug-report.md#)[bo](https://github.com/apache/dubbo-go/issues/new?assignees=&labels=kind%2Fbug&template=bug-report.md#)3/dubbo3_protocol.go:141  [Triple Protocol] Refer service: tri://localhost:20000/api.EchoService?app.version=3.0.0&application=dubbo.io&async=false&bean.name=EchoServiceClientImpl&cluster=failover&config.tracing=&environment=dev&generic=&group=&interface=api.EchoService&loadbalance=&metadata-type=local&module=sample&name=dubbo.io&organization=dubbo-go&owner=dubbo-go&provided-by=&reference.filter=cshutdown&registry.role=0&release=dubbo-golang-3.0.0&retries=&serialization=&side=consumer&sticky=false&timestamp=1666008135&version=
2022-10-17T20:02:15.707+0800  INFO  dubbo3/dubbo3_protocol.go:141  [Triple Protocol] Refer service: tri://localhost:20000/api.Greeter?app.version=3.0.0&application=dubbo.io&async=false&bean.name=GreeterClientImpl&cluster=failover&config.tracing=&environment=dev&generic=&group=&interface=api.Greeter&loadbalance=&metadata-type=local&module=sample&name=dubbo.io&organization=dubbo-go&owner=dubbo-go&provided-by=&reference.filter=cshutdown&registry.role=0&release=dubbo-golang-3.0.0&retries=&serialization=&side=consumer&sticky=false&timestamp=1666008135&version=
name:"Hello laurence" id:"12345" age:21
panic: reflect: call of reflect.Value.IsZero on zero Value

[...]

PR #3383, opened 2026-06-08
docs(echo): note reserved Echo method behavior

wyf027:
## What changed

- Expanded the echo filter package comment to explain that `Echo` is a reserved built-in health check method name.
- Documented that services should avoid business methods named `Echo` when the echo filter is enabled.
- Kept runtime behavior unchanged.

## Why

Issue #2084 reports a panic after defining a service named/methoded `Echo`; maintainers clarified that this is caused by conflict with dubbo-go's built-in echo filter behavior and is not a runtime bug. The package comment now makes that behavior explicit.

Refs #2084

## Validation

- `gofmt -w filter/echo/filter.go`
- `git diff --check`

Note: `go test ./filter/echo` was not completed in this sparse checkout because `common/extension` pulls additional packages that are outside the current sparse tree.

### DIFF
--- a/filter/echo/filter.go
+++ b/filter/echo/filter.go
@@ -15,9 +15,18 @@
  * limitations under the License.
  */
 
-// Package echo providers health check filter.
-// RPCService need a Echo method in consumer, if you want to use Filter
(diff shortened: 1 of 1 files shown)


Alanxtl:
not planned

PR #3384, opened 2026-06-08
docs(filter): describe Java overload dispatch

wyf027:
## What changed

- Added README guidance for implementing Java-style overloaded methods in dubbo-go.
- Documented the single Go entrypoint plus `MethodMapper` pattern.
- Included an argument-count dispatch example for `isForbidsSpeak`.

## Why

Issue #1780 asks how to handle Java overloaded methods from Go. Maintainer discussion points to dispatching by argument count or type instead of relying on Go method overloading, which the language does not support. This documents that pattern in the repo.

Refs #1780

## Validation

- `git diff --check`


Alanxtl:
not planned

Issue #257, opened 2019-11-20
Is must that the first alpha of the method name in java interface is upper case to compatible with golang?


fionawp:
**What would you like to be added**:

If apache/dubbo-go project can compatible with golang by comparing to the method name in client.yml. If the first alpha of the method name in the config is not Upper case, transfer it to lower case when searching the java method name.

**Why is this needed**:

I am trying to call java dubbo services by using apache/dubbo-go. Due to historical reasons, it is not possible to change java codes because they are called by other many projects.

fangyincheng:
We have supported this feature, you can reference to [https://github.com/dubbogo/dubbogo-samples/blob/master/general/dubbo/go-client/app/user.go#L127](url).
![image](https://user-images.githubusercontent.com/18375310/69251762-de6b9000-0bec-11ea-985c-045bea22a160.png)
`dubbo:"getUser"`  means calling `GetUser` is equivalent to calling `getUser` in java.
Welcome to give dubbo-go more suggestions!

fangyincheng:
@fionawp Hi,  has your problem been solved?

AlexStocks:
@fangyincheng we can keep this issue open for other beginners.

zouyx:
> @fangyincheng we can keep this issue open for other beginners.

If no another problems, i think should close this  issue.

Alanxtl:
close via https://github.com/apache/dubbo-go/pull/3277

AlexStocks:
> close via [#3277](https://github.com/apache/dubbo-go/pull/3277)

"good first issue" is a good use case. please do not close it.

PR #3385, opened 2026-06-08
docs: explain Java method name mapping

wyf027:
## What changed

- Added README guidance for Java method name compatibility.
- Documented first-letter case fallback for exported Go RPC methods.
- Added a `MethodMapper` example for explicit lower camel case Java method mapping.

## Why

Issue #257 is kept open as a beginner-facing interoperability reference. The behavior is already implemented, including the routing improvements from #3277, but the README did not clearly explain how Go `GetUser` maps to Java `getUser` or when to use `MethodMapper`.

Refs #257

## Validation

- `git diff --check`


Alanxtl:
we are not planning to add this doc in readme, if you really think this update is necessary, please add this doc to apache/dubbo-website instead of here

PR #3389, opened 2026-06-09
ci: add race test for core packages

wyf027:
### Description

Refs #3248.

This PR adds a focused race-detector CI path for core packages that guard shared framework state:

- adds `make test-race-core`
- runs `go test -race` for `common/...`, `cluster/router/chain`, `protocol/base`, and `registry/directory`
- wires the target into a separate GitHub Actions `Race Test` job so race failures are reported independently from the existing unit-test/coverage job

### Validation

- `make test-race-core`
- YAML parse check for `.github/workflows/github-actions.yml`
- `git diff --check`

### Checklist


Alanxtl:
not planned

PR #3402, opened 2026-06-12
perf: optimize IsEquals and ToMap to reduce allocations

### LINKED ISSUE #3394: perf(common): avoid unnecessary allocations in URL IsEquals and ToMap
### Background

`IsEquals()` currently converts both URLs to maps via `ToMap()`, deletes excluded keys, and then compares the maps. This creates avoidable allocations for a comparison-only path. `ToMap()` itself can also be tightened by pre-sizing the map and avoiding repeated string splits.

Related to #3248.

### Scope

Optimize equality and map conversion paths:

- Rework `IsEquals()` to compare fields and params directly without materializing two full maps when possible.
- Treat excluded params through a small lookup set instead of deleting from copied maps.
- Keep `ToMap()` public behavior unchanged, but reduce allocation work where possible.
- Avoid repeated parsing of `Location`; use a single-pass split strategy while preserving current host/port behavior.

### Acceptance criteria

- Existing equality and `ToMap()` tests continue to pass.
- Add or update benchmarks for `IsEquals()` and `ToMap()` with small and larger param sets.
- The PR shows `-benchmem` output demonstrating lower allocations or no regression.
- `go test ./common` passes.

JunLa-0:
Optimize URL equality comparison and map conversion to avoid unnecessary allocations. IsEquals() now compares fields directly instead of materializing two full maps via ToMap(). ToMap() pre-sizes the map and uses a single-pass split for Location parsing.

Fixes: #3394

Signed-off-by: 13927699266@163.com

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -820,7 +820,32 @@ func (c *URL) SetParams(m url.Values) {
 
 // ToMap transfer URL to Map
 func (c *URL) ToMap() map[string]string {
-	paramsMap := make(map[string]string)
+	c.paramsLock.RLock()
+	paramCount := len(c.params)
+	c.paramsLock.RUnlock()
+
+	fixedCount := 0
+	if c.Protocol != "" {
+		fixedCount++
+	}
+	if c.Username != "" {
+		fixedCount++
+	}
+	if c.Password != "" {
+		fixedCount++
+	}
+	if c.Location != "" {
+		fixedCount += 2 // host + port
+	}
+	if c.Path != "" {
+		fixedCount++
+	}
+
+	if paramCount == 0 && fixedCount == 0 {
+		return nil
+	}
+
+	paramsMap := make(map[string]string, paramCount+fixedCount)
 
 	c.RangeParams(
 		func(key, value string) bool {
@@ -839,23 +864,17 @@ func (c *URL) ToMap() map[string]string {
 		paramsMap["password"] = c.Password
 	}
 	if c.Location != "" {
-		paramsMap["host"] = strings.Split(c.Location, ":")[0]
-		var port string
-		if strings.Contains(c.Location, ":") {
-			port = strings.Split(c.Location, ":")[1]
+		// Single-pass split avoids repeated string scanning.
+		parts := strings.Split(c.Location, ":")
+		paramsMap["host"] = parts[0]
+		if len(parts) > 1 {
+			paramsMap["port"] = parts[len(parts)-1]
 		} else {
-			port = "0"
+			paramsMap["port"] = "0"
 		}
-		paramsMap["port"] = port
-	}
-	if c.Protocol != "" {
-		paramsMap[PROTOCOL] = c.Protocol
 	}
 	if c.Path != "" {
 		paramsMap["path"] = c.Path
-	}
-	if len(paramsMap) == 0 {
-		return nil
 	}
 	return paramsMap
 }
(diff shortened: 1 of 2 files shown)


Alanxtl:
the target branch should be develop

Alanxtl:
@ywxzm03 help review this :)

Alanxtl:
update to latest develop and run https://github.com/apache/dubbo-go/blob/develop/common/url_benchmark_test.go to see the performance shift before and after you commit, and paste the result comparison here

可以参考 https://github.com/apache/dubbo-go/pull/3403 的pr body写法

common/url.go
```diff
@@ -1019,23 +1038,76 @@ func IsEquals(left *URL, right *URL, excludes ...string) bool {
...
+	if left.Location != right.Location {
+		if isExcluded("host") && isExcluded("port") {
+			// Both excluded, skip Location comparison
+		} else if isExcluded("host") || isExcluded("port") {
+			// Only one side of host:port excluded — fall through to param comparison
+		} else {
 			return false
-		} else if lv != rv {
+		}
+	}
```
Alanxtl, line 1080: `IsEquals` 在只 exclude `host` 或 `port` 时会漏比较另一半 location

   PR 里当 `left.Location != right.Location` 且只排除 `host` 或 `port` 其中一个时，代码直接“fall through”，但后面只比较 params，不再比较未被排除的 host/port。  
   
你试一下`Location: 10.0.0.1:20880` vs `10.0.0.2:20880`，exclude `port` 后 PR 返回 `true`，base 返回 `false`。

common/url.go
```diff
@@ -820,7 +820,32 @@ func (c *URL) SetParams(m url.Values) {
...
+	if c.Path != "" {
+		fixedCount++
+	}
+
+	if paramCount == 0 && fixedCount == 0 {
+		return nil
+	}
+
+	paramsMap := make(map[string]string, paramCount+fixedCount)
```
Alanxtl, line 848: - `paramCount` 单独加锁读一次，后面 `RangeParams` 又加锁遍历一次，多了一次锁操作。
- 这不是一致性快照：两次锁之间 params 可能变，容量估算可能不准。
- `fixedCount` 只是容量估算，不影响逻辑；估多了也只是多分一点内存。
- 如果 params 里本来就有 `protocol`、`host`、`port` 这些 key，后面的固定字段会覆盖它们，`fixedCount` 仍然按新增 key 算，容量也会估大。
- 性能收益主要来自预分配，但可读性代价有点高。

建议就是保留容量估算，去掉无意义初始化，并确保 `protocol/location` 行为不变。

common/url.go
```diff
@@ -839,24 +864,18 @@ func (c *URL) ToMap() map[string]string {
...
-			port = strings.Split(c.Location, ":")[1]
+		// Single-pass split avoids repeated string scanning.
+		parts := strings.Split(c.Location, ":")
+		paramsMap["host"] = parts[0]
+		if len(parts) > 1 {
+			paramsMap["port"] = parts[len(parts)-1]
 		} else {
-			port = "0"
+			paramsMap["port"] = "0"
 		}
```
Alanxtl, line 874: `ToMap` 的多地址 port 解析语义被改了

   原逻辑对 `127.0.0.1:2181,127.0.0.1:2182,127.0.0.1:2183` 的 `port` 是 `2181,127.0.0.1` 确实不合理，
   PR 改成取最后一个 `:` 后的 `2183`。也不合理

  这里改成合理的逻辑，然后在pr简介里面说明

common/url_test.go
```diff
@@ -1507,3 +1507,82 @@ func TestCloneThreadSafe(t *testing.T) {
...
+	u := &URL{
+		Protocol: "dubbo",
+		Location: "127.0.0.1:20000",
+		Path:     "/com.test.Service",
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_ = u.ToMap()
+	}
+}
```
Alanxtl, line 1588: 不用自己写benchmark 直接复用已有的benchmark 或者在benchmark文件里面添加你需要的benchmark

common/url.go
```diff
@@ -1019,23 +1038,76 @@ func IsEquals(left *URL, right *URL, excludes ...string) bool {
...
 
-	if len(leftMap) != len(rightMap) {
-		return false
+	isExcluded := func(key string) bool {
+		if excludeSet == nil {
+			return false
+		}
+		_, ok := excludeSet[key]
+		return ok
 	}
```
Alanxtl, line 1058: This rewrite seems more complicated than necessary, and it also changes the original comparison semantics.

The old implementation compares the result of ToMap(), so excluding only "port" should still compare "host", and excluding only "host" should still compare "port". In the new logic, when left.Location != right.Location and only one of host/port is excluded, the code falls through to param comparison, but it never compares the remaining non-excluded part of Location.

For example, these should not be equal when only "port" is excluded:

left.Location = "10.0.0.1:20880"
right.Location = "10.0.0.2:20880"
IsEquals(left, right, "port")

Besides that, the code still materializes leftParams, so the optimization is not as simple as "avoid materializing maps"; it only avoids building two full ToMap() results. I suggest either keeping the original map-based comparison for correctness, or splitting host/port explicitly and comparing only the non-excluded fields.

common/url.go
```diff
@@ -1019,23 +1038,76 @@ func IsEquals(left *URL, right *URL, excludes ...string) bool {
...
+			// Only one side of host:port excluded — fall through to param comparison
+		} else {
 			return false
-		} else if lv != rv {
+		}
+	}
+
+	// Compare params directly without materializing maps.
+	// Build left params (excluding filtered keys), then verify right matches.
+	leftParams := make(map[string]string)
```
ywxzm03, line 1084: 这里可以做个容量预估，len(left.params) 就有了

common/url.go
```diff
@@ -1019,23 +1038,76 @@ func IsEquals(left *URL, right *URL, excludes ...string) bool {
...
+	if left.Username != right.Username && !isExcluded("username") {
+		return false
+	}
+	if left.Password != right.Password && !isExcluded("password") {
+		return false
+	}
+	if left.Path != right.Path && !isExcluded("path") {
+		return false
+	}
+	if left.Location != right.Location {
```
AlexStocks, line 1072: [P1] 这里把 `host` / `port` 的排除逻辑挂在整条 `Location` 上了：只传 `host` 或只传 `port` 时，`left.Location != right.Location` 也会直接落入“放行”分支，另一半地址差异被一并忽略。旧实现是按 `ToMap()` 删除单个 key 比较，两个字段的排除语义是独立的。请分别比较 host/port，或保留原来的按键删除语义。

Alanxtl:
@JunLa-0 any progess?

Alanxtl:
@JunLa-0  any progess? If no further update, this pr will be closed and we will assign others to work on this.

Alanxtl:
close via #3488

PR #3399, opened 2026-06-11
Feat: add benchmarks for common.URL

ywxzm03:
**What this PR does**:

Add focused benchmarks for `common.URL` to establish a quantifiable baseline (throughput + memory allocation) for future optimization, so changes can be evaluated with data rather than intuition.

This PR only adds `common/url_benchmark_test.go` and does not change any production code.

Methods covered:

- `URL.String()`
- `URL.Key()`
- `URL.GetCacheInvokerMapKey()`
- `URL.ServiceKey()`
- `URL.Clone()` / `URL.CloneWithFilter()`
- `URL.MergeURL()`
- `URL.CopyParams()` / `URL.GetParams()`
- `URL.IsEquals()` / `URL.ToMap()`

Test dimensions:

- Parameter sizes use four tiers: `1 / 32 / 256 / 1024`.
- `Clone` / `CloneWithFilter` include cases with `SubURL` to surface deep-copy costs.
- All methods with return values write their results to package-level sink variables to prevent the compiler from optimizing the calls away and skewing the data.

**Which issue(s) this PR fixes**:

Fixes #3391

**Special notes for your reviewer**:

Benchmark command:

```bash
go test ./common -bench 'BenchmarkURL' -benchmem -run '^$'
```

Output:

```text
goos: darwin
goarch: arm64
pkg: dubbo.apache.org/dubbo-go/v3/common
cpu: Apple M5
BenchmarkURLString/params_1-10                   5456774               200.9 ns/op           384 B/op          9 allocs/op
BenchmarkURLString/params_32-10                   843082              1515 ns/op            2297 B/op         15 allocs/op
BenchmarkURLString/params_256-10                   67254             17780 ns/op           21698 B/op         20 allocs/op
BenchmarkURLString/params_1024-10                  10000            109537 ns/op           98024 B/op         25 allocs/op
BenchmarkURLKey/params_1-10                      7146849               166.9 ns/op           144 B/op          5 allocs/op
BenchmarkURLKey/params_32-10                     6561790               181.4 ns/op           192 B/op          7 allocs/op
BenchmarkURLKey/params_256-10                    6466377               182.3 ns/op           192 B/op          7 allocs/op
BenchmarkURLKey/params_1024-10                   5680776               211.8 ns/op           192 B/op          7 allocs/op
BenchmarkURLGetCacheInvokerMapKey/params_1-10            5943622               196.2 ns/op           176 B/op          5 allocs/op
[...]

PR #3404, opened 2026-06-12
perf(common): optimize URL.String() locking and string building

### LINKED ISSUE #3393: perf(common): optimize URL.String() locking and string building
### Background

`URL.String()` is used in logging, registry, protocol, and comparison paths. It currently does read-only work but still takes the params write lock, and it builds the URL prefix through formatting before appending encoded params. This path can be made cheaper and less blocking while preserving current output.

Related to #3248.

### Scope

Optimize `common.URL.String()`:

- Use `paramsLock.RLock()` instead of the write lock because the method only reads params.
- Build the URL prefix with `strings.Builder` / `Grow` / `WriteString` instead of formatting into an intermediate string.
- Preserve existing string output exactly, including behavior when there are no query params.
- Keep `params.Encode()` behavior unless a benchmark-backed alternative is introduced with equivalent ordering semantics.

### Acceptance criteria

- Existing `common` tests continue to pass.
- Add or update URL string benchmarks so the PR can show allocation/throughput impact with `-benchmem`.
- No behavior change to URL formatting.

ywxzm03:
**What this PR does**:

Optimizes lock usage and string construction in `common.URL.String()`, making it lighter-weight and less blocking while keeping the existing output exactly unchanged.

This change includes:

- **Locking**: changes `paramsLock.Lock()` to `paramsLock.RLock()`, because this method only reads params. A read lock allows multiple concurrent `String()` calls while still remaining mutually exclusive with writes.
- **String construction**: replaces `fmt.Sprintf` with `strings.Builder` + `Grow` + `WriteString`. By precomputing the required capacity and allocating once, this removes intermediate strings and repeated growth.

Fixes #3393

- The benchmark reuses the existing `BenchmarkURLString` on develop (`common/url_benchmark_test.go`, from #3399). Comparison results from `go test ./common/ -bench BenchmarkURLString -benchmem` on Apple M5:

  | Case        | Before                                | After                                     |
  | ----------- | ------------------------------------- | ----------------------------------------- |
  | params_1    | 383.2 ns/op, 728 B/op, 13 allocs/op   | 309.7 ns/op, 536 B/op, **7 allocs/op**    |
  | params_32   | 1651 ns/op, 3385 B/op, 16 allocs/op   | 1804 ns/op, 3192 B/op, **10 allocs/op**   |
  | params_256  | 18138 ns/op, 21698 B/op, 20 allocs/op | 18499 ns/op, 21496 B/op, **14 allocs/op** |
  | params_1024 | 95501 ns/op, 98025 B/op, 25 allocs/op | 91318 ns/op, 97784 B/op, **19 allocs/op** |

  Latency drops noticeably for small URLs. For large params cases, runtime is mostly dominated by `params.Encode()`, so latency is roughly unchanged, but the allocation count still decreases.

- `go test ./common/` passes.

**Does this PR introduce a user-facing change?**

```release-note
NONE

### DIFF
--- a/common/url.go
+++ b/common/url.go
@@ -395,17 +395,36 @@ func isMatchCategory(category1 string, category2 string) bool {
 }
 
 func (c *URL) String() string {
-	c.paramsLock.Lock()
-	defer c.paramsLock.Unlock()
+	c.paramsLock.RLock()
+	defer c.paramsLock.RUnlock()
+
+	encodedParams := c.params.Encode()
 	var buf strings.Builder
-	if len(c.Username) == 0 && len(c.Password) == 0 {
-		buf.WriteString(fmt.Sprintf("%s://%s:%s%s", c.Protocol, c.Ip, c.Port, c.Path))
-	} else {
-		buf.WriteString(fmt.Sprintf("%s://%s:%s@%s:%s%s", c.Protocol, c.Username, c.Password, c.Ip, c.Port, c.Path))
+
+	size := len(c.Protocol) + len("://") + len(c.Ip) + len(":") + len(c.Port) + len(c.Path)
+	if len(c.Username) != 0 || len(c.Password) != 0 {
+		size += len(c.Username) + len(":") + len(c.Password) + len("@")
 	}
-	if encoded := c.params.Encode(); encoded != "" {
+	if encodedParams != "" {
+		size += len("?") + len(encodedParams)
+	}
+
+	buf.Grow(size)
+	buf.WriteString(c.Protocol)
+	buf.WriteString("://")
+	if len(c.Username) != 0 || len(c.Password) != 0 {
+		buf.WriteString(c.Username)
+		buf.WriteString(":")
+		buf.WriteString(c.Password)
+		buf.WriteString("@")
+	}
+	buf.WriteString(c.Ip)
+	buf.WriteString(":")
+	buf.WriteString(c.Port)
+	buf.WriteString(c.Path)
+	if encodedParams != "" {
 		buf.WriteByte('?')
-		buf.WriteString(encoded)
+		buf.WriteString(encodedParams)
 	}
 	return buf.String()
 }

### CODE REVIEW

**Comment by Alanxtl** [comment]
@Vanillaxi 可以互相review一下  :)

File: common/url.go [hunk 1]
```diff
@@ -395,21 +395,51 @@ func isMatchCategory(category1 string, category2 string) bool {
...
+	}
+	buf.WriteString(c.Ip)
+	buf.WriteString(":")
+	buf.WriteString(c.Port)
+	buf.WriteString(c.Path)
+}
+
+func (c *URL) hasAuth() bool {
+	return len(c.Username) != 0 || len(c.Password) != 0
+}
```
Alanxtl: 这个就不用单独抽象一个函数了

common/url.go
```diff
@@ -395,21 +395,51 @@ func isMatchCategory(category1 string, category2 string) bool {
...
+		buf.WriteString(c.Username)
+		buf.WriteString(":")
+		buf.WriteString(c.Password)
+		buf.WriteString("@")
+	}
+	buf.WriteString(c.Ip)
+	buf.WriteString(":")
+	buf.WriteString(c.Port)
+	buf.WriteString(c.Path)
+}
```
Alanxtl: 这个也不用单独抽象函数了 又没有别的地方用

common/url.go
```diff
@@ -395,21 +395,51 @@ func isMatchCategory(category1 string, category2 string) bool {
...
 	return buf.String()
 }
 
+func (c *URL) stringPrefixLen() int {
+	size := len(c.Protocol) + len("://") + len(c.Ip) + len(":") + len(c.Port) + len(c.Path)
+	if c.hasAuth() {
+		size += len(c.Username) + len(":") + len(c.Password) + len("@")
+	}
+	return size
+}
```
Alanxtl: 这个也不用

Alanxtl (approved):
lgtm, good performance improvement

common/url.go
ywxzm03: fixed

common/url.go
ywxzm03: fixed

common/url.go
ywxzm03: fixed

PR #3410, opened 2026-06-13
perf(loadbalance): reuse timestamp when computing invoker weights

### LINKED ISSUE #3408: perf(loadbalance): reuse timestamp when calculating invoker weights
## Background

`loadbalance.GetWeight` calls `time.Now().Unix()` internally. Several load balance strategies call `GetWeight` once per invoker during a single `Select`, so the current implementation may repeat the same time lookup many times per request.

Reusing one timestamp per selection keeps the warmup calculation consistent within the same selection and avoids repeated calls to `time.Now()`.

## Scope

Introduce a small internal helper that accepts the current Unix timestamp, and update load balance implementations that loop over invokers to compute `now` once per `Select`.

Relevant code:

- `cluster/loadbalance/util.go`
- `cluster/loadbalance/random/loadbalance.go`
- `cluster/loadbalance/roundrobin/loadbalance.go`
- `cluster/loadbalance/leastactive/loadbalance.go`
- other load balance implementations that repeatedly call `GetWeight`

## Acceptance Criteria

- Public API compatibility is preserved, including the existing `GetWeight` function.
- Weight and warmup behavior remains equivalent except that one `Select` call uses one consistent timestamp.
- Existing load balance tests pass.
- Add or update benchmarks if possible to show the effect on selection overhead.
- `go test ./cluster/loadbalance` passes.

HarshMehta112:
Fixes :  #3408

### Description

  Add `GetWeightAt(invoker, invocation, now int64)` alongside the existing
  `GetWeight` function. Load balance strategies that loop over all invokers in
  a single `Select` previously called `time.Now().Unix()` once per invoker
  inside `GetWeight`, repeating the same syscall N times and allowing the
  warmup calculation to drift slightly within a single selection.

  `GetWeight` is preserved unchanged as a thin wrapper, keeping the public API
  fully compatible. The following strategies now compute `now` once per `Select`
  and forward it via `GetWeightAt`: `random`, `roundrobin`, `leastactive`,
  `aliasmethod`, and `iwrr` (init path).

  Benchmark (Apple M4 Pro, 255 invokers):
  | Benchmark          | ns/op | B/op | allocs/op |
  |--------------------|-------|------|-----------|
  | BenchmarkGetWeight    | 120.0 | 192  | 4 |
  | BenchmarkGetWeightAt  |  96.2 | 192  | 4 |
  
  ~20% reduction per weight lookup; savings scale linearly with invoker count.
Fixes # (issue)

### Checklist

### DIFF
--- a/cluster/loadbalance/aliasmethod/alias_method.go
+++ b/cluster/loadbalance/aliasmethod/alias_method.go
@@ -20,6 +20,7 @@ package aliasmethod // weighted random with alias-method algorithm
 
 import (
 	"math/rand"
+	"time"
 )
 
 import (
@@ -56,8 +57,9 @@ func (am *aliasMethodPicker) init(invocation base.Invocation) {
 	small := make([]int, 0, n)
 	large := make([]int, 0, n)
 
+	now := time.Now().Unix()
 	for i, invoker := range am.invokers {
-		weight := loadbalance.GetWeight(invoker, invocation)
+		weight := loadbalance.GetWeightAt(invoker, invocation, now)
 		weights[i] = weight
 		totalWeight += weight
 	}
--- a/cluster/loadbalance/iwrr/iwrr.go
+++ b/cluster/loadbalance/iwrr/iwrr.go
@@ -20,6 +20,7 @@ package iwrr
 import (
 	"math/rand"
 	"sync"
+	"time"
 )
 
 import (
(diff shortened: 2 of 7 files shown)


Alanxtl:
pls comment in the issue before summiting pr

HarshMehta112:
> pls comment in the issue before summiting pr

I will take care of this from next time

Alanxtl (approved):
lgtm

PR #3416, opened 2026-06-14
perf(triple): add fast paths for common content-type canonicalization

### LINKED ISSUE #3405: perf(triple): add fast paths for common content-type canonicalization
## Background

`protocol/triple/triple_protocol.canonicalizeContentType` already has a fast path for simple lowercase content types, but common values with parameters such as `application/json; charset=utf-8` still fall back to `mime.ParseMediaType` and `mime.FormatMediaType`.

This path is used while handling Triple/gRPC request content types, so avoiding the slow parser for the common cases can reduce small per-request allocations.

## Scope

Add explicit fast paths for common content types that are already in canonical form or only need trivial charset normalization. Keep the slow path for uncommon or malformed values.

Relevant code:

- `protocol/triple/triple_protocol/protocol.go`
- `canonicalizeContentType`
- `canonicalizeContentTypeSlow`

## Acceptance Criteria

- Existing content-type behavior remains unchanged.
- Add or update tests covering common fast-path values and slow-path fallback values.
- Add or update benchmarks for simple content type, charset content type, and non-charset parameter content type.
- `go test ./protocol/triple/triple_protocol` passes.

HarshMehta112:
### Description
Fixes #3405 

Summary

  `canonicalizeContentType` already had a fast path for parameter-free types like `application/json`, but content types with ; `charset=utf-8` fell through to `mime.ParseMediaType + mime.FormatMediaType` on every
  request — allocating even when the value was already canonical.

  This PR adds a second fast path that handles the common case of a single ; `charset=<token>` parameter without invoking the MIME parser:
  - If the charset value is already lowercase, returns the original string unchanged (zero allocs).
  - If the charset value needs lowercasing (e.g. `UTF-8`), builds the result with a single string concat instead of going through the full parse/format cycle.
  - All other cases (non-charset params, multiple params, quoted values, malformed input) fall through to `canonicalizeContentTypeSlow` unchanged.
  
  Two helpers extracted from the inline logic: `isSimpleMediaType` (byte-loop, no allocs) and `isSimpleToken` (same).

  Changes

  - `protocol/triple/triple_protocol/protocol.go`: refactored fast path, added charset fast path, added isSimpleMediaType / isSimpleToken
  - `protocol/triple/triple_protocol/protocol_test.go`: added fast-path test cases; moved b.ReportAllocs() before loop; added charset canonical, charset uppercase, non-charset param sub-benchmarks

### Checklist
  - [x] go test ./protocol/triple/triple_protocol/ passes (528 tests)
  - [x] TestCanonicalizeContentType covers both fast-path and slow-path cases

### DIFF
--- a/protocol/triple/triple_protocol/protocol.go
+++ b/protocol/triple/triple_protocol/protocol.go
@@ -335,19 +335,31 @@ func canonicalizeContentType(contentType string) string {
 	//
 	// See https://www.rfc-editor.org/rfc/rfc2045.html#section-5.1 for a full
 	// grammar.
-	var slashes int
-	for _, r := range contentType {
-		switch {
-		case r >= 'a' && r <= 'z':
-		case r == '.' || r == '+' || r == '-':
-		case r == '/':
-			slashes++
-		default:
-			return canonicalizeContentTypeSlow(contentType)
+	semi := strings.IndexByte(contentType, ';')
+	if semi < 0 {
+		if isSimpleMediaType(contentType) {
+			return contentType
 		}
+		return canonicalizeContentTypeSlow(contentType)
 	}
-	if slashes == 1 {
-		return contentType
+
+	// Fast path for a single "; charset=<token>" parameter, avoiding
+	// mime.ParseMediaType + mime.FormatMediaType for the common case of
+	// "application/json; charset=utf-8" and its uppercase variants.
+	base := contentType[:semi]
+	rest := contentType[semi+1:]
+	if isSimpleMediaType(base) && strings.IndexByte(rest, ';') < 0 {
+		const charsetParam = " charset="
+		if strings.HasPrefix(rest, charsetParam) {
+			val := rest[len(charsetParam):]
+			if isSimpleToken(val) {
+				lower := strings.ToLower(val)
+				if lower == val {
+					return contentType
+				}
+				return base + "; charset=" + lower
+			}
+		}
 	}
 	return canonicalizeContentTypeSlow(contentType)
 }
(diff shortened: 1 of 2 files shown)


protocol/triple/triple_protocol/protocol.go
```diff
@@ -367,3 +379,41 @@ func canonicalizeContentTypeSlow(contentType string) string {
...
+		case c >= 'a' && c <= 'z':
+		case c == '.' || c == '+' || c == '-':
+		case c == '/':
+			slashes++
+		default:
+			return false
+		}
+	}
+	return slashes == 1
+}
```
Alanxtl: I think the new charset fast path should make `isSimpleMediaType` stricter before using it for the parameterized case. Right now it only checks that there is exactly one `/`, so malformed bases like `"/; charset=UTF-8"` or `"application/; charset=UTF-8"` take the fast path and get rewritten to lowercase charset. The previous slow path would call `mime.ParseMediaType`, fail, and return the original string unchanged. Since the PR intends malformed input to fall through unchanged, can we require both sides of `/` to be non-empty before accepting the fast path?

Alanxtl (commented):
I suggest 
```golang
func canonicalizeContentType(contentType string) string {
	if canonical, ok := canonicalizeContentTypeFast(contentType); ok {
		return canonical
	}
	return canonicalizeContentTypeSlow(contentType)
}

func canonicalizeContentTypeFast(contentType string) (string, bool) {
	semi := strings.IndexByte(contentType, ';')
	if semi < 0 {
		if isLowercaseMediaType(contentType) {
			return contentType, true
		}
		return "", false
	}

	base := contentType[:semi]
	param := contentType[semi+1:]

	if !isLowercaseMediaType(base) {
		return "", false
	}
	if strings.Contains(param, ";") {
		return "", false
	}

	const prefix = " charset="
	if !strings.HasPrefix(param, prefix) {
		return "", false
	}

	charset := param[len(prefix):]
	if !isSimpleCharsetToken(charset) {
		return "", false
	}

	lower := strings.ToLower(charset)
	if lower == charset {
		return contentType, true
	}
	return base + "; charset=" + lower, true
}


func isLowercaseMediaType(s string) bool {
	slash := strings.IndexByte(s, '/')
	if slash <= 0 || slash == len(s)-1 {
		return false
	}
	if strings.IndexByte(s[slash+1:], '/') >= 0 {
		return false
	}
	return isLowercaseToken(s[:slash]) && isLowercaseToken(s[slash+1:])
}

func isLowercaseToken(s string) bool {
	if s == "" {
		return false
	}
	for i := 0; i < len(s); i++ {
		c := s[i]
		switch {
		case c >= 'a' && c <= 'z':
		case c >= '0' && c <= '9':
		case c == '.' || c == '+' || c == '-':
		default:
			return false
		}
	}
	return true
}
```

protocol/triple/triple_protocol/protocol.go
HarshMehta112: Good catch. Fixed — `isLowercaseMediaType` now checks `slash > 0` and `slash < len(s)-1`, so both sides of `/` must be non-empty. Malformed inputs like `/json; charset=UTF-8` and `application/; charset=UTF-8` now fall through to `canonicalizeContentTypeSlow` and return unchanged. Added test cases for both. Also took the opportunity to split the fast-path logic into `canonicalizeContentTypeFast` as suggested and renamed helpers for clarity (`isLowercaseToken`, `isSimpleCharsetToken`).

Alanxtl (approved):
lgtm

PR #3428, opened 2026-06-15
fix(triple): canonicalize context metadata headers

### LINKED ISSUE #2422: Triple header/trailer usage pattern proposal
New **Triple** is based on connect-go (https://github.com/connectrpc/connect-go) which makes use of Golang generic feature. For instance, client and server side code of unary invocation would be like this:
```go
// client side
Ping(context.Context, *connect_go.Request[v1.PingRequest]) (*connect_go.Response[v1.PingResponse], error)

// server side
Ping(context.Context, *connect_go.Request[v1.PingRequest]) (*connect_go.Response[v1.PingResponse], error)

// uniform Request
type Request[T any] struct {
	Msg *T

	spec   Spec
	peer   Peer
	header http.Header
}

// uniform Response
type Response[T any] struct {
	Msg *T

	header  http.Header
	trailer http.Header
}
```
Users could inject or receive headers and trailers directly from Request and Response just like this:
```go
// client side
req := connect.NewRequest(&pingv1.PingRequest{})
req.Header().Set("hello", "wourld")
resp, err := Ping(context.Background(), req)
headers := resp.Header().Values("hello")
trailers := resp.Trailer().Values()
// do something with headers and trailers
```
```go
// server side
func (p *pingServer) Ping(ctx context.Context, req *connect.Request[pingv1.PingRequest]) (*connect.Response[pingv1.PingResponse], error) {
	headers := req.Headers().Values("hello")
        // do something with headers

	response := connect.NewResponse(
		&pingv1.PingResponse{},
	)
        // return headers and trailers
	response.Header().Set(handlerHeader, headerValue)
	response.Trailer().Set(handlerTrailer, trailerValue)
	return response, nil
}
```
This pattern is very uniform and easy to understand. But since we decided not to use generics API for compatibility, new **Triple** has to make use of some grpc-like pattern to send and receive headers and trailers.

wyf027:
## What

Canonicalize Triple context metadata header keys when storing incoming and outgoing headers. This keeps `NewOutgoingContext` and `AppendToOutgoingContext` values under the same key and makes standard `http.Header.Get/Values` access work as documented.

Also document how unary clients read response headers and trailers from the response wrapper.

Refs #2422

## Validation

- `go test ./protocol/triple/triple_protocol`
- `go test ./protocol/triple`
- `git diff --check`

### DIFF
--- a/protocol/invocation/rpcinvocation.go
+++ b/protocol/invocation/rpcinvocation.go
@@ -267,10 +267,11 @@ func (r *RPCInvocation) MergeAttachmentFromContext(ctx context.Context) {
 		return
 	}
 	for k, v := range header {
+		key := strings.ToLower(k)
 		if len(v) == 1 {
-			r.SetAttachment(k, v[0])
+			r.SetAttachment(key, v[0])
 		} else {
-			r.SetAttachment(k, v)
+			r.SetAttachment(key, v)
 		}
 	}
 }
--- a/protocol/invocation/rpcinvocation_test.go
+++ b/protocol/invocation/rpcinvocation_test.go
@@ -371,13 +371,9 @@ func TestRPCInvocation_GetAttachmentAsContext(t *testing.T) {
 	header := triple_protocol.ExtractFromOutgoingContext(ctx)
 	assert.NotNil(t, header)
 
-	// Verify that string attachments are in the header
-	// NewOutgoingContext stores keys as lowercase, so check both ways
-	assert.Contains(t, header, "key1")
-	assert.Equal(t, []string{"value1"}, header["key1"]) //nolint:staticcheck
-
-	assert.Contains(t, header, "key2")
-	assert.Equal(t, []string{"value2", "value3"}, header["key2"]) //nolint:staticcheck
+	// Verify that string attachments are in the header.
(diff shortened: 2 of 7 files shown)


Alanxtl:
This looks like a reasonable compromise to me. Since these APIs expose `http.Header`, storing context headers in canonical form makes standard `Header.Get/Values` work as users would expect. At the same time, converting keys back to lowercase when merging into `Invocation` attachments keeps the Dubbo/gRPC metadata semantics aligned with grpc-go, dubbo-java, and HTTP/2 wire behavior.

Could we add a short comment around the `CanonicalHeaderKey` / `ToLower` boundary to make this intentional layering explicit? Something like: context headers use canonical `http.Header` keys for Go API compatibility, while invocation attachments remain lowercase metadata keys for Dubbo/gRPC compatibility.

wyf027:
Added short comments around the canonical context-header storage and lowercase exported-attachment boundary in 7fd64c2. Local validation passes: `go test ./protocol/triple/triple_protocol`, `git diff --check`, and `golangci-lint run ./protocol/triple/triple_protocol --timeout=10m`.

Alanxtl (approved):
lgtm

PR #3432, opened 2026-06-17
fix: toKey now uses configured hash.arguments indexes instead of slice positions

### LINKED ISSUE #3431: [BUG] consistenthashing: toKey ignores configured hash.arguments indexes
## Problem

In `cluster/loadbalance/consistenthashing/selector.go`, `newSelector` parses the configured `hash.arguments` values into `selector.argumentIndex`:

```go
selector.argumentIndex = append(selector.argumentIndex, i)
```

However, `toKey` currently iterates with `range` over the slice and uses the loop index instead of the configured value:

```go
for i := range c.argumentIndex {
    if i >= 0 && i < len(args) {
        _, _ = fmt.Fprint(&sb, args[i])
    }
}
```

This means `methods.<method>.hash.arguments=1` still uses `args[0]`, because `i` is the position in `argumentIndex`, not the configured argument index.

## Java Dubbo behavior

Java Dubbo iterates over the configured argument index values and then reads `args[i]`:

```java
private String toKey(Object[] args) {
    StringBuilder buf = new StringBuilder();
    for (int i : argumentIndex) {
        if (i >= 0 && args != null && i < args.length) {
            buf.append(args[i]);
        }
    }
    return buf.toString();
}
```

So `hash.arguments=1` should use the second request argument, not the first configured-position entry.

## Example

With:

```text
methods.echo.hash.arguments=1
args = []any{"tenantA", 42}
```

Expected key should be based on `args[1]`, i.e. `"42"`.

Actual key is based on `args[0]`, i.e. `"tenantA"`.

This can make consistent hashing route by the wrong request argument and may produce different routing behavior from the configured intent and from Java Dubbo's implementation.

## Suggested fix

Use the configured argument index values when building the key:

```go
for _, i := range c.argumentIndex {
    if i >= 0 && i < len(args) {
        _, _ = fmt.Fprint(&sb, args[i])
    }
}
```

It would also be useful to add regression tests for non-zero `hash.arguments`, for example `hash.arguments=1` with `[]any{"ignored", 123}` should produce key `"123"`.

## Optional performance follow-up

After the index bug is fixed, `toKey` can keep the non-string safety while avoiding the `fmt` path for the common string case:

```go
for _, i := range c.argumentIndex {
    if i >= 0 && i < len(args) {
        switch v := args[i].(type) {
        case string:
            sb.WriteString(v)
        default:
            _, _ = fmt.Fprint(&sb, v)
        }
    }
}
```

This is not required for correctness, but it preserves the safe behavior for non-string arguments and gives the string hot path a cheaper append than `fmt.Fprint`.

lesbass:
## Summary

Fixes #3431.

In `cluster/loadbalance/consistenthashing/selector.go`, `toKey` was using the slice index from `range c.argumentIndex` as the argument position to read from `args`. With a configuration like `hash.arguments=1`, this still read `args[0]` instead of `args[1]`, causing consistent hashing to route by the wrong request argument.

## Changes

1. **`selector.go`**: Changed `for i := range c.argumentIndex` to `for _, i := range c.argumentIndex` so that `i` is the configured argument index, not the slice position.
2. **`selector.go`**: Replaced `args[i].(string)` with `fmt.Fprint(&sb, args[i])` to avoid panicking on non-string arguments.
3. **`loadbalance_test.go`**: Added `TestToKeyWithArgumentIndex` regression test with `hash.arguments=1` verifying that the key is built from the correct argument position.

## Verification

```
$ go test -v -count=1 ./cluster/loadbalance/consistenthashing/
=== RUN   TestConsistentHashSelectorSuite
=== RUN   TestConsistentHashSelectorSuite/TestSelectForKey
=== RUN   TestConsistentHashSelectorSuite/TestToKey
--- PASS: TestConsistentHashSelectorSuite (0.00s)
=== RUN   TestToKeyWithArgumentIndex
--- PASS: TestToKeyWithArgumentIndex (0.00s)
=== RUN   TestConsistentHashLoadBalanceSuite
=== RUN   TestConsistentHashLoadBalanceSuite/TestConcurrentSelect
=== RUN   TestConsistentHashLoadBalanceSuite/TestSelect
--- PASS: TestConsistentHashLoadBalanceSuite (0.00s)
PASS
ok      dubbo.apache.org/dubbo-go/v3/cluster/loadbalance/consistenthashing    0.007s
```

### DIFF
--- a/cluster/loadbalance/consistenthashing/loadbalance_test.go
+++ b/cluster/loadbalance/consistenthashing/loadbalance_test.go
@@ -68,6 +68,24 @@ func (s *consistentHashSelectorSuite) TestToKey() {
 	s.Equal("usernameage", result)
 }
 
+// TestToKeySingleNonZeroIndex verifies that a single non-zero hash.arguments
+// index produces a key from the correct argument position.
+func TestToKeyWithArgumentIndex(t *testing.T) {
+	var invokers []base.Invoker
+	// hash.arguments=1 means only the second argument (index 1) is used
+	url, err := common.NewURL("dubbo://192.168.1.0:20000/org.apache.demo.HelloService?methods.echo.hash.arguments=1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	invokers = append(invokers, base.NewBaseInvoker(url))
+	sel := newSelector(invokers, "echo", 999944)
+
+	key := sel.toKey([]any{"ignored", "used"})
+	if key != "used" {
+		t.Errorf("expected key %q, got %q", "used", key)
+	}
+}
+
 func (s *consistentHashSelectorSuite) TestSelectForKey() {
 	url1, _ := common.NewURL(url8080Short)
 	url2, _ := common.NewURL(url8081Short)
--- a/cluster/loadbalance/consistenthashing/selector.go
+++ b/cluster/loadbalance/consistenthashing/selector.go
@@ -85,7 +85,12 @@ func (c *selector) toKey(args []any) string {
 	var sb strings.Builder
 	for _, i := range c.argumentIndex {
 		if i >= 0 && i < len(args) {
-			_, _ = fmt.Fprint(&sb, args[i])
+			switch v := args[i].(type) {
+			case string:
+				sb.WriteString(v)
+			default:
+				_, _ = fmt.Fprint(&sb, v)
+			}
 		}
 	}
 	return sb.String()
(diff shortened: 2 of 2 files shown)


Alanxtl:
1. the target branch should be develop
2. take a look at the issue carefully, especially the last section
3. comment in the issue before summitting pr

lesbass:
@Alanxtl thanks for the review. I've addressed all three points:

1. Rebased onto `develop` instead of `main`
2. Added the type-switch optimization from the issue's last section — strings now use `sb.WriteString` directly instead of `fmt.Fprint`
3. Commented on #3431

Updated diff now has both the index fix (`for _, i := range`) and the string fast path.

Alanxtl (approved):
lgtm

PR #3434, opened 2026-06-17
fix(getty): check error before using buf in server Write

Aias00:
## Problem

In `RpcServerPackageHandler.Write` (`remoting/getty/readwriter.go`), both `EncodeResponse` and `EncodeRequest` were using `buf.Len()` before checking `err`. If the encode call returns an error, `buf` could be nil, causing a nil pointer panic at `buf.Len()`.

## Fix

Move the `err != nil` check before `buf.Len()` in both code paths, aligning with the client-side `RpcClientPackageHandler.Write` which already checks `err` first.

### Before (server EncodeResponse, line 130-139)
```go
buf, err := (p.server.codec).EncodeResponse(res)
bufLength := buf.Len()   // ← panic if err != nil and buf is nil
if bufLength > maxBufLength { ... }
if err != nil { ... }    // ← too late
```

### After
```go
buf, err := (p.server.codec).EncodeResponse(res)
if err != nil { ... }    // ← check first
bufLength := buf.Len()   // ← safe now
if bufLength > maxBufLength { ... }
```

Same fix applied to the `EncodeRequest` path (line 145-154).

Co-Authored-By: Claude <noreply@anthropic.com>


AlexStocks (approved):

Alanxtl (approved):
lgtm

PR #3438, opened 2026-06-17
fix(remoting): use atomic.Bool for ExchangeClient.init to fix data race

Aias00:
## Problem

`ExchangeClient.init` is a plain `bool` field with a `// FIXME atomic operation` comment acknowledging the race. It is written in `doInit`/`Close` and read (indirectly via `doInit`) in `Request`/`AsyncRequest`/`Send` without any synchronization.

**Race scenarios:**
- Multiple goroutines concurrently call `Request`/`AsyncRequest`/`Send` → all enter `doInit` → all see `init == false` → `Connect` is called multiple times
- `Close()` writes `init = false` concurrently with `doInit` reading/writing `init` → memory visibility issue

## Fix

Replace `init bool` with `uatomic.Bool` (already imported in this file as `go.uber.org/atomic`), consistent with the existing `activeNum uatomic.Uint32` field in the same struct:

| Change | Detail |
|--------|--------|
| `init bool` → `init uatomic.Bool` | Same package as `activeNum`, already imported |
| `doInit` check-then-set → `CAS(false, true)` | Atomic compare-and-swap ensures only one goroutine performs `Connect` |
| CAS failure → spin-wait | Goroutines that lose the CAS wait for init to complete instead of skipping or duplicating |
| Connect failure → `Store(false)` | Reset flag on failure so future calls can retry (important behavior missing in original code) |
| `Close` → `Store(false)` | Atomic write ensures visibility |

This follows the same pattern used by `BaseInvoker` (`uatomic.Bool`), `BaseClusterInvoker` (`atomic.Bool` with CAS), and `Directory` (`atomic.Bool` with CAS + mutex) elsewhere in the codebase.

## Tests

- **`TestExchangeClientConcurrentDoInit`**: 20 goroutines concurrently call `doInit` — verifies `Connect` is called exactly once
- **`TestExchangeClientDoInitFailureResets`**: verifies failed `doInit` resets the init flag and retry succeeds
- All existing tests pass with `-race` flag enabled

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Aias00:
## Addressed Copilot Review Feedback

All three review comments have been addressed in the latest commit (8bb4ab5):

### 1. Three-state atomic instead of two-state bool
**Before:** `uatomic.Bool` with CAS(false→true) before Connect, causing waiters to see init==true before initialization completes.

**After:** Replaced with `uatomic.Int32` using three states:
- `0` = uninitialized
- `1` = initializing  
- `2` = initialized

CAS now transitions `0→1` (not `0→2`), so waiters never observe a "ready" state until Connect actually succeeds. On failure, state resets to `0` for retry.

### 2. sync.Cond instead of spin-wait
**Before:** `for !cl.init.Load() { time.Sleep(10ms) }` — polling with fixed sleep.

**After:** `sync.Cond` with `Broadcast()` when initialization completes (success or failure). Waiters call `cl.initCond.Wait()` and are woken immediately when the initializing goroutine finishes, eliminating both latency and CPU waste.

### 3. Test: buffered channel instead of t.Errorf from goroutines
**Before:** `t.Errorf` called directly from concurrent goroutines.

**After:** Errors are collected into a buffered channel and asserted from the main test goroutine after `wg.Wait()`.

Additionally, all concurrent waiters now receive the same error result when initialization fails (previously they silently returned nil).

PR #3461, opened 2026-06-29
fix(#3453): resolve concurrency, leak, and panic issues from code audit

### LINKED ISSUE #3453: Code audit: memory leaks, goroutine leaks, deadlocks and performance issues
## Overview

I re-checked this issue against the current `develop` branch of `apache/dubbo-go` locally, based on source inspection instead of keeping the original audit text unchanged.

This version keeps only items that are still justified by current code, removes or downgrades items that no longer hold, and adds a small number of additional source-grounded findings that are worth community follow-up.

The goal of this issue is triage quality, not volume: every remaining item below is something that still looks actionable on current `develop`.

---

### 1. RoundRobin load balancer selection is not an atomic operation under concurrency

**File**: `cluster/loadbalance/roundrobin/loadbalance.go`

Relevant code path:
- `increaseCurrent()` uses `atomic.AddInt64`
- the outer selection loop compares `currentWeight > maxCurrentWeight`
- selection and `Current(totalWeight)` adjustment are not protected as one atomic transaction

Why this still matters:
- individual field updates are atomic,
- but the full weighted-round-robin selection procedure is not,
- so concurrent callers can observe and mutate intermediate state and skew distribution.

Potential impact:
- uneven request distribution under load,
- especially for services relying on weighted balancing fairness.

Suggested remediation:
- protect the full selection/update sequence with a stronger synchronization model,
- or redesign the algorithm so the multi-step state transition is concurrency-safe as a whole rather than only per-field atomic.

---

### 2. Getty server still panics on configuration handling failures

**File**: `remoting/getty/getty_server.go`

Relevant code path still contains direct `panic(err)` in config loading / validation paths, including YAML decode and `srvConf.CheckValidity()` failure handling.

Why this still matters:
- malformed protocol config can still escalate to process-level failure,
- instead of returning an error to caller and letting upper layers decide startup behavior.

Potential impact:
- startup failure during config rollout,
- poor fault isolation for misconfiguration.

Suggested remediation:
- replace panic-based config handling with explicit error propagation,
- return structured startup errors to the caller,
- reserve panic only for invariant violations that truly indicate impossible internal state.

---

### 3. Graceful termination state handling still deserves hardening

**File**: `graceful_shutdown/shutdown.go`

Relevant state:
- package-level `shutdownResult error`
- background goroutine assigns it in `startShutdownOnce()`
- other goroutines read it in `Shutdown()` after waiting on `shutdownDone`
- timeout fallback still uses delayed forced process exit via `os.Exit(0)`

Why this still matters:
- the main code path works, but the termination flow remains fairly blunt,
- and result/state publication is fragile enough that it deserves explicit hardening rather than relying on subtle ordering assumptions.

Potential impact:
- difficult-to-reason-about shutdown behavior,
- incomplete cleanup on timeout path,
- weaker maintainability for future changes in graceful termination.

Suggested remediation:
- make result publication explicit with clearer synchronization semantics,
- isolate forced-exit behavior behind a narrower policy boundary,
- and ensure all cleanup/notification paths are either completed or explicitly abandoned with observable logging.

---

### 4. TPS limiter cache is unbounded

**File**: `filter/tps/limiter/method_service.go`

Relevant code path:
- `MethodServiceTpsLimiter` stores state in `tpsState`
- entries are inserted by `LoadOrStore(limitTarget, ...)`
- no cleanup / TTL / size bound exists

Why this still matters:
- distinct service+method combinations accumulate for process lifetime.

Potential impact:
- long-lived provider processes can accumulate stale limiter entries,
- memory usage can drift upward in dynamic service/method scenarios.

Suggested remediation:
- add eviction by TTL, LRU, or explicit cleanup on service lifecycle changes,
- or scope limiter state to a registry/service lifecycle instead of permanent process lifetime.

---

### 5. Script router JS runtime reuse may retain cross-request state

**File**: `cluster/router/script/instance/js_instance.go`

Relevant code path:
- `jsInstances.insPool` reuses `*goja.Runtime`
- per-run code resets selected bindings (`invokers`, `invocation`, `context`, result var)
- runtime-global state created by user script is not comprehensively reset before reuse

Why this still matters:
- script-defined globals may survive in pooled runtime instances,
- which can create cross-run state pollution.

Potential impact:
- routing behavior may depend on prior requests or prior script execution state,
- difficult-to-debug correctness issues in production.

Suggested remediation:
- either fully reset runtime state before returning instances to pool,
- or avoid pooling mutable runtime instances and instead pool only immutable compiled programs.

---

### 6. Forking cluster mode does not actively stop losing parallel requests after the first response wins

**File**: `cluster/cluster/forking/cluster_invoker.go`

Relevant code path:
- `Invoke()` starts one goroutine per selected invoker,
- waits only for the first response via `resultQ.Poll(1, ...)`,
- then returns immediately,
- but the other in-flight goroutines continue executing `k.Invoke(ctx, invocation)` until they complete on their own.

Why this still matters:
- the “winner” response ends the user-visible call,
- but the losing parallel requests are not actively canceled or short-circuited by the cluster implementation.

Potential impact:
- unnecessary downstream load,
- wasted CPU / network / backend work,
- extra pressure when `forks` is large or providers are slow.

Suggested remediation:
- derive a cancelable child context inside forking mode,
- cancel sibling requests as soon as the first successful result is accepted,
- or otherwise provide bounded winner-take-all semantics with explicit cleanup of losing branches.

---

### 7. Several runtime paths still use panic/fatal style failure handling where ordinary error propagation would be safer

Representative files:
- `registry/servicediscovery/service_discovery_registry.go`
- `filter/sentinel/filter.go`
- `dubbo.go`
- `loader.go`
- `metrics/prometheus/registry.go`

Representative examples:
- `metadata.GetMetadataInfo(...) == nil` triggers panic in service registration path
- Sentinel default initialization failure triggers panic in filter construction
- background `srv.Serve()` failure in `dubbo.go` uses `logger.Fatalf(...)`
- config loader eagerly panics on file read / suffix validation failures

Why this still matters:
- these are not all “memory leak” or “deadlock” bugs,
- but they are operational stability issues and can turn recoverable deployment/runtime problems into full process termination.

Potential impact:
- poor fault isolation,
- difficult rollout behavior,
- production exits caused by configuration or integration failures.

Suggested remediation:
- audit panic/fatal sites and classify them into:
  1. genuine invariants that may remain fatal,
  2. configuration / dependency / startup errors that should be returned,
  3. background serving failures that should be surfaced through lifecycle management instead of unconditional fatal termination.

---

### A. Accesslog `send on closed channel` panic

**File**: `filter/accesslog/filter.go`

Original claim:
- `shutdown()` closes `logChan`
- concurrent `logIntoChannel()` may panic with `send on closed channel`

Re-check result:
- this was valid in current local source before patching,
- and I reproduced it with a focused test,
- but it has now been fixed in my local working tree by switching shutdown signaling away from directly closing `logChan`.

Local regression test used:
- `filter/accesslog/memory_leak_test.go`
- `TestAccessLogFilterLogIntoChannelAfterShutdownDoesNotPanic`

Current issue status:
- keep as historical note for now,
- remove this item once the patch is merged.

---

### B. Service Discovery metadata cache grows unbounded

**File**: `registry/servicediscovery/service_instances_changed_listener_impl.go`

Original claim:
- `revisionToMetadata` keeps accumulating stale revisions forever

Re-check result on current `develop`:
- `OnEvent()` rebuilds `newRevisionToMetadata`
- then replaces `lstn.revisionToMetadata` with the rebuilt map
- so the in-memory listener map is no longer a monotonic append-only structure

There have also been follow-up fixes in this area on current history (`#3353`, `#3369`, `#3395`, `#3414` related work).

Current issue status: remove this item; original description is no longer accurate on current `develop`.

---

### C. ProviderActiveCount counter leak

**File**: `filter/graceful_shutdown/provider_filter.go`

Original claim:
- active counter increments but may fail to decrement on error path

Re-check result on current `develop`:
- current code marks counted results via attachment
- `OnResponse()` decrements conditionally through `shouldDecrementProviderActive(...)`

Current issue status: already fixed upstream; remove this item.

---

### Getty client deadlock claim

**File**: `remoting/getty/getty_client.go`

I do see nested lock usage:
- `c.mux.RLock()`
- then `c.gettyClientMux.Lock()` / `RLock()` in `selectSession()`

However, from current source inspection alone I do **not** yet have strong enough proof of a real deadlock cycle on `develop`.

So this should not remain stated as a confirmed bug in its current wording.

If someone wants to keep tracking it, it should be rewritten as:
- a **lock-ordering risk**,
- plus a concrete reproducer, goroutine dump, or race/deadlock evidence.

---

## Suggested Follow-up Order

A reasonable community follow-up order would be:
1. RoundRobin concurrency skew
2. Getty config panic path
3. Forking cluster losing-branch cancellation
4. TPS limiter cache cleanup
5. graceful termination hardening
6. JS runtime pooling isolation
7. broader panic/fatal policy audit

---

## Notes on Scope

This issue is now intentionally narrower and more evidence-driven than the original version.

I did **not** keep every suspicious pattern I saw in the tree. Some things may still deserve benchmark or stress testing, but if they do not yet have enough source-grounded evidence, they should not be stated here as confirmed bugs.

HarshMehta112:
Fixes #3453

Addresses six source-grounded findings around concurrency correctness, memory/goroutine
  leaks, and panic-based failure handling. Each fix is narrowly scoped with focused tests.

  #### 1. RoundRobin selection now atomic under concurrency
  `cluster/loadbalance/roundrobin/loadbalance.go`
  The weighted-round-robin select+update sequence was only per-field atomic, so concurrent
  callers could observe each other's intermediate `currentWeight` and skew distribution.
  Added a per-key `sync.Mutex` on `cachedInvokers` so the full select+update transaction is
  serialized per service+method (contention bounded to the same key).

  #### 2. Getty server no longer panics on config errors
  `remoting/getty/getty_server.go`
  Replaced `panic(err)` in config marshal/unmarshal, `CheckValidity()`, and `newSession()`
  with structured error logging / error returns. Added `safeYAMLMarshal` to recover the
  panic `yaml.v3` raises on unmarshalable types (channels, funcs) and surface it as an error.
  Malformed protocol config no longer escalates to process-level failure.

  #### 3. Graceful shutdown result publication hardened
  `graceful_shutdown/shutdown.go`
  Added `shutdownResultMu` guarding all reads/writes of the package-level `shutdownResult`,
  removing reliance on subtle ordering assumptions between the background shutdown goroutine
  and concurrent `Shutdown()` readers.

  #### 4. TPS limiter cache is now bounded
  `filter/tps/limiter/method_service.go`
  The limiter map grew for process lifetime as distinct service+method combos accumulated.
  Wrapped each strategy in a `tpsLimitEntry` tracking last-access time, switched to `sync.Map`,
  and added a singleton background `runCleanup` goroutine that evicts entries idle past a TTL.

  #### 5. Script-router JS runtime no longer leaks cross-request state
  `cluster/router/script/instance/js_instance.go`
  Pooled `goja.Runtime` instances were never returned to the pool, and request/script globals
  survived reuse. Added `defer reset()+Put()` in `Run()`. `reset()` snapshots the built-in
  global keys at construction and deletes every global outside that set — clearing both request
  bindings and user-script-defined globals while preserving built-ins.

  #### 6. Forking cluster cancels losing parallel branches
  `cluster/cluster/forking/cluster_invoker.go`
  Losing parallel invocations ran to completion after the winner returned, wasting downstream
  work. Derived a cancelable `forkCtx`; `defer cancel()` stops sibling requests as soon as the
  first accepted result wins.

  ### Tests
  - `TestRoundRobinByWeightConcurrent` — 50 goroutines × 30 selections, race-clean
  - `TestInitServerDoesNotPanicOnUnmarshalableParams` — channel in Params no longer panics
  - `TestShutdownResultConsistentUnderConcurrentReaders`, `TestShutdownPanicSetsResultError`
  - `TestTpsLimitEntryUpdatesLastAccess`, `TestTpsLimiterStaleEntryEviction`
  - `TestJsInstanceResetClearsBindings` (incl. user-defined global), `TestJsInstanceRunReturnsRuntimeToPool`
  - `TestForkingInvokeCancelsLosingBranches` — loser branch ctx canceled after winner succeeds

  All 74 tests across the 6 packages pass; `go vet` and `gofmt` clean.

  ### Scope
  Covers audit items 1–6 only. Item 7 (broader panic/fatal policy audit) is left for follow-up
  PRs per the issue's suggested slicing. Items A/B/C are already fixed upstream.

  ### Checklist

### DIFF
--- a/cluster/cluster/forking/cluster_invoker.go
+++ b/cluster/cluster/forking/cluster_invoker.go
@@ -71,10 +71,15 @@ func (invoker *forkingClusterInvoker) Invoke(ctx context.Context, invocation pro
 		}
 	}
 
+	// forkCtx is canceled when Invoke returns, signaling losing parallel
+	// goroutines to stop rather than running to completion.
+	forkCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	resultQ := queue.New(1)
 	for _, ivk := range selected {
 		go func(k protocolbase.Invoker) {
-			result := k.Invoke(ctx, invocation)
+			result := k.Invoke(forkCtx, invocation)
 			if err := resultQ.Put(result); err != nil {
 				logger.Errorf("[Cluster][Forking] resultQ put failed with exception err=%v", err)
 			}
--- a/cluster/cluster/forking/cluster_test.go
+++ b/cluster/cluster/forking/cluster_test.go
@@ -163,3 +163,38 @@ func TestForkingInvokeHalfTimeout(t *testing.T) {
 	assert.Equal(t, mockResult, result)
 	wg.Wait()
 }
+
+func TestForkingInvokeCancelsLosingBranches(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	forkingUrl.AddParam(constant.ForksKey, strconv.Itoa(2))
+
+	loserCtxCanceled := make(chan struct{})
+
+	winner := mock.NewMockInvoker(ctrl)
+	winner.EXPECT().IsAvailable().Return(true).AnyTimes()
+	winner.EXPECT().Invoke(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(_ context.Context, _ base.Invocation) result.Result {
+			return &result.RPCResult{}
+		})
+
+	loser := mock.NewMockInvoker(ctrl)
+	loser.EXPECT().IsAvailable().Return(true).AnyTimes()
+	loser.EXPECT().Invoke(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(ctx context.Context, _ base.Invocation) result.Result {
+			<-ctx.Done()
+			close(loserCtxCanceled)
+			return &result.RPCResult{}
+		})
+
+	clusterInvoker := registerForking(winner, loser)
+	res := clusterInvoker.Invoke(context.Background(), &invocation.RPCInvocation{})
+	assert.NoError(t, res.Error())
+
+	select {
+	case <-loserCtxCanceled:
+	case <-time.After(2 * time.Second):
+		t.Fatal("loser branch context was not canceled after winner succeeded")
+	}
+}
--- a/cluster/cluster/forking/cluster_test.go
+++ b/cluster/cluster/forking/cluster_test.go
@@ -190,7 +190,7 @@ func TestForkingInvokeCancelsLosingBranches(t *testing.T) {
 
 	clusterInvoker := registerForking(winner, loser)
 	res := clusterInvoker.Invoke(context.Background(), &invocation.RPCInvocation{})
-	assert.NoError(t, res.Error())
+	require.NoError(t, res.Error())
 
 	select {
 	case <-loserCtxCanceled:
--- a/cluster/loadbalance/roundrobin/loadbalance.go
+++ b/cluster/loadbalance/roundrobin/loadbalance.go
@@ -69,6 +69,12 @@ func (lb *rrLoadBalance) Select(invokers []base.Invoker, invocation base.Invocat
 	cache, _ := methodWeightMap.LoadOrStore(key, &cachedInvokers{})
 	cachedInvokers := cache.(*cachedInvokers)
 
+	// Serialize the full select+update sequence per service+method key so that
+	// concurrent callers cannot observe each other's intermediate currentWeight
+	// state and skew the weighted distribution.
+	cachedInvokers.mu.Lock()
+	defer cachedInvokers.mu.Unlock()
+
 	var (
 		clean               = false
 		totalWeight         = int64(0)
@@ -166,5 +172,6 @@ func (robin *weightedRoundRobin) Current(delta int64) {
 }
 
 type cachedInvokers struct {
+	mu       sync.Mutex
 	sync.Map /*[string]weightedRoundRobin*/
 }
--- a/cluster/loadbalance/roundrobin/loadbalance_test.go
+++ b/cluster/loadbalance/roundrobin/loadbalance_test.go
@@ -20,6 +20,7 @@ package roundrobin
 import (
 	"fmt"
 	"strconv"
+	"sync"
 	"testing"
 )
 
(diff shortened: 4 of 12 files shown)


Alanxtl:
1. comment in the issue before summitting pr
2. split this pr into 6 small prs, this one is too large to review

HarshMehta112:
Closing this PR as suggested above. I'll split it into six smaller PRs for easier review.

PR #3462, opened 2026-06-29
test(triple): add concurrent send/close guards for duplexHTTPCall

### LINKED ISSUE #3458: Investigate duplexHTTPCall concurrent Send and CloseAndReceive safety
### Background

Dubbo-go Triple is built on a forked and heavily modified copy of connect-go runtime code under `protocol/triple/triple_protocol`. We do not depend on upstream connect-go directly anymore, but the core stream transport shape, including `duplexHTTPCall`, still comes from that lineage.

Because of that, upstream connect-go runtime fixes can still be relevant to Dubbo-go when they touch shared transport/lifecycle code. One such fix landed in connect-go v1.19.2: `Fix nil pointer deref in duplexHTTPCall under concurrent Send + CloseAndReceive` (connectrpc/connect-go#919).

Our local Triple runtime also has `duplexHTTPCall` and client stream APIs where `Send`, `CloseRequest`, `CloseResponse`, `ResponseHeader`, and `ResponseTrailer` may be used around the same stream lifecycle. This should be checked against the upstream fix instead of doing a broad connect-go upgrade.

### Scope

- Inspect `protocol/triple/triple_protocol/duplex_http_call.go` for the same concurrent `Send` + close/receive race.
- Add a focused regression test that exercises concurrent send and close/receive behavior for Triple client streams.
- Port only the relevant synchronization/lifecycle fix if the race is reproducible or structurally possible.

### Notes

This is not a request to adopt unrelated connect-go features such as HTTP GET, transport retry, or new codegen flags. The goal is to selectively backport the runtime safety fix that matters for Dubbo-go Triple streaming.

Reference: https://github.com/connectrpc/connect-go/releases/tag/v1.19.2

chaojixinren:
### Description

Investigate whether connect-go v1.19.2's fix for the concurrent `Send` +
`CloseAndReceive` nil-pointer dereference (connectrpc/connect-go#919) applies to
Dubbo-go's forked Triple runtime under `protocol/triple/triple_protocol`.

**Conclusion: `duplexHTTPCall` is not affected.** `newDuplexHTTPCall` allocates
the request-body `io.Pipe` at construction and never reassigns
`requestBodyWriter`, so it is always non-nil and concurrent `Write`/`CloseWrite`
are safe in any order. The upstream race — where the pipe was created lazily on
the first `Send` and a racing `CloseWrite` could leave it nil — is structurally
impossible here, so no synchronization change is ported.

What this PR does:

- adds `TestDuplexHTTPCallConcurrentWriteAndCloseWrite` and
  `TestDuplexHTTPCallConcurrentWriteCloseRace` as regression guards (pass under
  `go test -race`) that pin the construction-time pipe invariant
- documents the invariant on `requestBodyWriter` to keep the allocation out of a
  lazy/first-send path

This is intentionally not a broad connect-go upgrade; it only checks and guards
the runtime safety property called out in the issue.

Fixes #3458

### Checklist

### DIFF
--- a/protocol/triple/triple_protocol/duplex_http_call_test.go
+++ b/protocol/triple/triple_protocol/duplex_http_call_test.go
@@ -1,0 +1,184 @@
+// Copyright 2021-2023 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package triple_protocol
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"sync"
+	"testing"
+)
+
+import (
+	"dubbo.apache.org/dubbo-go/v3/protocol/triple/triple_protocol/internal/assert"
+)
+
+// newTestDuplexClientCall builds a duplexHTTPCall for a client-streaming RPC
+// backed by a server that simply drains the request body and replies 200. It is
+// the minimal setup needed to exercise the Write/CloseWrite/CloseRead lifecycle
+// without pulling in the full codec/stream stack.
+func newTestDuplexClientCall(t *testing.T, client HTTPClient, serverURL *url.URL) *duplexHTTPCall {
+	t.Helper()
+	call := newDuplexHTTPCall(
+		t.Context(),
+		client,
+		serverURL,
+		Spec{StreamType: StreamTypeClient, Procedure: "/triple.test.v1.RaceService/Sum"},
+		http.Header{},
+	)
+	// makeRequest dereferences validateResponse, so it must be set.
+	call.SetValidateResponse(func(*http.Response) *Error { return nil })
+	return call
+}
+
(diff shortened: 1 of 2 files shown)


protocol/triple/triple_protocol/duplex_http_call_test.go
```diff
@@ -0,0 +1,184 @@
+// Copyright 2021-2023 Buf Technologies, Inc.
```
AlexStocks, line 1: [P2] 新文件的 copyright 年份为 `2021-2023`，建议更新为 `2021-2025`（或至少 `2025`）。仓库内部分旧文件沿用旧年份可以理解，但新增文件建议使用当前年份。

AlexStocks (commented):
## Review Summary

**结论: ✅ 可 Merge**

变更范围很小且精准：仅在 `duplexHTTPCall` 结构体上补充了 `requestBodyWriter` 构建时赋值不可变的注释不变量，并新增两个并发 guard test 作为回归保护。

核心判断：
1. `newDuplexHTTPCall` 中 `io.Pipe()` 在构造时一次性分配，`requestBodyWriter` 永不被重新赋值 — 从代码结构上杜绝了 connect-go#919 的 nil-pointer dereference 场景。
2. `d.response` 的写入发生在 `makeRequest()` 中 `close(d.responseReady)` 之前，所有读取方通过 `BlockUntilResponseReady()` 等待 channel 关闭后才访问 — channel close 提供了 Go memory model 的 happens-before 保证，无数据竞争。
3. 两个测试明确定位为 guard（非 reproducer），在正确代码上永远通过；但如果 pipe 创建被移到 lazy path，`assert.NotNil` 会立即失败 — 回归保护设计合理。
4. 测试在 `-race` 下运行可有效检测潜在 data race。

仅有一个 P2 建议，见 inline comment。

PR #3515, opened 2026-07-26
fix(cluster): nil-guard lb.Select results to avoid doSelectInvoker panic

### LINKED ISSUE #3513: [BUG] cluster: nil-pointer panic in doSelectInvoker/adaptivesvc when lb.Select returns nil (P2C/consistenthash)
### Problem

`base.BaseClusterInvoker.doSelectInvoker` calls `selectedInvoker.IsAvailable()` immediately after `lb.Select(invokers, invocation)` without a nil check. Several load-balance implementations can return `nil` for a non-empty invoker list, so the cluster layer panics on a nil-pointer dereference.

Affected `lb.Select` implementations that return `nil`:

- **P2C** (`cluster/loadbalance/p2c/loadbalance.go:128,139`) returns `nil` when `metrics.GetMethodMetrics` returns a non-`ErrMetricsNotFound` error (a transient metrics-layer error). P2C also `panic(...)` outright when the metrics value is not `uint64` (~line 145).
- **ConsistentHash** (`cluster/loadbalance/consistenthashing/loadbalance.go:71`) returns `nil` when `json.Marshal(invoker)` fails.

The `adaptivesvc` fast path forces P2C and has the same unguarded dereference (`cluster/cluster/adaptivesvc/cluster_invoker.go:66-70`).

### Current behavior

`cluster/cluster/base/cluster_invoker.go:160-163`
```go
selectedInvoker := lb.Select(invokers, invocation)

// judge if the selected Invoker is invoked and available
if (!selectedInvoker.IsAvailable() && invoker.AvailableCheck) || isInvoked(selectedInvoker, invoked) {
```

`cluster/cluster/adaptivesvc/cluster_invoker.go:66-70`
```go
lb := extension.GetLoadbalance(lbKey)             // forced to P2C
invoker := lb.Select(invokers, invocation)        // can return nil
...
res := invoker.Invoke(ctx, invocation)            // nil dereference → panic
```

`cluster/loadbalance/p2c/loadbalance.go:128,139,~145`
```go
if err != nil {
    if errors.Is(err, metrics.ErrMetricsNotFound) {
        return invokers[i]
    }
    logger.Warnf(...)
    return nil
}
...
remainingI, ok := remainingIIface.(uint64)
if !ok {
    panic(fmt.Sprintf("[Loadbalance][P2C] type check failed: ..."))
}
```

### Expected behavior

1. `lb.Select` should have a clear contract for non-empty input: either never return `nil`, or callers must handle `nil` by degrading to a fallback selection.
2. `doSelectInvoker` and the `adaptivesvc` fast path must nil-check before dereferencing.
3. P2C must not `panic` on an unexpected metrics type; it should degrade and warn.

### Suggested approach

- Add `if selectedInvoker == nil { /* degrade to first available / random */ }` in `doSelectInvoker` and in the `adaptivesvc` fast path.
- Tighten the `LoadBalance.Select` contract (return `(Invoker, error)`, or guarantee non-nil for non-empty input) and audit all implementations.
- Replace P2C's `panic(...)` on type mismatch with a warning + degraded selection.

### Acceptance criteria

- [ ] P2C transient metrics error no longer causes a panic in the cluster layer; the call degrades to an available invoker.
- [ ] ConsistentHash `json.Marshal` failure no longer causes a nil-deref in `doSelectInvoker`.
- [ ] `adaptivesvc` fast path nil-checks the P2C result before `Invoke`.
- [ ] P2C no longer `panic`s on a non-`uint64` metrics value.
- [ ] Regression tests cover P2C transient error and ConsistentHash marshal failure paths (`-race` clean).

Aias00:
## What

`base.doSelectInvoker` (and the `adaptivesvc` fast path) nil-dereference the result of `lb.Select` when a load-balance implementation returns `nil` for a non-empty invoker list, crashing the process on a transient error.

## Why

Several `LoadBalance` implementations can return `nil` for non-empty input:

- **P2C** (`cluster/loadbalance/p2c/loadbalance.go:128,139`) returns `nil` on a non-`ErrMetricsNotFound` metrics error, and `panic()` outright when the metrics value is not `uint64`.
- **ConsistentHash** (`cluster/loadbalance/consistenthashing/loadbalance.go:71`) returns `nil` when `json.Marshal(invoker)` fails.

`base/cluster_invoker.go:163` then calls `selectedInvoker.IsAvailable()` (and the reselect loop dereferences `reselectedInvoker`) with no nil guard → nil-pointer panic. The `adaptivesvc` fast path (`cluster/cluster/adaptivesvc/cluster_invoker.go:66-70`) forces P2C and dereferences the result the same way.

## Fix

- `base.doSelectInvoker`: if `lb.Select` returns `nil` for the initial selection, warn and degrade to `invokers[0]`; on a `nil` reselect, stop retrying and fall through.
- `adaptivesvc` fast path: degrade to `invokers[0]` when P2C returns `nil`.
- `p2c`: replace the two `panic()` calls on an unexpected metrics type with `logger.Warnf` + `return nil`, so the cluster nil-guard can fall back instead of crashing.

## Tests

Added `TestDoSelectDegradesWhenLbReturnsNil` with a nil-returning `LoadBalance` mock asserting `DoSelect` degrades to an invoker from the list rather than panicking. `p2c`, `base`, `adaptivesvc`, `consistenthashing` packages pass under `-race`.

Note: `TestFailbackRetryFailed` / `TestRouteCacheGenerationRace` fail on `develop` HEAD already (pre-existing, unrelated to this change).

Fixes #3513

### DIFF
--- a/cluster/cluster/adaptivesvc/cluster_invoker.go
+++ b/cluster/cluster/adaptivesvc/cluster_invoker.go
@@ -64,6 +64,12 @@ func (ivk *adaptiveServiceClusterInvoker) Invoke(ctx context.Context, invocation
 
 	// select a node by the loadBalance
 	invoker := lb.Select(invokers, invocation)
+	if invoker == nil {
+		// P2C may return nil on a transient metrics-layer error. Degrade to the
+		// first invoker instead of nil-dereferencing. See #3513.
+		logger.Warnf("[Cluster][AdaptiveSvc] load balance %s returned no invoker, degrade to the first invoker", lbKey)
+		invoker = invokers[0]
+	}
 
 	// invoke
 	invocation.SetAttachment(constant.AdaptiveServiceEnabledKey, constant.AdaptiveServiceIsEnabled)
--- a/cluster/cluster/base/cluster_invoker.go
+++ b/cluster/cluster/base/cluster_invoker.go
@@ -158,6 +158,15 @@ func (invoker *BaseClusterInvoker) doSelectInvoker(lb loadbalance.LoadBalance, i
 	}
 
 	selectedInvoker := lb.Select(invokers, invocation)
+	if selectedInvoker == nil {
+		// A load-balance implementation may return nil on a transient internal
+		// error (e.g. P2C on a metrics-layer error, consistent-hash on a marshal
+		// failure). Degrade to the first invoker instead of nil-dereferencing.
+		// See #3513.
+		logger.Warnf("[Cluster] load balance %T returned nil for serviceKey=%s, degrade to the first invoker",
+			lb, invokers[0].GetURL().ServiceKey())
+		selectedInvoker = invokers[0]
+	}
 
 	// judge if the selected Invoker is invoked and available
 	if (!selectedInvoker.IsAvailable() && invoker.AvailableCheck) || isInvoked(selectedInvoker, invoked) {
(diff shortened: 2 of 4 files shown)


Alanxtl:
not planned

PR #3517, opened 2026-07-26
fix(registry/servicediscovery): lock + snapshot listener map in DataChange/callbacks

### LINKED ISSUE #3512: [BUG] registry/servicediscovery: DataChange/callbacks read listener map unlocked and nil-deref panic (zk/etcdv3/nacos)
### Problem

In the Zookeeper / etcdv3 / Nacos service-discovery backends, the watcher / subscription callback (`DataChange` or the Nacos `SubscribeCallback`) reads `instanceListenerMap[serviceName]` **without holding the lock** that protects it, then iterates `.Values()` on the returned `*gxset.HashSet`.

`instanceListenerMap` is written under a lock (`listenLock` / `listenerLock` / the package-global `initLock`) in `AddListener` / `registerInstanceListener` / `Destroy`, but the callback reads it unlocked. Concurrent subscribe/unsubscribe therefore hits a fatal `fatal error: concurrent map read and map write`.

> Note: `gxset.HashSet.Values()` does not panic on a nil receiver (it ranges a nil map, which is a zero-iteration no-op), so the crash here is the concurrent map access, not a nil dereference. The nil-guard is still added defensively.

### Current behavior

`registry/zookeeper/service_discovery.go` (`DataChange`, ~line 279)
```go
instances := zksd.GetInstances(serviceName)
for _, lis := range zksd.instanceListenerMap[serviceName].Values() {   // no listenLock
    instanceListener := lis.(registry.ServiceInstancesChangedListener)
    err = instanceListener.OnEvent(registry.NewServiceInstancesChangedEvent(serviceName, instances))
}
```

`registry/etcdv3/service_discovery.go` (`DataChange`, ~lines 284-292)
```go
err := jsonutil.DecodeJSON([]byte(eventType.Content), &instance)
if err != nil {
    instance.ServiceName = ""          // decode failure → looks up "" (silent no-op)
}
name := instance.ServiceName
instances := e.GetInstances(name)
for _, lis := range e.instanceListenerMap[instance.ServiceName].Values() {   // no initLock
    instanceLis := lis.(registry.ServiceInstancesChangedListener)
    err = instanceLis.OnEvent(registry.NewServiceInstancesChangedEvent(name, instances))
}
```

`registry/nacos/service_discovery.go` (`SubscribeCallback`, ~line 320)
```go
for _, lis := range n.instanceListenerMap[serviceName].Values() {   // no listenerLock; Destroy can clear the map after this callback is scheduled
    instanceListener := lis.(registry.ServiceInstancesChangedListener)
    e = instanceListener.OnEvent(registry.NewServiceInstancesChangedEvent(serviceName, instances))
}
```

### Expected behavior

1. Reads of `instanceListenerMap` in callbacks must hold the same lock used by writers.
2. `OnEvent` should be dispatched **outside** the lock so a slow listener does not block every other listener for the app.
3. etcdv3 must not blank `ServiceName` to `""` and look it up; a decode failure should drop the event with a warning.

### Suggested approach

- Snapshot the listener set under the writer lock (`listenLock` / `initLock` / `listenerLock`), then iterate the snapshot outside the lock.
- Guard every `m[name]` lookup with `if set, ok := m[name]; ok && set != nil { ... }`.
- etcdv3: on decode error, log and `return false` instead of setting `ServiceName = ""`.
- Add a concurrency test: snapshot vs concurrent map mutation under `-race`.

### Acceptance criteria

- [ ] zk / etcdv3 / nacos callback paths read `instanceListenerMap` under the writer lock and guard nil.
- [ ] `OnEvent` is dispatched outside the lock in all three backends.
- [ ] etcdv3 no longer silently no-ops on an undecodable value; it logs and returns.
- [ ] `Destroy` racing an in-flight callback no longer triggers a fatal map race (`-race` clean).

Aias00:
## What

The Zookeeper / etcdv3 / Nacos service-discovery backends read `instanceListenerMap[serviceName]` inside their watcher/subscription callbacks **without holding the lock** that `AddListener` / `registerInstanceListener` / `Destroy` mutate it under. Concurrent subscribe/unsubscribe therefore triggers a fatal `concurrent map read and map write`.

## Why

- **zookeeper** `registry/zookeeper/service_discovery.go:279` — `DataChange` iterates `zksd.instanceListenerMap[serviceName].Values()` with no `listenLock`, while `AddListener` (`:238`) and (via `Destroy`) mutate the map under `listenLock`.
- **etcdv3** `registry/etcdv3/service_discovery.go:290` — `DataChange` reads `e.instanceListenerMap[...]` with no `initLock`, while `registerServiceInstanceListener`/`registerServiceWatcher` write under `initLock`. Additionally, the decode-error path blanked `ServiceName` to `""` and then looked up `instanceListenerMap[""]` — a silent no-op rather than a real failure.
- **nacos** `registry/nacos/service_discovery.go:320` — the `SubscribeCallback` reads `n.instanceListenerMap[serviceName]` with no `listenerLock`, while `registerInstanceListener` and `Destroy` mutate it under `listenerLock`; an in-flight callback delivered after `Destroy` clears the map is a live race.

(`gxset.HashSet.Values()` does not panic on a nil receiver, so the issue is the concurrent map access, not a nil deref.)

## Fix

Snapshot the listener set under the writer lock, then iterate the snapshot **outside** the lock to dispatch — this both eliminates the race and stops a slow `OnEvent` from blocking every other listener:

- **zookeeper** `DataChange`: snapshot under `listenLock` via a new `snapshotListeners` helper; dispatch outside the lock.
- **etcdv3** `DataChange`: snapshot under `initLock`; on decode error, log and `return false` instead of blanking `ServiceName`.
- **nacos** `SubscribeCallback`: snapshot under `listenerLock` before `OnEvent`.

## Tests

Added `TestZookeeperServiceDiscovery_SnapshotListeners` covering: missing key (safe no-op), present key (returns registered listeners), and concurrent snapshot-vs-mutation race-freedom under `-race`. `registry/zookeeper`, `registry/etcdv3`, `registry/nacos` all pass under `-race`.

Fixes #3512

### DIFF
--- a/registry/etcdv3/service_discovery.go
+++ b/registry/etcdv3/service_discovery.go
@@ -277,26 +277,41 @@ func (e *etcdV3ServiceDiscovery) registerServiceWatcher(serviceName string) erro
 
 // DataChange when child data change should DispatchEventByServiceName
 func (e *etcdV3ServiceDiscovery) DataChange(eventType remoting.Event) bool {
-	if eventType.Action == remoting.EventTypeUpdate {
-		instance := &registry.DefaultServiceInstance{}
-		err := jsonutil.DecodeJSON([]byte(eventType.Content), &instance)
-		if err != nil {
-			instance.ServiceName = ""
-		}
-
-		// notify instance listener instance change
-		name := instance.ServiceName
-		instances := e.GetInstances(name)
-		for _, lis := range e.instanceListenerMap[instance.ServiceName].Values() {
-
-			instanceLis := lis.(registry.ServiceInstancesChangedListener)
-			err = instanceLis.OnEvent(registry.NewServiceInstancesChangedEvent(name, instances))
-		}
-		if err != nil {
-			return false
-		}
+	if eventType.Action != remoting.EventTypeUpdate {
+		return true
+	}
+	instance := &registry.DefaultServiceInstance{}
+	if err := jsonutil.DecodeJSON([]byte(eventType.Content), &instance); err != nil {
+		// Previously the decode failure blanked ServiceName to "" and then looked
+		// up instanceListenerMap[""], a silent no-op. Fail fast and log instead.
+		logger.Errorf("[Registry][Etcdv3] decode instance failed, content=%s err=%v", eventType.Content, err)
+		return false
 	}
 
+	// notify instance listener instance change
+	name := instance.ServiceName
+	instances := e.GetInstances(name)
+	// Snapshot the listener set under initLock (the writers
+	// registerServiceInstanceListener/registerServiceWatcher use the same lock)
+	// and dispatch outside the lock: an unlocked read races the writers and
+	// holding the lock across OnEvent would block other listeners. See #3512.
+	initLock.Lock()
+	set, ok := e.instanceListenerMap[name]
+	listeners := make([]registry.ServiceInstancesChangedListener, 0)
+	if ok && set != nil {
+		for _, lis := range set.Values() {
+			listeners = append(listeners, lis.(registry.ServiceInstancesChangedListener))
+		}
+	}
+	initLock.Unlock()
+
+	var err error
+	for _, lis := range listeners {
+		err = lis.OnEvent(registry.NewServiceInstancesChangedEvent(name, instances))
+	}
+	if err != nil {
+		return false
+	}
 	return true
 }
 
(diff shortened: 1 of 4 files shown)


Alanxtl:
not planned

PR #3519, opened 2026-07-26
fix(registry/servicediscovery): guard ServiceMappingChangedListenerImpl.Stop with mux

### LINKED ISSUE #3518: [BUG] servicediscovery: ServiceMappingChangedListenerImpl.Stop writes `stop` outside the mutex OnEvent reads it under
### Problem

`ServiceMappingChangedListenerImpl.OnEvent` reads the `stop` field under `lstn.mux` (the listener's own mutex), but `Stop` writes `stop` **without** that mutex:

```go
// registry/servicediscovery/service_mapping_change_listener_impl.go
func (lstn *ServiceMappingChangedListenerImpl) OnEvent(e observer.Event) error {
    lstn.mux.Lock()
    defer lstn.mux.Unlock()
    if lstn.stop == ServiceMappingListenerStop {   // read under mux
        return nil
    }
    ...
}

func (lstn *ServiceMappingChangedListenerImpl) Stop() {
    lstn.stop = ServiceMappingListenerStop            // write WITHOUT mux
}
```

`Stop` is called from `serviceDiscoveryRegistry.stopListen` under `s.lock` (the registry lock), not `lstn.mux`. So `OnEvent` (reading `stop` under `mux`) races `Stop` (writing `stop` without `mux`) — a data race on the `int` field, and a window where a stopped listener still processes a mapping event (spurious `SubscribeURL`/listener creation after teardown). `go test -race` flags it.

### Current behavior

- `Stop()` writes `lstn.stop` with no synchronization.
- `OnEvent()` reads `lstn.stop` under `lstn.mux`.
- Concurrent `UnSubscribe` (→ `stopListen` → `Stop`) and a `ServiceMappingChangedEvent` delivered to `OnEvent` race on `stop`.

### Expected behavior

`stop` should be read and written under the same mutex so that once `Stop` returns, no in-flight `OnEvent` proceeds past the stop check, and the field access is race-free.

### Suggested approach

- Have `Stop` acquire `lstn.mux` before writing `stop` (the same mutex `OnEvent` reads `stop` under).
- Verify `Stop` is never called re-entrantly from within `OnEvent`'s held-`mux` call stack (it is not — `Stop` is reached via `UnSubscribe`/`stopListen`, independent of the `OnEvent` → `SubscribeURL` path), so there is no self-deadlock.

### Acceptance criteria

- [ ] `Stop` writes `stop` under `lstn.mux`.
- [ ] A concurrency test (concurrent `Stop` vs `OnEvent`) passes under `-race`.
- [ ] After `Stop` returns, `OnEvent` returns early without processing the event.

Aias00:
## What

`ServiceMappingChangedListenerImpl.Stop` writes the `stop` field **without** `lstn.mux`, while `OnEvent` reads `stop` **under** `lstn.mux`. The unlocked write races `OnEvent` (data race on the `int` field) and lets a mapping event be processed after the listener has been stopped.

## Why

```go
// registry/servicediscovery/service_mapping_change_listener_impl.go
func (lstn *ServiceMappingChangedListenerImpl) OnEvent(e observer.Event) error {
    lstn.mux.Lock()
    defer lstn.mux.Unlock()
    if lstn.stop == ServiceMappingListenerStop { ... }   // read under mux
    ...
}

func (lstn *ServiceMappingChangedListenerImpl) Stop() {
    lstn.stop = ServiceMappingListenerStop                 // write WITHOUT mux
}
```

`Stop` is reached from `serviceDiscoveryRegistry.stopListen` (under the registry's `s.lock`, not `lstn.mux`), so a concurrent `UnSubscribe` and a `ServiceMappingChangedEvent` delivered to `OnEvent` race on `stop`. `go test -race` flags it.

## Fix

Acquire `lstn.mux` in `Stop` before writing `stop`. `Stop` is not called re-entrantly from `OnEvent`'s held-`mux` call stack (it is reached via `UnSubscribe`/`stopListen`, independent of the `OnEvent` → `SubscribeURL` path), so there is no self-deadlock.

## Tests

Added `TestServiceMappingChangedListener_StopRace`: 100 rounds of concurrent `Stop()` vs `OnEvent(nil)` (which acquires `mux`, reads `stop`, early-returns), passing under `-race`.

Fixes #3518

### DIFF
--- a/registry/servicediscovery/service_mapping_change_listener_impl.go
+++ b/registry/servicediscovery/service_mapping_change_listener_impl.go
@@ -114,5 +114,10 @@ func (lstn *ServiceMappingChangedListenerImpl) OnEvent(e observer.Event) error {
 
 // Stop on ServiceMappingChangedEvent the service mapping change event
 func (lstn *ServiceMappingChangedListenerImpl) Stop() {
+	// Guard with the same mux that OnEvent reads `stop` under. The previous
+	// unlocked write raced OnEvent (which reads `stop` under mux) and let a
+	// mapping event be processed after the listener had been stopped. See #3518.
+	lstn.mux.Lock()
+	defer lstn.mux.Unlock()
 	lstn.stop = ServiceMappingListenerStop
 }
--- a/registry/servicediscovery/service_mapping_change_listener_impl_test.go
+++ b/registry/servicediscovery/service_mapping_change_listener_impl_test.go
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package servicediscovery
+
+import (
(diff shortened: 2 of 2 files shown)


Alanxtl:
not planned

PR #3521, opened 2026-07-26
fix(cluster/directory/static): guard invokers with a dedicated RWMutex

### LINKED ISSUE #3520: [BUG] cluster/directory/static: List/IsAvailable read invokers unlocked, racing Destroy
### Problem

`cluster/directory/static.directory` reads the `invokers` slice header (and backing array) in `List` and `IsAvailable` **without any lock**, while `Destroy` reassigns `dir.invokers = []protocolbase.Invoker{}` under `base.Directory.mutex` (via `DoDestroy`). This is a data race on the slice header (ptr/len/cap) per the Go memory model: a concurrent `Destroy` (provider destroy cascade) and an in-flight `List`/`IsAvailable` (e.g. `available` cluster invoker calls `List` before `CheckWhetherDestroyed`) trip `go test -race`.

`base.Directory.mutex` is unexported, so the `static` package cannot acquire it directly; the static directory needs its own lock for `invokers`.

### Current behavior

```go
// cluster/directory/static/directory.go
type directory struct {
    *base.Directory
    invokers []protocolbase.Invoker   // no dedicated lock
}

func (dir *directory) IsAvailable() bool {
    ...
    if len(dir.invokers) == 0 { ... }      // unlocked read of slice header
    for _, invoker := range dir.invokers { ... }   // unlocked range
}

func (dir *directory) List(...) []protocolbase.Invoker {
    l := len(dir.invokers)                 // unlocked read
    invokers := make(...)
    copy(invokers, dir.invokers)           // unlocked read of header + backing array
    ...
}

func (dir *directory) Destroy() {
    dir.DoDestroy(func() {
        for _, ivk := range dir.invokers { ivk.Destroy() }
        dir.invokers = []protocolbase.Invoker{}   // write under base.Directory.mutex
    })
}
```

### Expected behavior

`invokers` reads in `List`/`IsAvailable` and the write in `Destroy` should be guarded by the same lock, so destroy never races an in-flight read.

### Suggested approach

- Add a `sync.RWMutex` to the static `directory` (since `base.Directory.mutex` is not accessible cross-package).
- `IsAvailable`/`List` snapshot `invokers` under `RLock` and iterate the snapshot outside the lock.
- `List` releases the lock before calling `RouterChain()` (which takes `base.Directory.mutex`) so the two locks never nest and there is no deadlock.
- `Destroy`'s closure writes `invokers` under the new lock.

### Acceptance criteria

- [ ] `List`/`IsAvailable` read `invokers` under the directory lock.
- [ ] `Destroy` writes `invokers` under the same lock.
- [ ] A concurrency test (concurrent `List`/`IsAvailable` vs `Destroy`) passes under `-race`.
- [ ] No deadlock between the directory lock and `base.Directory.mutex`.

Aias00:
## What

`List` and `IsAvailable` read `dir.invokers` without a lock, while `Destroy` reassigns it under `base.Directory.mutex` (via `DoDestroy`). The unlocked reads race the write (data race on the slice header).

## Why

```go
// cluster/directory/static/directory.go
func (dir *directory) IsAvailable() bool {
    ...
    if len(dir.invokers) == 0 { ... }                 // unlocked read
    for _, invoker := range dir.invokers { ... }      // unlocked range
}
func (dir *directory) List(...) []protocolbase.Invoker {
    l := len(dir.invokers); ... copy(invokers, dir.invokers)  // unlocked read
}
func (dir *directory) Destroy() {
    dir.DoDestroy(func() { ...; dir.invokers = []protocolbase.Invoker{} })  // write under base.Directory.mutex
}
```

`base.Directory.mutex` is unexported and not accessible from the `static` package, so the static directory needs its own lock for `invokers`.

## Fix

Add a `sync.RWMutex` to the static `directory`:

- `IsAvailable`/`List` snapshot `invokers` under `RLock` and iterate the snapshot.
- `List` releases the lock **before** calling `RouterChain()` (which takes `base.Directory.mutex`) so the two locks never nest — no deadlock.
- `Destroy`'s closure writes `invokers` under the new lock (inside `DoDestroy`, which holds `base.Directory.mutex`; no other path holds the new lock while waiting for `base.Directory.mutex`).

## Tests

Added `TestStaticDirListVsDestroyRace`: 50 rounds of concurrent `List`/`IsAvailable`/`Destroy`, passing under `-race`.

Fixes #3520

### DIFF
--- a/cluster/directory/static/directory.go
+++ b/cluster/directory/static/directory.go
@@ -18,6 +18,10 @@
 package static
 
 import (
+	"sync"
+)
+
+import (
 	"github.com/dubbogo/gost/log/logger"
 
 	perrors "github.com/pkg/errors"
@@ -33,6 +37,10 @@ import (
 type directory struct {
 	*base.Directory
 	invokers []protocolbase.Invoker
+	// lock guards invokers against the race between List/IsAvailable (read) and
+	// Destroy (write). base.Directory.mutex is unexported and not accessible
+	// from this package. See #3520.
+	lock sync.RWMutex
 }
 
 // NewDirectory Create a new staticDirectory with invokers
@@ -62,10 +70,16 @@ func (dir *directory) IsAvailable() bool {
 		return false
 	}
 
-	if len(dir.invokers) == 0 {
+	// Snapshot invokers under lock: Destroy reassigns the slice header (under
+	// lock below), so an unlocked read races it. See #3520.
+	dir.lock.RLock()
+	invokers := dir.invokers
+	dir.lock.RUnlock()
+
+	if len(invokers) == 0 {
 		return false
 	}
-	for _, invoker := range dir.invokers {
+	for _, invoker := range invokers {
 		if !invoker.IsAvailable() {
 			return false
 		}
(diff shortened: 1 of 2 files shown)


Alanxtl:
not planned

PR #3523, opened 2026-07-26
fix(p2c/adaptivesvc): read singleton instance only through sync.Once

### LINKED ISSUE #3522: [BUG] p2c/adaptivesvc: singleton `instance` read outside sync.Once races first-time init
### Problem

Both `cluster/loadbalance/p2c` and `cluster/cluster/adaptivesvc` initialize a package-level singleton with the same buggy pattern: an `if instance == nil` check **outside** `sync.Once.Do`, then `return instance` also outside `Do`.

```go
var (
    once     sync.Once
    instance loadbalance.LoadBalance
)

func newP2CLoadBalance(r randomPicker) loadbalance.LoadBalance {
    if r == nil { r = defaultRnd }
    if instance == nil {              // unsynchronized read
        once.Do(func() { instance = &p2cLoadBalance{...} })
    }
    return instance                  // unsynchronized read when the check was false
}
```

`sync.Once.Do` only establishes a happens-before for callers that **call `Do`**. A goroutine that observes `instance != nil` (check false) skips `Do` entirely and reads `instance` with no happens-before to the goroutine that wrote it inside `Do`. Two concurrent first-time callers therefore race on `instance`: `go test -race` flags it. The identical pattern exists in `cluster/cluster/adaptivesvc/cluster.go:newAdaptiveServiceCluster`.

Because both factories are registered in `init()` via `extension.SetLoadbalance`/`SetCluster`, every concurrent RPC at process start (before first use) hits this path.

### Current behavior

- `cluster/loadbalance/p2c/loadbalance.go` `newP2CLoadBalance` reads `instance` outside `once.Do`.
- `cluster/cluster/adaptivesvc/cluster.go` `newAdaptiveServiceCluster` reads `instance` outside `once.Do`.

### Expected behavior

The read of `instance` must have a happens-before relation to the write, so concurrent first-time callers cannot race.

### Suggested approach

- Drop the `if instance == nil` fast path; always call `once.Do(...)` then `return instance`. `sync.Once.Do` guarantees all callers see the writes done inside `Do`, and there is no read of `instance` outside `Do`.
- The `r == nil` defaulting in p2c moves inside the `once.Do` closure (only the first call's `r` is used, identical to before).

### Acceptance criteria

- [ ] No read of `instance` outside `once.Do` in either package.
- [ ] A concurrency test (100 goroutines racing the factory) passes under `-race`.

Aias00:
## What

`newP2CLoadBalance` and `newAdaptiveServiceCluster` read the package-level `instance` **outside** `sync.Once.Do` (an `if instance == nil` fast path + `return instance`), racing first-time concurrent initialization.

## Why

```go
var (
    once     sync.Once
    instance loadbalance.LoadBalance
)

func newP2CLoadBalance(r randomPicker) loadbalance.LoadBalance {
    if r == nil { r = defaultRnd }
    if instance == nil {              // unsynchronized read
        once.Do(func() { instance = &p2cLoadBalance{...} })
    }
    return instance                  // unsynchronized read when check was false
}
```

`sync.Once.Do` only establishes happens-before for callers that **call `Do`**. A goroutine that observes `instance != nil` (check false) skips `Do` and reads `instance` with no happens-before to the goroutine that wrote it inside `Do`. Two concurrent first-time callers race on `instance`. The identical pattern is in `cluster/cluster/adaptivesvc/cluster.go:newAdaptiveServiceCluster`. Both factories are registered in `init()`, so concurrent RPCs at process start hit this path. `go test -race` flags it.

## Fix

Drop the `if instance == nil` fast path; always `once.Do(...)` then `return instance`. `sync.Once.Do` guarantees all callers see the writes done inside `Do`, and there is no read of `instance` outside `Do`. The p2c `r == nil` defaulting moves inside the `once.Do` closure (only the first call's `r` is used, identical to before).

## Tests

Added `TestNewP2CLoadBalanceConcurrent`: 100 goroutines racing the factory, passing under `-race`. `p2c` package passes under `-race`.

Fixes #3522

### DIFF
--- a/cluster/cluster/adaptivesvc/cluster.go
+++ b/cluster/cluster/adaptivesvc/cluster.go
@@ -42,11 +42,12 @@ func init() {
 type adaptiveServiceCluster struct{}
 
 func newAdaptiveServiceCluster() clusterpkg.Cluster {
-	if instance == nil {
-		once.Do(func() {
-			instance = &adaptiveServiceCluster{}
-		})
-	}
+	// Always go through once.Do so the read of `instance` has a happens-before
+	// relation to the write inside Do. The previous `if instance == nil` fast
+	// path read `instance` without synchronization, racing the first-time write.
+	once.Do(func() {
+		instance = &adaptiveServiceCluster{}
+	})
 	return instance
 }
 
--- a/cluster/loadbalance/p2c/loadbalance.go
+++ b/cluster/loadbalance/p2c/loadbalance.go
@@ -90,14 +90,15 @@ func defaultRnd(n int) (i, j int) {
 // randomPicker parameter is designed ONLY FOR TEST purposes.
 // Thread-safe via sync.Once.
 func newP2CLoadBalance(r randomPicker) loadbalance.LoadBalance {
-	if r == nil {
-		r = defaultRnd
-	}
-	if instance == nil {
-		once.Do(func() {
-			instance = &p2cLoadBalance{randomPicker: r}
-		})
-	}
+	// Always go through once.Do so the read of `instance` has a happens-before
+	// relation to the write inside Do. The previous `if instance == nil` fast
+	// path read `instance` without synchronization, racing the first-time write.
+	once.Do(func() {
+		if r == nil {
+			r = defaultRnd
+		}
+		instance = &p2cLoadBalance{randomPicker: r}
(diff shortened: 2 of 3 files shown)


Alanxtl:
not planned

PR #3525, opened 2026-07-26
fix(hessian2): guard argsTypes.(string) in unpackRequestBody

### LINKED ISSUE #3524: [BUG] hessian2: unpackRequestBody panics on unchecked argsTypes.(string) for null/non-string descriptor
### Problem

`unpackRequestBody` decodes the Dubbo Hessian2 request body field by field. The 5th field (`argsTypes`, the argument-types descriptor, e.g. `"Ljava/lang/String;"`) is decoded and then used in an **unchecked** type assertion:

```go
// protocol/dubbo/hessian2/hessian_request.go:226-232
argsTypes, err = decoder.Decode()
if err != nil {
    return perrors.WithStack(err)
}
req[4] = argsTypes

ats := DescRegex.FindAllString(argsTypes.(string), -1)   // unchecked assertion
```

`decoder.Decode()` can return `(nil, nil)` (a Hessian null marker decodes to nil with no error) or any non-string value for a malformed/hostile request. `argsTypes.(string)` then panics, crashing the server-side goroutine handling the request (the Dubbo protocol listener), dropping the connection.

### Current behavior

1. A Dubbo Hessian2 request whose argument-types field is Hessian `null` (or any non-string) is decoded to `(nil, nil)`.
2. `argsTypes.(string)` panics.
3. The server goroutine handling that request crashes (connection drop / `recover`).

### Expected behavior

A malformed argument-types descriptor should return a clear protocol error instead of panicking; an empty string (zero args) must keep working.

### Suggested approach

- Use a comma-ok type assertion: `argsTypesStr, ok := argsTypes.(string); if !ok { return error }`.
- `DescRegex.FindAllString(argsTypesStr, -1)` on the validated string.

### Acceptance criteria

- [ ] A request whose argsTypes field is Hessian null no longer panics; `unpackRequestBody` returns a descriptive error.
- [ ] A request with an empty-string argsTypes (zero args) still decodes successfully.
- [ ] Existing `TestPackRequest` / request-decoding tests remain green.
- [ ] Regression test covers the null-argsTypes path.

Aias00:
## What

`unpackRequestBody` panics on an unchecked `argsTypes.(string)` assertion when the argument-types descriptor field of a Dubbo Hessian2 request is a Hessian `null` (or any non-string), crashing the server-side goroutine handling the request.

## Why

```go
// protocol/dubbo/hessian2/hessian_request.go:226-232
argsTypes, err = decoder.Decode()
if err != nil {
    return perrors.WithStack(err)
}
req[4] = argsTypes

ats := DescRegex.FindAllString(argsTypes.(string), -1)   // unchecked assertion
```

`decoder.Decode()` returns `(nil, nil)` for a Hessian null marker (no error), so a malformed/hostile request whose argsTypes field is null reaches the assertion and panics — dropping the connection / tripping `recover`.

## Fix

Comma-ok assertion; on a non-string argsTypes, return a descriptive error instead of panicking. An empty-string argsTypes (zero args) still decodes unchanged.

```go
argsTypesStr, ok := argsTypes.(string)
if !ok {
    return perrors.Errorf("hessian2: invalid argument types descriptor, expected string, got %T", argsTypes)
}
ats := DescRegex.FindAllString(argsTypesStr, -1)
```

## Tests

Added `TestUnpackRequestBody_NilArgsTypes`: encodes a request whose argsTypes field is a Hessian null and asserts `unpackRequestBody` returns an error (no panic). `protocol/dubbo/hessian2` package passes.

Fixes #3524

### DIFF
--- a/protocol/dubbo/hessian2/hessian_request.go
+++ b/protocol/dubbo/hessian2/hessian_request.go
@@ -229,7 +229,14 @@ func unpackRequestBody(decoder *hessian.Decoder, reqObj any) error {
 	}
 	req[4] = argsTypes
 
-	ats := DescRegex.FindAllString(argsTypes.(string), -1)
+	// argsTypes must be a string descriptor (e.g. "Ljava/lang/String;"). A null
+	// or non-string value (malformed/hostile request) would otherwise panic on
+	// the type assertion. Surface it as an error instead. See #3523.
+	argsTypesStr, ok := argsTypes.(string)
+	if !ok {
+		return perrors.Errorf("hessian2: invalid argument types descriptor, expected string, got %T", argsTypes)
+	}
+	ats := DescRegex.FindAllString(argsTypesStr, -1)
 	var arg any
 	for range ats {
 		arg, err = decoder.Decode()
--- a/protocol/dubbo/hessian2/hessian_request_test.go
+++ b/protocol/dubbo/hessian2/hessian_request_test.go
@@ -158,3 +158,22 @@ func TestIssue192(t *testing.T) {
 		})
 	}
 }
+
+// TestUnpackRequestBody_NilArgsTypes verifies that a request whose
+// argument-types descriptor field is a Hessian null no longer panics on the
+// unchecked argsTypes.(string) assertion; instead it returns an error.
+// Regression for #3523.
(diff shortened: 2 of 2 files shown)


Alanxtl:
not planned

PR #3540, opened 2026-07-26
fix(config_center/zookeeper): fix RemoveListener key mismatch + guard listener set

### LINKED ISSUE #3539: [BUG] config_center/zookeeper: RemoveListener no-op (key mismatch leak) + inner listener map race
### Problem

Two bugs in the Zookeeper config_center listener:

1. **`RemoveListener` uses a different key than `AddListener` — listener leak.** `AddListener` qualifies the key (`namespace/key` → `buildPath(rootPath, ...)`) before storing under the qualified key, but `RemoveListener` passes the **raw** caller key, so `CacheListener.RemoveListener` does `keyListeners.Load(rawKey)` which never matches the stored qualified key → `loaded` is false → the listener is never deleted. Every remove is a silent no-op; stale listeners accumulate and keep receiving `DataChange` after unsubscribe.

   ```go
   // config_center/zookeeper/impl.go
   func (c *zookeeperDynamicConfiguration) AddListener(key string, listener ..., ...) {
       key = strings.Join([]string{c.GetURL().GetParam(constant.ConfigNamespaceKey, ...), key}, "/")
       qualifiedKey := buildPath(c.rootPath, key)
       c.cacheListener.AddListener(qualifiedKey, listener)   // stored qualified
   }
   func (c *zookeeperDynamicConfiguration) RemoveListener(key string, listener ..., ...) {
       c.cacheListener.RemoveListener(key, listener)           // looked up RAW -> no-op
   }
   ```

2. **`CacheListener` inner listener map has no lock — fatal concurrent map read+write.** `keyListeners` is a `sync.Map` whose **value** is a plain `map[ConfigurationListener]struct{}` with no mutex (unlike `metadata/report/zookeeper`'s locked `ListenerSet`). `AddListener`/`RemoveListener` (router/config goroutines) write/delete the inner map; `DataChange` (zk event goroutine) ranges it. Concurrent map read+write is a fatal Go runtime error (`fatal error: concurrent map read and map write`), not just a benign race.

### Current behavior

- Router `RemoveListener` calls (condition/script/affinity routers all call `dynamicConfiguration.RemoveListener(key, ...)`) are no-ops → leaked stale listeners keep firing `Process` against destroyed routers.
- A config push arriving while a router subscribes/unsubscribes crashes the process with a fatal concurrent-map-access error.

### Expected behavior

1. `RemoveListener` must qualify the key identically to `AddListener` so the stored entry is actually removed.
2. The inner listener set must be guarded by a mutex (or replaced with a locked set), and `DataChange` must not hold that lock across the external `Process` callback.

### Suggested approach

- Add a `qualifyKey(key)` helper used by **both** `AddListener` and `RemoveListener` so the two keys cannot drift.
- Replace the inner `map[ConfigurationListener]struct{}` value with a mutex-guarded `listenerSet` (`add`/`remove`/`snapshot`); `DataChange` snapshots under the lock and dispatches `Process` outside the lock.

### Acceptance criteria

- [ ] `RemoveListener` removes the listener that `AddListener` stored (same qualified key).
- [ ] The inner listener set is mutex-guarded; no fatal concurrent-map-access under `-race`.
- [ ] `DataChange` dispatches `Process` outside the set lock.
- [ ] Regression tests: `qualifyKey` equality between Add/Remove; `listenerSet` concurrency under `-race`.

Aias00:
## What

Two bugs in the Zookeeper config_center listener: `RemoveListener` was a silent no-op (key mismatch → leak), and the inner listener map had no lock (fatal concurrent map read+write).

## Why

**1. Key mismatch (leak).** `AddListener` stored under the qualified key (`buildPath(rootPath, namespace/key)`), but `RemoveListener` passed the **raw** caller key, so `CacheListener.RemoveListener` did `keyListeners.Load(rawKey)` → never matched → listener never deleted. Every router `RemoveListener` call (condition/script/affinity routers) was a no-op; stale listeners kept receiving `DataChange` after unsubscribe.

```go
// config_center/zookeeper/impl.go (before)
func (c *zookeeperDynamicConfiguration) AddListener(key string, ...) {
    key = strings.Join([]string{c.GetURL().GetParam(constant.ConfigNamespaceKey, ...), key}, "/")
    qualifiedKey := buildPath(c.rootPath, key)
    c.cacheListener.AddListener(qualifiedKey, listener)   // stored qualified
}
func (c *zookeeperDynamicConfiguration) RemoveListener(key string, ...) {
    c.cacheListener.RemoveListener(key, listener)           // looked up RAW -> no-op
}
```

**2. Inner map race (fatal).** `keyListeners` is a `sync.Map` whose value is a plain `map[ConfigurationListener]struct{}` with no mutex (unlike `metadata/report/zookeeper`'s locked `ListenerSet`). `AddListener`/`RemoveListener` (router goroutines) write/delete the inner map; `DataChange` (zk event goroutine) ranges it → `fatal error: concurrent map read and map write`.

## Fix

- Add a `qualifyKey(key)` helper used by **both** `AddListener` and `RemoveListener` so the keys cannot drift.
- Replace the inner `map` value with a mutex-guarded `listenerSet` (`add`/`remove`/`snapshot`); `DataChange` snapshots under the lock and dispatches `Process` outside it.

## Tests

- `TestZookeeperDynamicConfigurationQualifyKey`: `qualifyKey` produces the same key for Add/Remove (with and without an explicit namespace).
- `TestListenerSetConcurrency`: 100 rounds of concurrent `add`/`remove`/`snapshot` under `-race`.
- Updated existing `CacheListener` tests for the new `listenerSet` value type.

`config_center/zookeeper` passes under `-race`.

Fixes #3539

### DIFF
--- a/config_center/zookeeper/listener.go
+++ b/config_center/zookeeper/listener.go
@@ -33,7 +33,7 @@ import (
 
 // CacheListener defines keyListeners and rootPath
 type CacheListener struct {
-	// key is zkNode Path and value is set of listeners
+	// key is zkNode Path and value is *listenerSet (mutex-guarded set of listeners)
 	keyListeners    sync.Map
 	zkEventListener *zookeeper.ZkEventListener
 	rootPath        string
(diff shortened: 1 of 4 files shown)


config_center/zookeeper/listener.go
```diff
@@ -44,27 +44,58 @@ func NewCacheListener(rootPath string, listener *zookeeper.ZkEventListener) *Cac
...
+	actual, _ := l.keyListeners.LoadOrStore(key, newListenerSet())
+	actual.(*listenerSet).add(listener)
 }
 
 // RemoveListener will delete a listener if loaded
 func (l *CacheListener) RemoveListener(key string, listener config_center.ConfigurationListener) {
-	listeners, loaded := l.keyListeners.Load(key)
-	if loaded {
-		delete(listeners.(map[config_center.ConfigurationListener]struct{}), listener)
+	if listeners, ok := l.keyListeners.Load(key); ok {
```
zouyx, line 97: 是不是 Load 的动作和 Remove 动作在同一个锁内会更合适？

PR #3580, opened 2026-07-30
fix: preserve wrapped error handling

### LINKED ISSUE #3560: [Bug] Improper error comparison/wrapping breaks errors.Is/As (== io.EOF, %v instead of %w) (P1/P2)
## Summary

Several places compare errors with `==` against sentinel errors, or wrap errors with `%v` instead of `%w`. Once an error is wrapped (or comes from a wrapped source), `==` silently fails and the error chain is broken, so error-chain-based handling (`errors.Is`/`errors.As`) no longer works.

## Affected locations

| File:Line | Code | Problem |
|-----------|------|---------|
| `protocol/jsonrpc/server.go:329` | `if err == io.EOF \|\| err == io.ErrUnexpectedEOF { ... }` | Sentinel comparison fails if `err` is wrapped. Use `errors.Is(err, io.EOF)`. |
| `protocol/jsonrpc/server.go:127` | `err == io.EOF` (bufio.Peek) | Same — prefer `errors.Is`. |
| `protocol/triple/reflection/serverreflection.go:169` | `err == io.EOF` (grpc stream) | Same — prefer `errors.Is`. |
| `otel/trace/exporter.go:77` | `err = fmt.Errorf("failed to create %s exporter: %v", config.Exporter, err)` | Uses `%v`, breaking the error chain; downstream `errors.Is`/`As` can't unwrap it. Should be `%w`. |

Note: the core transport paths already use `errors.Is` correctly (e.g. `dual_transport.go`, `error.go` compare `context.DeadlineExceeded`/`context.Canceled` with `errors.Is`), so this is an inconsistency rather than a pervasive bug.

## Impact

- A wrapped `io.EOF`/cancelled error bypasses the sentinel branches → incorrect handling or silent mismatch.
- `%v` wrapping in the OTel exporter hides the root cause from `errors.Is`/`As` consumers.

## Suggested fix

- Replace `err == io.EOF` / `== io.ErrUnexpectedEOF` with `errors.Is(err, ...)`.
- Replace `%v` with `%w` when wrapping errors that should remain inspectable.

## Verification

`GOTOOLCHAIN=local go vet ./...` on develop tip (HEAD 53d81d17) reports **zero** warnings (see #3552). `go vet` does not flag `==` vs `errors.Is` or `%v` vs `%w` — needs the `errorlint` linter (proposed in the enhanced golangci config) to catch these.

nightcityblade:
### Description
Fixes #3560

Use `errors.Is` for EOF sentinel checks so wrapped errors follow the existing control flow, and wrap exporter creation errors with `%w` so callers can inspect their cause.

Validation:
- `go test ./protocol/jsonrpc ./protocol/triple/reflection ./otel/trace`
- `go vet ./protocol/jsonrpc ./protocol/triple/reflection ./otel/trace`
- `golangci-lint run ./protocol/jsonrpc ./protocol/triple/reflection ./otel/trace --timeout=10m`

### Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works (N/A: this change updates standard-library error matching/wrapping behavior; the affected package suites and `errorlint` pass)

### DIFF
--- a/otel/trace/exporter.go
+++ b/otel/trace/exporter.go
@@ -74,7 +74,7 @@ func NewExporter(config *ExporterConfig, customFunc func() (sdktrace.SpanExporte
 
 	exporter, err := customFunc()
 	if err != nil {
-		err = fmt.Errorf("failed to create %s exporter: %v", config.Exporter, err)
+		err = fmt.Errorf("failed to create %s exporter: %w", config.Exporter, err)
 		logger.Errorf("[OTel][Trace] failed to create %s exporter, err=%v", config.Exporter, err)
 		return
 	}
--- a/protocol/triple/reflection/serverreflection.go
+++ b/protocol/triple/reflection/serverreflection.go
@@ -20,6 +20,7 @@ package reflection
 
 import (
 	"context"
+	"errors"
 	"io"
 	"slices"
 	"sort"
@@ -166,7 +167,7 @@ func (s *ReflectionServer) ServerReflectionInfo(ctx context.Context, stream rpb.
 	sentFileDescriptors := make(map[string]bool)
 	for {
 		in, err := stream.Recv()
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
 			return nil
 		}
 		if err != nil {
--- a/otel/trace/exporter_test.go
+++ b/otel/trace/exporter_test.go
@@ -76,8 +76,9 @@ func TestNewExporter_CustomFuncError(t *testing.T) {
 		ServiceName: "test-service",
 	}
 
+	cause := errors.New("custom func error")
 	customFunc := func() (sdktrace.SpanExporter, error) {
-		return nil, errors.New("custom func error")
+		return nil, cause
 	}
 
 	tracerProvider, propagator, err := NewExporter(config, customFunc)
@@ -85,6 +86,7 @@ func TestNewExporter_CustomFuncError(t *testing.T) {
 	assert.Nil(t, propagator)
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "failed to create test exporter")
+	require.ErrorIs(t, err, cause)
 }
 
 func TestNewExporter_InvalidSampleMode(t *testing.T) {
--- a/protocol/jsonrpc/server.go
+++ b/protocol/jsonrpc/server.go
@@ -21,6 +21,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"io"
 	"mime"
 	"net"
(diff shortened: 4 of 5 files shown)


Alanxtl (approved):
lgtm

otel/trace/exporter.go
```diff
@@ -74,7 +74,7 @@ func NewExporter(config *ExporterConfig, customFunc func() (sdktrace.SpanExporte
 
 	exporter, err := customFunc()
 	if err != nil {
-		err = fmt.Errorf("failed to create %s exporter: %v", config.Exporter, err)
+		err = fmt.Errorf("failed to create %s exporter: %w", config.Exporter, err)
```
AlexStocks, line 77: [P1] 补上错误链回归断言

现有 `TestNewExporter_CustomFuncError` 只检查错误字符串；我在 Base 的 `%v` 实现上单独运行它仍然通过，所以它无法防止本次修复回退。使用同一 cause 调用 `NewExporter` 的 Base/Head 探针中，`errors.Is(err, cause)` 在 Base 失败、当前 Head 通过。请在现有测试中保留 cause 并使用 `require.ErrorIs` 断言原始错误仍位于链上；需要覆盖 `errors.As` 时再加入自定义错误类型。

protocol/triple/reflection/serverreflection.go
```diff
@@ -166,7 +167,7 @@ func (s *ReflectionServer) ServerReflectionInfo(ctx context.Context, stream rpb.
 	sentFileDescriptors := make(map[string]bool)
 	for {
 		in, err := stream.Recv()
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
```
AlexStocks, line 170: [P1] 覆盖 wrapped EOF 的实际完成路径

这个包当前没有任何测试，但真实调用链会产生 wrapped EOF：handler connection 的 Receive 错误经过 `wrapIfUncoded` 后成为实现 `Unwrap` 的 `*triple_protocol.Error`。使用 `Recv()` 返回 `fmt.Errorf("transport: %w", io.EOF)` 的 fake stream 做 Base/Head 区分时，Base 把它作为错误返回，当前 Head 才正常返回 nil。请增加回归测试，至少断言 wrapped EOF 正常结束，并断言 wrapped 非 EOF 仍原样传播；否则把 `errors.Is` 退回 `==` 时现有 CI 仍会全绿。

otel/trace/exporter.go
nightcityblade: Addressed in 3806e50. The test now keeps the original cause and verifies it with require.ErrorIs.

protocol/triple/reflection/serverreflection.go
nightcityblade: Addressed in 3806e50. Added a fake-stream regression test covering both wrapped EOF completion and propagation of a non-EOF receive error. The two affected packages pass go test, and targeted golangci-lint reports 0 issues.

1kasa (approved):
lgtm

AlexStocks (approved):

PR #3599, opened 2026-08-06
fix: make lifecycle waits cancellable

### LINKED ISSUE #3555: [Bug] Goroutine/context leaks: failback retry ticker, registry retry loop, getty connect loop (P0/P1)
## Description

Several goroutines and retry loops have no cancellation / context-exit path, so they can leak or block indefinitely when the component is shutting down or the upstream is unavailable.

### 1. `cluster/cluster/failback/cluster_invoker.go` — `process` (L94-L121)

```go
func (invoker *failbackClusterInvoker) process(ctx context.Context) {
    invoker.ticker = time.NewTicker(time.Second * 1)
    for range invoker.ticker.C {            // L96: no ctx.Done() branch
        for {
            value, err := invoker.taskList.Peek()
            if err == queue.ErrDisposed { return }
            ...
            go invoker.tryTimerTaskProc(ctx, retryTask)   // L118
        }
    }
}
```

- `ctx` is passed in but never selected on; the only exit is `taskList.Dispose()`.
- The ticker has no `defer invoker.ticker.Stop()`.
- Risk: if `Dispose` is never called, the goroutine (and the ticker) leak for the process lifetime.

### 2. `registry/base_registry.go` — `Subscribe` (L330-L347)

```go
for {                                    // L330: no ctx exit
    if !r.IsAvailable() { return ... }
    listener, err := r.facadeBasedRegistry.DoSubscribe(url)
    if err != nil { time.Sleep(...); continue }  // L343: unbounded retry, no ctx cancel
    for { ... }                         // L347: inner loop also has no ctx
}
```

- No upper bound on retries; cannot be cancelled by an external `context`.
- Risk: a permanently unavailable registry causes an infinite blocking retry loop.

### 3. `remoting/getty/pool.go` — connect loop (L81-L97)

```go
for {
    idx++
    if c.isAvailable() { break }
    if time.Since(start) > connectTimeout { ... return }
    interval := time.Millisecond * time.Duration(idx)
    time.Sleep(interval)                // L96: blocks; cannot react to ctx cancel
}
```

- Uses `time.Sleep` polling instead of `select` + `ctx.Done()`; a cancelled context still waits for the sleep to return.
- Risk: connection attempts cannot be interrupted promptly.

## Suggested fix

- `failback`: `defer invoker.ticker.Stop()`; add `select { case <-ctx.Done(): return; case <-invoker.ticker.C: ... }`.
- `base_registry`: select on `ctx.Done()` in the retry loops and return on cancellation.
- `getty/pool`: replace the `time.Sleep` polling with `select { case <-ctx.Done(): return; case <-time.After(interval): }`.

## Severity

P0/P1 — goroutine leak + non-cancellable retry loops affect long-running service stability.

## Environment

- Reproducible on current `develop` tip (3.3.2 prep).

DadaVinqi:
## Summary

This PR fixes three shutdown and cancellation issues tracked by #3555:

- Failback retry workers may outlive the invoker lifecycle.
- `BaseRegistry.Subscribe` cannot be interrupted while waiting between retries.
- Getty connection establishment may block `Close()` until the full connection timeout.

No public API signatures are changed.

Fixes #3555

## Problem

During shutdown, several background loops do not respond promptly to the component lifecycle:

1. Failback retry processing can continue after the invoker has been destroyed.
2. Registry subscription retries use an uninterruptible `time.Sleep`, so `Destroy()` may leave `Subscribe()` blocked for the full retry delay.
3. Getty connection establishment waits in a polling loop while holding a read lock required by `Client.Close()`. As a result, `Close()` cannot publish the closed state until connection establishment finishes.

These behaviors can delay graceful shutdown, leave retry work running against destroyed components, and retain resources longer than expected.

### Failback

Failback retries were not tied to an invoker-owned lifecycle context. The retry processor and retry goroutines did not have a common cancellation path coordinated by `Destroy()`, and enqueueing new retry tasks was not fully guarded after destruction.

### BaseRegistry

`BaseRegistry.Subscribe` used:

```go
time.Sleep(time.Duration(RegistryConnDelay) * time.Second)
```

The sleep could not observe `BaseRegistry.Done()`. In addition, the subscription loop did not check registry availability after `listener.Next()` returned, so an event could still be delivered during a shutdown race.

### Getty

`selectSession` held the client read lock while calling `newGettyRPCClientConn`, including its connection wait loop. `Client.Close()` requires the corresponding write lock, so it could not mark the client closed while connection establishment was in progress.

The connection wait loop also used `time.Sleep`, which could not be interrupted immediately by client shutdown.

### Failback lifecycle cancellation

- Added a retry lifecycle context owned by the failback invoker.
- Created the context with `context.WithCancel(context.WithoutCancel(...))` so retries retain the intended lifecycle behavior while remaining cancellable by the invoker.
- Added cancellation checks to the retry processor and retry execution path.
- Made `Destroy()` idempotent.
- `Destroy()` now:
  - Marks the invoker stopped.
  - Cancels the retry context.
  - Disposes the retry queue.
  - Waits for the retry processor and active retry tasks with a bounded timeout.
  - Delegates to the underlying invoker destroy logic.
- Prevented new retry tasks from being enqueued after destruction.
- Prevented invocation and retry processing from continuing after shutdown.

### Interruptible Registry subscription retry

- Replaced the retry `time.Sleep` with a timer selecting on:
  - The retry timer.
  - `BaseRegistry.Done()`.
- Added availability checks before and after `listener.Next()`.
- Close the listener and return the existing unavailable error when destruction is observed.
- Kept the public `Subscribe` API unchanged.
- Did not modify provider-specific listener implementations.

### Getty connection shutdown

- Replaced the mutable `clientClosed` flag with an atomic closed state.
- Added a client shutdown channel that is closed immediately by `Client.Close()`.
- Added a dedicated connection mutex to serialize slow connection creation.
- Restricted `gettyClientMux` to protecting the published Getty connection pointer.
- Removed the long-held client read lock around connection establishment.
- Added double checks before and after connection creation so a connection cannot be published after shutdown.
- Close an unpublished connection if shutdown occurs during creation.
- Changed connection reset logic to clear only the expected connection, preventing an old connection from clearing a newer one.
- Replaced the Getty polling sleep with an interruptible timer wait.
- Preserved the existing connection timeout behavior.

## Regression Tests

Added coverage for:

- Failback cancellation during retry execution.
- Idempotent failback destruction.
- Bounded failback shutdown waiting.
- Preventing retry enqueue after destruction.
- Registry retry cancellation during `Destroy()`.
- Preventing registry notifications after destruction.
- Getty connection wait cancellation.
- Getty timeout behavior.
- `Client.Close()` not waiting for the connection mutex.
- Closing a Getty connection creation attempt in progress.

## Test Plan

- [x] `make test`
- [x] `make lint`
- [x] `make fmt`
- [x] `go test -count=1 ./cluster/... ./registry/... ./remoting/...`
- [x] `go test -count=1 -race ./registry ./cluster/cluster/failback`
- [x] Focused Getty shutdown tests with `-race`
- [x] `go vet`
- [x] `git diff --check`

### DIFF
--- a/remoting/getty/getty_client_test.go
+++ b/remoting/getty/getty_client_test.go
@@ -20,6 +20,7 @@ package getty
 import (
 	"bytes"
 	"context"
+	"net"
 	"reflect"
 	"sync"
 	"testing"
@@ -330,4 +331,91 @@ func TestInitClientTLS(t *testing.T) {
 		initClient(url)
 		assert.False(t, clientConf.SSLEnabled)
 	})
+}
+
+func TestGettyConnectWaitStopsWhenClosed(t *testing.T) {
+	client := NewClient(Options{ConnectTimeout: 5 * time.Second})
+	started := make(chan struct{})
+	var startOnce sync.Once
+	available := func() bool {
+		startOnce.Do(func() { close(started) })
+		return false
+	}
+	waitDone := make(chan error, 1)
+	go func() {
+		waitDone <- waitForGettyClient("127.0.0.1:1", client.opts.ConnectTimeout, available, client.done)
+	}()
+	select {
+	case <-started:
+	case <-time.After(time.Second):
+		t.Fatal("connection wait did not start")
+	}
+
+	start := time.Now()
+	client.Close()
+	err := <-waitDone
+
+	require.Error(t, err)
+	require.ErrorIs(t, err, errClientClosed)
+	require.Less(t, time.Since(start), time.Second)
+}
+
+func TestGettyConnectWaitHonorsTimeout(t *testing.T) {
+	start := time.Now()
+	err := waitForGettyClient("127.0.0.1:1", 30*time.Millisecond,
+		func() bool { return false },
+		nil,
+	)
+
+	require.Error(t, err)
+	require.NotErrorIs(t, err, errClientClosed)
+	require.Less(t, time.Since(start), time.Second)
+}
+
+func TestGettyNewConnectionStopsWhenClientCloses(t *testing.T) {
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	addr := listener.Addr().String()
+	require.NoError(t, listener.Close())
+
+	client := NewClient(Options{ConnectTimeout: 5 * time.Second})
+	client.conf = *GetDefaultClientConfig()
+	connectDone := make(chan error, 1)
+	go func() {
+		_, connectErr := newGettyRPCClientConn(client, addr)
+		connectDone <- connectErr
+	}()
+	time.AfterFunc(20*time.Millisecond, client.Close)
+
+	start := time.Now()
+	select {
+	case err := <-connectDone:
+		require.Error(t, err)
+		require.ErrorIs(t, err, errClientClosed)
+	case <-time.After(time.Second):
+		t.Fatal("newGettyRPCClientConn remained blocked after Close")
+	}
+	require.Less(t, time.Since(start), time.Second)
+}
+
+func TestClientCloseDoesNotWaitForConnectLock(t *testing.T) {
+	client := NewClient(Options{ConnectTimeout: time.Second, RequestTimeout: time.Second})
+	client.connectMu.Lock()
+	closeDone := make(chan struct{})
+	go func() {
+		client.Close()
+		close(closeDone)
+	}()
+
+	select {
+	case <-closeDone:
+	case <-time.After(time.Second):
+		t.Fatal("Client.Close waited for the connection lock")
+	}
+	client.connectMu.Unlock()
+
+	require.True(t, client.closed.Load())
+	_, _, err := client.selectSession("")
+	require.Error(t, err)
+	require.ErrorIs(t, err, errClientClosed)
 }
--- a/remoting/getty/getty_client_test.go
+++ b/remoting/getty/getty_client_test.go
@@ -28,6 +28,8 @@ import (
 )
 
 import (
+	dubboGetty "github.com/apache/dubbo-getty"
+
 	hessian "github.com/apache/dubbo-go-hessian2"
 
 	perrors "github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


remoting/getty/getty_client_test.go
```diff
@@ -331,3 +332,90 @@ func TestInitClientTLS(t *testing.T) {
...
+		func() bool { return false },
+		nil,
+	)
+
+	require.Error(t, err)
+	require.NotErrorIs(t, err, errClientClosed)
+	require.Less(t, time.Since(start), time.Second)
+}
+
+func TestGettyNewConnectionStopsWhenClientCloses(t *testing.T) {
```
AlexStocks, line 431: [P1] 补上连接 ready 后、发布前 Close 的回归测试

当前测试只在 newGettyRPCClientConn 仍等待连接时关闭 client，没有经过 getOrCreateGettyClient 创建完成后的两次 closed 门禁。我在 exact Head mutant 中删除这两个检查后，4 个新增关闭测试以 -race -count=20 仍全部通过；该实现可在 Close 已消费 closeOnce 后再发布连接，连接将无法被后续 Close 回收。建议通过可控 connection factory/同步点把创建暂停在 ready 与 publish 之间，先完成 Close，再断言创建返回 errClientClosed、未发布连接被关闭且 c.gettyClient 仍为 nil。

DadaVinqi: 做了修改：新增 `TestGettyCloseAfterConnectionReadyBeforePublish`，用同步 factory 覆盖连接 ready 后、发布前执行 `Close()` 的竞态，并断言返回 `errClientClosed`、未发布连接被关闭且共享 `gettyClient` 保持 nil。同时将连接发布前的关闭检查收敛到 `gettyClientMux` 锁内，避免慢速建连期间持有指针锁。已通过 `go test -race -count=20 ./remoting/getty -run TestGettyCloseAfterConnectionReadyBeforePublish`、`make test` 和 `make lint`。修复提交：`ecac3321`。

AlexStocks (approved):

DadaVinqi:
感谢 @AlexStocks 的 review 和 approval。

PR #3599 当前所有 checks 已通过，状态为 MERGEABLE/CLEAN，代码侧没有待处理修改。维护者方便时，麻烦帮忙合并一下，谢谢！

PR #3622, opened 2026-08-08
fix(protocol/triple): handle ParseDuration error for keepalive config

### LINKED ISSUE #3621: protocol/triple: ParseDuration errors silently discarded in Dubbo3Invoker keepalive config
## Description

In `protocol/triple/dubbo3_invoker.go:114-116`, the `time.ParseDuration` errors for `KeepAliveInterval` and `KeepAliveTimeout` are silently discarded:

```go
// TODO: handle ParseDuration error
keepAliveInterval, _ = time.ParseDuration(tripleConf.KeepAliveInterval)
keepAliveTimeout, _ = time.ParseDuration(tripleConf.KeepAliveTimeout)
```

This means that if a user configures an invalid duration string (e.g., "30sec" instead of "30s"), the value silently becomes 0 with no warning or error. This can lead to unexpected keepalive behavior that is difficult to debug.

### Expected Behavior
When ParseDuration fails, the code should:

1. Log a warning with the invalid value
2. Fall back to the default keepalive settings derived from URL parameters

### Affected Code

protocol/triple/dubbo3_invoker.go:111-117

### Reproduction

Configure a Triple service with an invalid keepalive duration:

```go
triple:
  keep-alive-interval: "30sec"  # invalid, should be "30s"
  keep-alive-timeout: "10x"     # invalid, should be "10s"
```

yashisrani:
### Description

Fixes a bug in `Dubbo3Invoker` where `time.ParseDuration` errors for `KeepAliveInterval` and `KeepAliveTimeout` configuration were silently discarded. Invalid duration strings (e.g., `"30sec"`) would silently zero out the keepalive values, leading to hard-to-debug connection behavior.

Now the invoker checks the parse error, logs a warning with the invalid value, and falls back to the URL-param-derived default when parsing fails.

Fixes #3621

### Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/protocol/triple/dubbo3_invoker.go
+++ b/protocol/triple/dubbo3_invoker.go
@@ -111,9 +111,16 @@ func NewDubbo3Invoker(url *common.URL) (*DubboInvoker, error) {
 	tripleConfRaw, ok := url.GetAttribute(constant.TripleConfigKey)
 	if ok {
 		tripleConf := tripleConfRaw.(*global.TripleConfig)
-		// TODO: handle ParseDuration error
-		keepAliveInterval, _ = time.ParseDuration(tripleConf.KeepAliveInterval)
-		keepAliveTimeout, _ = time.ParseDuration(tripleConf.KeepAliveTimeout)
+		if parsedInterval, err := time.ParseDuration(tripleConf.KeepAliveInterval); err == nil {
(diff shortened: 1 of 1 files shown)


Alanxtl (approved):
lgtm

PR #3625, opened 2026-08-09
fix: self-heal application-level discovery after transient metadata failure (#3615)

AsperforMias:
### Description

Fixes #3615

**Problem.** With application-level service discovery, a single transient `MetadataService` fetch failure during a provider restart could leave the consumer directory permanently empty (`No provider available`) until the consumer was restarted. Root cause chain, confirmed by a deterministic docker reproduction (details in [this comment](https://github.com/apache/dubbo-go/issues/3615#issuecomment-5226228151)):

1. `ServiceInstancesChangedListenerImpl.OnEvent` skipped instances whose metadata fetch failed, overwrote the service URLs with the (possibly empty) result, and returned success. No retry queue, no error propagation — and since Nacos only pushes on instance-list *changes*, no later event ever retried the fetch.
2. Even when a later push did re-fetch metadata successfully, `RegistryDirectory` refused to rebuild the invoker: the graceful shutdown of the provider had left a **closing tombstone** on the instance key (default TTL 30s), and a genuine same-address restart was indistinguishable from a stale pre-shutdown registry snapshot.

**Fix (two layers, both required — verified by e2e):**

- **Metadata retry** (`service_instances_changed_listener_impl.go`): failed revisions are tracked in `unresolvedRevisions` and re-resolved by replaying the latest instance snapshot through the existing build path, using one shared timer per listener with capped exponential backoff (1s→30s, 25% jitter, **unlimited attempts** — a retry cap would re-introduce the permanent failure this fixes). Retries stop naturally when instances leave the snapshot, when the last subscriber is removed, or on registry `Destroy()` (`service_discovery_registry.go`). Metadata RPCs no longer run under the listener state mutex (build is serialized by a separate `buildMu`; the state mutex only guards field access), so a slow/dead provider cannot block event processing.
- **Tombstone restart discrimination** (`directory.go`): the closing tombstone now records the closing invoker's export `timestamp`. A re-add carrying the *same* timestamp is a stale pre-shutdown snapshot and stays vetoed (existing graceful-shutdown semantics unchanged); a *different* timestamp proves a genuine restart, so the tombstone is cleared and the invoker rebuilt immediately. URLs without a timestamp (older providers) keep the conservative veto.

[...]

PR #3634, opened 2026-08-10
fix: retry failed application-level AddListener with capped backoff (#3624)

### LINKED ISSUE #3624: [BUG] Application-level service discovery: failed AddListener is never retried, consumer permanently misses instance events
## ✅ Verification Checklist

- [x] 🔍 I have searched the [existing issues](https://github.com/apache/dubbo-go/issues) and confirmed this is not a duplicate

---

### 🚀 Go Version

go1.26.1

### 📦 Dubbo-go Version

v3.3.2 (also present on `main` @ `48d6e696`)

### 🌐 Protocol Configuration

Triple

### 📋 Registry Configuration

Nacos v2.5.1, application-level service discovery (`service-discovery-registry://...`)

### 💾 Operating System

🐧 Linux

---

### 📝 Bug Description

In application-level service discovery, installing the Nacos instance listener is fire-and-forget: `serviceDiscoveryRegistry.subscribeAndNotify` runs `serviceDiscovery.AddListener(listener)` in a goroutine, and if it fails the error is only logged once:

https://github.com/apache/dubbo-go/blob/v3.3.2/registry/servicediscovery/service_discovery_registry.go

```go
go func() {
    err = s.serviceDiscovery.AddListener(listener)
    ...
    if err != nil {
        logger.Errorf("[Registry][ServiceDiscovery] add instance listener catch error, ...")
    }
}()
```

There is no retry, and `SubscribeURL` does not report the failure. From that moment the consumer never receives any instance-change push for those applications: the initial `GetInstances` snapshot has already been processed, so the current state looks fine, but every subsequent provider scale up/down or restart is invisible to this consumer permanently (until restart).

Note that the interface-level Nacos registry already has exponential-backoff subscribe retry (#3178, `registry/nacos/registry.go`), but the application-level path (`service-discovery-registry://`) does not use it. This is the same "one-shot attempt, failure swallowed, permanent divergence" pattern as #3615 (metadata fetch without retry), just at the subscription-channel layer.

Impact: a transient Nacos error at subscribe time (reconnect storm, leader election, brief unavailability) silently turns into a permanently stale consumer. Combined with #3615's masking scenario, it can affect a single consumer pod while others stay healthy, making it look like sporadic failures.

### 🔄 Steps to Reproduce

With a real Nacos this needs a transient subscribe failure at exactly the subscribe moment (e.g. Nacos leader re-election); the deterministic variant is a unit test:

1. Wrap/stub `ServiceDiscovery.AddListener` so the first call returns an error and later calls succeed (the failure point is the async `s.serviceDiscovery.AddListener(listener)` call in `subscribeAndNotify`, `registry/servicediscovery/service_discovery_registry.go`).
2. Subscribe an application-level reference; the initial `GetInstances` succeeds, so the directory is built and calls work.
3. Trigger any provider change (restart/scale). The consumer receives no event because the Nacos subscribe was never established.
4. No further `AddListener` attempt happens; the consumer stays stale until restart. The only trace is one `add instance listener catch error` log line.

A minimal regression test asserting "first AddListener fails, second succeeds, consumer eventually receives instance events without restart" currently fails on v3.3.2 because no retry exists.

### ✅ Expected Behavior

A failed application-level `AddListener` is retried with backoff (with jitter, cancellation on unsubscribe/destroy, and dedup per listener) until it succeeds, similar to the interface-level retry added in #3178. While the subscription is not established, this state should be observable (log/metric, see #3356).

### ❌ Actual Behavior

The failure is logged once (`add instance listener catch error`) and never retried; the consumer silently misses all future instance events.

### 💡 Possible Solution

- Reuse the backoff/dedup/cancel machinery introduced for the metadata-fetch retry fix for #3615 (single timer per registry, exponential backoff with jitter and a capped interval, unlimited attempts while the subscriber is still active).
- Cancel pending retries on `UnSubscribe`/`Destroy`.
- Expose the "subscription not established" state via the observability work tracked in #3356 (e.g. a gauge/counter), so a single stuck consumer pod can be alerted on.
- Regression test: first `AddListener` fails, second succeeds; assert the consumer eventually receives instance events without a restart. Also assert retries stop after `UnSubscribe`/`Destroy`.

Related: #3615 (metadata fetch retry), #3356 (observability), #3178 (interface-level subscribe backoff).

AsperforMias:
### Description
Fixes #3624

**Root cause.** In application-level service discovery, `serviceDiscoveryRegistry.subscribeAndNotify` runs `serviceDiscovery.AddListener(listener)` fire-and-forget in a goroutine. A transient registry error (reconnect storm, leader election, brief unavailability) was logged once and never retried. Since the initial `GetInstances` snapshot has already been processed, the consumer looks healthy but silently misses every subsequent instance-change push until restart — a single pod can go permanently stale while others stay fine.

**Fix.** Mirror the retry discipline of the interface-level nacos subscribe backoff (#3178), at the subscription-channel layer:

- One pending retry per `serviceNamesKey` (i.e. per listener): repeated failing subscribes share a single timer instead of stacking.
- Exponential backoff 1s → 30s cap + up to 25% jitter; attempts are intentionally unlimited while subscribers remain (a capped count would re-introduce the permanent staleness this fixes).
- A successful retry **re-syncs the latest instance snapshot** via `GetInstances`/`OnEvent`, so instance changes missed while the subscription was down are picked up instead of relying on a future push.
- Pending retries are canceled when the last subscriber unsubscribes and on `Destroy`; no new timer is armed once the registry is destroyed or no subscriber remains.
- Observability: initial failure keeps the existing error log; each failed retry logs one warning with the attempt count and error; a successful recovery logs once; every attempt publishes the existing `NewServerSubscribeEvent` metric. A dedicated "subscription not established" gauge is left to #3356.

Two latent bugs in the touched paths are fixed along the way:

- `UnSubscribe` removed the notify listener by `url.ServiceKey()` while `SubscribeURL` had registered it under `ServiceKey()+":"+protocol`, so the entry was never removed and subscriber-count-based cleanup could never fire. Both sides now use the same `protocolServiceKeyOf` key.
- The `event.Succ = err != nil` flag in the subscribe path was inverted (and a `%s`/`%d` log mismatch next to it).

**Verification.**

- Unit/regression tests (`registry/servicediscovery/subscribe_retry_test.go`, run with `-race`): first `AddListener` fails → retry establishes the subscription → the consumer receives instance events without a restart; backoff stays in the `[delay, delay+25%]` band and is capped; retries stop after `UnSubscribe` and after `Destroy`, and no timer is armed post-`Destroy`; repeated failing subscribes share one timer.
- Real-Nacos e2e (nacos 2.5.1, docker compose, application-level discovery): the consumer wraps the real nacos `ServiceDiscovery` so the first `AddListener` fails (the deterministic variant suggested in the issue). On `upstream/main`: baseline calls work from the initial snapshot, the provider is then replaced, and the consumer stays at `No provider available` for 60s+ until restarted — bug reproduced. On this branch: `established after 1 retries, re-syncing latest instances`, the consumer follows the provider replacement and calls keep succeeding **without a restart**.

Note: this adds a small `metadataInfoFetcher` indirection for test fault injection; #3625 introduces the same variable, so whichever PR lands second keeps one copy (trivial conflict).

### Checklist


registry/servicediscovery/service_discovery_registry.go
```diff
@@ -576,28 +582,185 @@ func (s *serviceDiscoveryRegistry) getServiceListener(serviceNamesKey string) re
...
 	logger.Infof("[Registry][ServiceDiscovery] start subscribing to registry for applications=%s with a new go routine", serviceNamesKey)
 	go func() {
-		err := s.serviceDiscovery.AddListener(listener)
-		event.Succ = err != nil
-		event.End = time.Now()
-		event.Attachment[constant.InterfaceKey] = url.Interface()
-		metrics.Publish(event)
-		metrics.Publish(metricsRegistry.NewServerSubscribeEvent(err == nil))
-		if err != nil {
+		if err := s.addInstanceListener(url, listener); err != nil {
```
AlexStocks: [P1] 后续成功的 `AddListener` 不会撤销同一 `serviceNamesKey` 已挂起的重试。首个订阅失败后，如果 timer 触发前另一个 `SubscribeURL` 成功，这个 goroutine 的成功分支什么都不做，旧 timer 仍会再次调用 `AddListener`；当前 Head 的公开入口探针得到 2 次必要调用后仍出现第 3 次。Polaris 的 `AddSubscriber` 和 ZooKeeper 的 `ListenServiceEvent` 会重复注册回调/监听，不能依赖幂等。成功时应在确认 listener/state 匹配后取消 pending retry，并补“先失败、后续订阅成功、等待超过 backoff 仍只有两次调用”的测试。

registry/servicediscovery/service_discovery_registry.go
```diff
@@ -576,28 +582,185 @@ func (s *serviceDiscoveryRegistry) getServiceListener(serviceNamesKey string) re
...
+	active := !s.destroyed &&
+		s.serviceListeners[serviceNamesKey] == state.listener &&
+		listenerHasSubscribers(state.listener)
+	s.lock.Unlock()
+
+	if !active {
+		// Registry destroyed, listener replaced, or no subscriber left:
+		// stop retrying so the loop cannot leak.
+		return
+	}
```
Alanxtl, line 833: retry state is deleted while the external retry attempt is still in flight


deletes `subscribeRetries[serviceNamesKey]` before calling `AddListener` at line 729. During that call, another `SubscribeURL` for the same key can fail and schedule a fresh `subscribeRetry{attempts: 0}` because the map is temporarily empty. If the in-flight retry then also fails, its higher-attempt state is dropped by `scheduleSubscribeRetry` because the new low-attempt timer already exists. That breaks the PR’s “one pending retry per key” and sustained exponential backoff guarantees, and can cause repeated low-delay retry bursts during reconnect storms. Keeping the retry state present with an “in flight” marker, or replacing only when preserving/maxing attempts, would close this window.

AsperforMias: done

registry/servicediscovery/service_discovery_registry.go
AsperforMias: done

Alanxtl (approved):
lgtm

Alanxtl:
pls fix conflict

AsperforMias:
done

Alanxtl:
pls fix conflict after merging #3625

AlexStocks (approved):

PR #3658, opened 2026-08-13
fix(router): preserve v3.0 condition flags

### LINKED ISSUE #3656: Condition router v3.0 swaps enabled and force flags
### Environment

- dubbo-go branch: `develop`
- affected code: `cluster/router/condition/generateConditionsRoute`
- config center: ZooKeeper

### Description

The v3.0 condition-router parser assigns the two boolean fields in reverse order:

```go
force, enable := *routerConfig.Enabled, *routerConfig.Force
```

As a result, a normal rule with `enabled: true` and `force: false` is interpreted as `enable=false` and `force=true`. `DynamicRouter.Route` then bypasses the condition router and returns the original invoker list.

The v3.1 parser is not affected because `generateMultiConditionRoute` reads and returns the fields in the correct order.

### Reproduction

Store the following service-level rule in the config center:

```yaml
configVersion: v3.0
priority: 1
enabled: true
force: false
runtime: true
key: org.apache.dubbo.quickstart.Greeter:1.0.0:demo
scope: service
conditions:
  - method=SayHello => application=router-rule-e2e-condition-provider-hangzhou
```

Use two providers whose final consumer-side invoker URLs contain different `application` values. Calls continue to select both providers because the v3.0 router is treated as disabled.

Changing the same rule to v3.1 enters the separate multi-destination parser and routes normally.

### Expected behavior

- `enabled` controls whether the router executes.
- `force` controls whether an empty routing result falls back to the original invokers.
- Updating the same config-center node from v3.1 to v3.0 should replace the router state and apply the v3.0 condition immediately.

### Proposed fix

Assign the fields in their declared order and add regression coverage for the asymmetric `enabled: true` / `force: false` combination, including a v3.1-to-v3.0 runtime update.

xiaobaicai66695:
### What this PR does

Fixes the v3.0 condition-router parser assigning `enabled` and `force` in reverse order.

Before this change, the common combination below disabled the router at runtime:

```yaml
enabled: true
force: false
```

The parser interpreted it as `enable=false` and `force=true`, so `DynamicRouter.Route` returned the original invoker list without applying the condition.

### Changes

- assign `RouterConfig.Force` to `force` and `RouterConfig.Enabled` to `enable`;
- add regression coverage for asymmetric v3.0 boolean flags;
- verify that a runtime update from v3.1 to v3.0 replaces the active router and applies the v3.0 destination condition.

### Testing

```text
go test ./cluster/router/condition -count=1
ok dubbo.apache.org/dubbo-go/v3/cluster/router/condition
```

Fixes #3656

### DIFF
--- a/cluster/router/condition/dynamic_router.go
+++ b/cluster/router/condition/dynamic_router.go
@@ -293,7 +293,7 @@ func generateConditionsRoute(rawConfig string) (stateRouters, bool, bool, error)
 		return nil, false, false, err
 	}
 
-	force, enable := *routerConfig.Enabled, *routerConfig.Force
+	force, enable := *routerConfig.Force, *routerConfig.Enabled
 	if !enable {
 		return nil, false, false, nil
 	}
--- a/cluster/router/condition/router_test.go
+++ b/cluster/router/condition/router_test.go
@@ -187,6 +187,74 @@ func TestDynamicRouterSetStaticConfig(t *testing.T) {
 	})
 }
 
+func TestGenerateConditionsRoutePreservesExplicitBooleans(t *testing.T) {
+	raw := `conditions:
+- method=SayHello => application=router-rule-e2e-condition-provider-hangzhou
+configVersion: v3.0
+enabled: true
+force: false
+key: org.apache.dubbo.quickstart.Greeter:1.0.0:demo
+priority: 1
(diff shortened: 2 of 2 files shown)


Alanxtl (approved):
lgtm

PR #3660, opened 2026-08-14
fix(configurator): align v3 rule version and empty match

### LINKED ISSUE #3661: [Bug]: v3 dynamic configurator reads the wrong version key and cannot handle empty match
## What happened?

A standard v3 dynamic configurator rule cannot reliably execute the v3 matching path.

Expected:

- The rule-level YAML field `configVersion: v3.0` selects `configureIfMatchV3`.
- An omitted match or `match: {}` means match-all.
- A partial match evaluates only the dimensions explicitly configured.
- Matching parameters such as `loadbalance` are applied to the target URL.

Actual:

- `DefaultConfigurationParser` writes the rule version to the generated URL as `configVersion`.
- `overrideConfigurator` reads `constant.ConfigVersionKey`, whose value is `config-center.configVersion` and belongs to the config-center component parameter namespace.
- The version lookup therefore returns empty and the standard v3 rule falls into `configureDeprecated`, bypassing v3 match semantics. Simple rules may still appear to work through the deprecated path, producing a false-positive integration result.
- When the v3 path is reached, an omitted or empty `ConditionMatch` contains nil pointers. `ConditionMatch.IsMatch` dereferences `Address`, `ProviderAddress`, `Service`, and `App` unconditionally, so match-all and partial matches can panic before parameters are applied.

## Configuration contract evidence

The rule contract consistently uses `configVersion`:

- Dubbo Admin traffic-rule protobufs, API models, codecs, YAML editors, and validation use `configVersion`.
- Dubbo-Go `ConfiguratorConfig` maps its `ConfigVersion` field to the YAML key `configVersion`.
- `DefaultConfigurationParser` serializes the parsed value into configurator URLs as `configVersion`.
- Condition, Tag, Script, and Affinity rule configurations use the same rule-level spelling.
- Existing parser and configurator tests also construct or assert URLs with `configVersion`.

No equivalent traffic-rule model, codec, YAML document, parser, or Admin API emits `config-center.configVersion`. In the affected rule path, that spelling appears only through `constant.ConfigVersionKey` and the configurator lookup that reused it.

The namespaced value belongs to a different domain. It was introduced when config-center component parameters such as `namespace`, `timeout`, and `username` were renamed to `config-center.namespace`, `config-center.timeout`, and `config-center.username`. `ConfigVersionKey` was namespaced in the same batch. The dynamic configurator continued reusing that component constant even though its rule schema remained `configVersion`.

## Affected chain and nodes

For an application-scoped rule, Admin stores the YAML at a configuration-center node such as:

```text
/dubbo/config/dubbo/<application>.configurators
```

The E2E node is:

```text
/dubbo/config/dubbo/router-rule-e2e-dynamic-consumer.configurators
```

The complete affected chain is:

```text
Admin Dynamic Config form/YAML
  -> Console API request with configVersion
  -> ZooKeeper <rule-key>.configurators node containing configVersion: v3.0
  -> dubbo-go configuration listener
  -> DefaultConfigurationParser.ParseToUrls
  -> appItemToUrls/serviceItemToUrls creates an override URL with configVersion=v3.0
  -> overrideConfigurator.Configure reads the version and selects a matching path
  -> configureIfMatchV3
  -> ConditionMatch.IsMatch
  -> parameters are copied to the target consumer/provider URL
  -> directory/load balancer uses the resulting URL
  -> provider invocation behavior changes
```

Admin persistence and the ZooKeeper node are correct. The first failure occurs after the node is read: `overrideConfigurator.Configure` looks up `config-center.configVersion`, misses the parser-generated `configVersion`, and selects the deprecated path. Correcting that lookup exposes the second failure at `ConditionMatch.IsMatch`, before parameters can be copied to the target URL.

Consequently, an Admin response, a correct ZooKeeper node, and even a configuration-change listener event are not sufficient evidence that the v3 rule is active. The final URL parameters and invocation behavior must also be checked.

## How can we reproduce it?

Use a v3 application-scoped dynamic configuration such as:

```yaml
configVersion: v3.0
enabled: true
key: demo-consumer
scope: application
configs:
  - side: consumer
    match: {}
    enabled: true
    parameters:
      loadbalance: roundrobin
```

1. Save the rule through Dubbo Admin and confirm the `<application>.configurators` ZooKeeper node contains the YAML above.
2. Parse the YAML with `DefaultConfigurationParser.ParseToUrls`.
3. Observe that the generated configurator URL contains `configVersion=v3.0`.
4. Pass that URL to `overrideConfigurator.Configure`.
5. The configurator looks for `config-center.configVersion`, does not recognize the v3 rule, and enters the deprecated path.
6. If version dispatch is corrected independently, evaluating the empty match reaches nil matcher fields and can panic.

A non-matching application constraint also demonstrates the incorrect dispatch: the parameter is applied by the deprecated path even though the v3 match should reject the target URL.

Without both fixes, an end-to-end test cannot prove the intended chain:

```text
Admin rule -> ZooKeeper node -> v3 parser/configurator -> target URL parameters -> invocation behavior
```

## Environment

- dubbo-go: `develop` at `a10df2f7`
- Go: `go1.25.5`
- OS: Windows amd64; the issue is platform-independent
- Config center: ZooKeeper

xiaobaicai66695:
Fixes #3661

## Background

Dubbo Admin publishes v3 dynamic configuration with the rule-level YAML field `configVersion` and optional `configs[].match` constraints. Dubbo-Go parses that YAML into configurator URLs before `overrideConfigurator` decides which matching semantics to use and applies parameters to consumer or provider URLs.

The Dynamic Config E2E exposed two runtime problems in this path. A rule could be present in Admin and ZooKeeper while the consumer either followed the deprecated configurator path or failed when evaluating an empty match. This made a persistence-only check insufficient and could produce a false-positive E2E result.

## Root cause

1. The configurator read the wrong version key.

   `ConfigVersionKey` belongs to the config-center component parameter namespace and currently resolves to `config-center.configVersion`. The dynamic rule schema and `DefaultConfigurationParser`, however, use `configVersion`. The constant originally also contained `configVersion`, but it was namespaced together with other config-center component keys in a historical registry-parameter cleanup. `overrideConfigurator` kept reusing it, so standard v3 rule URLs had no version from the configurator's perspective and fell into the deprecated pre-2.7 path.

2. Empty or partial v3 matches dereferenced nil fields.

   An omitted match or the valid YAML form `match: {}` leaves `ConditionMatch` or its optional fields nil. `ConditionMatch.IsMatch` called every matcher unconditionally, so the real v3 path could panic before applying the configured parameters. A partially specified match had the same problem for every omitted dimension.

## Why both fixes are necessary

The version-key fix is required to ensure a standard Admin v3 rule actually executes the v3 configurator path instead of accidentally succeeding through deprecated behavior. Once that dispatch is corrected, empty-match handling is required for the default match-all rule emitted by Admin and for partial match constraints.

Without both changes, the Dynamic Config E2E cannot prove the intended chain:

Admin rule -> config center data -> v3 parser/configurator -> final URL parameters -> observable invocation behavior.

These are runtime contract fixes, not test-only accommodations. The E2E uses load-balancing behavior for its assertion and does not require any CallOption semantic changes.

## Changes

- Add `constant.RuleConfigVersionKey` for the rule-level `configVersion` field, keeping it separate from config-center component keys.
- Use the rule constant consistently when parser-generated URLs are built, when configurator versions are read, and when metadata keys are excluded from applied parameters.
- Treat an absent or empty `ConditionMatch` as match-all and evaluate only explicitly configured match dimensions.
- Add regression coverage for rule-version dispatch, empty match behavior, parser output, and v3 configuration application.

## Testing

```text
go test ./config_center/parser ./config_center/configurator ./cluster/router/condition
```

### DIFF
--- a/config_center/configurator/override.go
+++ b/config_center/configurator/override.go
@@ -56,7 +56,7 @@ func (c *overrideConfigurator) Configure(url *common.URL) {
 	}
 
 	// branch for version 2.7.x
-	apiVersion := c.configuratorUrl.GetParam(constant.ConfigVersionKey, "")
+	apiVersion := c.configuratorUrl.GetParam(constant.RuleConfigVersionKey, "")
 	if len(apiVersion) != 0 {
 		var host string
 		currentSide := url.GetParam(constant.SideKey, "")
@@ -153,7 +153,7 @@ func getConditionKeys() *gxset.HashSet {
 	conditionKeys.Add(constant.VersionKey)
 	conditionKeys.Add(constant.ApplicationKey)
 	conditionKeys.Add(constant.SideKey)
-	conditionKeys.Add(constant.ConfigVersionKey)
+	conditionKeys.Add(constant.RuleConfigVersionKey)
 	conditionKeys.Add(constant.CompatibleConfigKey)
 	return conditionKeys
 }
--- a/config_center/configurator/override.go
+++ b/config_center/configurator/override.go
@@ -55,8 +55,11 @@ func (c *overrideConfigurator) Configure(url *common.URL) {
 		return
 	}
 
-	// branch for version 2.7.x
+	// Prefer the rule-level key for versioned rules (2.7.x and v3), while retaining the legacy key as a fallback.
 	apiVersion := c.configuratorUrl.GetParam(constant.RuleConfigVersionKey, "")
+	if len(apiVersion) == 0 {
+		apiVersion = c.configuratorUrl.GetParam(constant.ConfigVersionKey, "")
+	}
 	if len(apiVersion) != 0 {
 		var host string
 		currentSide := url.GetParam(constant.SideKey, "")
@@ -154,6 +157,7 @@ func getConditionKeys() *gxset.HashSet {
 	conditionKeys.Add(constant.ApplicationKey)
 	conditionKeys.Add(constant.SideKey)
 	conditionKeys.Add(constant.RuleConfigVersionKey)
+	conditionKeys.Add(constant.ConfigVersionKey)
 	conditionKeys.Add(constant.CompatibleConfigKey)
 	return conditionKeys
 }
--- a/config_center/parser/configuration_parser.go
+++ b/config_center/parser/configuration_parser.go
@@ -82,16 +82,19 @@ type ConditionMatch struct {
 }
 
 func (c *ConditionMatch) IsMatch(host string, url *common.URL) bool {
-	if !c.Address.IsMatch(host) {
+	if c == nil {
+		return true
+	}
+	if c.Address != nil && !c.Address.IsMatch(host) {
 		return false
 	}
-	if !c.ProviderAddress.IsMatch(url.Location) {
+	if c.ProviderAddress != nil && !c.ProviderAddress.IsMatch(url.Location) {
 		return false
 	}
-	if !c.Service.IsMatch(url.ServiceKey()) {
+	if c.Service != nil && !c.Service.IsMatch(url.ServiceKey()) {
 		return false
 	}
-	if !c.App.IsMatch(url.GetParam(constant.ApplicationKey, "")) {
+	if c.App != nil && !c.App.IsMatch(url.GetParam(constant.ApplicationKey, "")) {
 		return false
 	}
 	if c.Param != nil {
@@ -165,7 +168,7 @@ func serviceItemToUrls(item ConfigItem, config ConfiguratorConfig) ([]*common.UR
 		urlStr = urlStr + getEnabledString(item, config)
 		urlStr = urlStr + "&category="
 		urlStr = urlStr + constant.DynamicConfiguratorsCategory
-		urlStr = urlStr + "&configVersion="
+		urlStr = urlStr + "&" + constant.RuleConfigVersionKey + "="
 		urlStr = urlStr + config.ConfigVersion
 		apps := item.Applications
 		if len(apps) > 0 {
@@ -221,7 +224,7 @@ func appItemToUrls(item ConfigItem, config ConfiguratorConfig) ([]*common.URL, e
 			urlStr = urlStr + getEnabledString(item, config)
 			urlStr = urlStr + "&category="
 			urlStr = urlStr + constant.AppDynamicConfiguratorsCategory
-			urlStr = urlStr + "&configVersion="
+			urlStr = urlStr + "&" + constant.RuleConfigVersionKey + "="
 			urlStr = urlStr + config.ConfigVersion
 			url, err := common.NewURL(urlStr)
 			if err != nil {
--- a/common/constant/key.go
+++ b/common/constant/key.go
@@ -234,6 +234,7 @@ const (
 	MethodKey              = "method"
 	MethodKeys             = "methods"
 	RuleKey                = "rule"
+	RuleConfigVersionKey   = "configVersion"
 	RuntimeKey             = "runtime"
 	BackupKey              = "backup"
 	RoutersCategory        = "routers"
--- a/config_center/configurator/override_test.go
+++ b/config_center/configurator/override_test.go
@@ -30,6 +30,7 @@ import (
 	"dubbo.apache.org/dubbo-go/v3/common"
 	"dubbo.apache.org/dubbo-go/v3/common/constant"
 	"dubbo.apache.org/dubbo-go/v3/common/extension"
+	"dubbo.apache.org/dubbo-go/v3/config_center/parser"
 )
 
 const (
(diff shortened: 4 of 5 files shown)


Alanxtl:
update to latest develop to fix ci fail

xiaobaicai66695:
Addressed the Codecov patch-coverage report in 3b38d390. TestConditionMatchIsMatch now covers nil and empty match-all semantics, successful and rejected Address/ProviderAddress/Service/App/Param constraints, and a fully populated match. Local function coverage for ConditionMatch.IsMatch is 100%, and go test ./config_center/parser ./config_center/configurator ./cluster/router/condition passes. The Codecov note that its report is 922 commits behind develop refers to Codecov's comparison baseline; the PR branch itself remains rebased on the current PR base.

config_center/configurator/override.go
```diff
@@ -153,7 +153,7 @@ func getConditionKeys() *gxset.HashSet {
 	conditionKeys.Add(constant.VersionKey)
 	conditionKeys.Add(constant.ApplicationKey)
 	conditionKeys.Add(constant.SideKey)
-	conditionKeys.Add(constant.ConfigVersionKey)
+	conditionKeys.Add(constant.RuleConfigVersionKey)
 	conditionKeys.Add(constant.CompatibleConfigKey)
```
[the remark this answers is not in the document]
AlexStocks: [P1] 这里不能只把排除键替换成新的 configVersion，旧 config-center.configVersion 仍需要兼容处理

固定当前 Base/Head 的同一探针使用旧键 config-center.configVersion=v3.0、不匹配当前 application 的 V3 ConditionMatch，并配置 loadbalance=roundrobin。Base 保持目标 URL 的 random 且不复制控制键；当前 Head 则落入 configureDeprecated()，绕过 ConditionMatch 把目标改成 roundrobin，还把旧键 v3.0 复制进业务 URL。已有旧格式规则会因此无条件作用到非目标实例。请同时识别并排除两个键，明确两者并存时的优先级，并增加旧键 + 不匹配 V3 条件的兼容回归测试。

config_center/configurator/override.go
```diff
@@ -55,8 +55,11 @@ func (c *overrideConfigurator) Configure(url *common.URL) {
...
-	// branch for version 2.7.x
-	apiVersion := c.configuratorUrl.GetParam(constant.ConfigVersionKey, "")
+	// Prefer the rule-level key for versioned rules (2.7.x and v3), while retaining the legacy key as a fallback.
+	apiVersion := c.configuratorUrl.GetParam(constant.RuleConfigVersionKey, "")
+	if len(apiVersion) == 0 {
+		apiVersion = c.configuratorUrl.GetParam(constant.ConfigVersionKey, "")
+	}
 	if len(apiVersion) != 0 {
 		var host string
 		currentSide := url.GetParam(constant.SideKey, "")
```
Alanxtl, line 65: `currentSide != configuratorSide` 时只是让 `host` 为空，但仍继续执行 `configureIfMatchV3`。空 `match` 会返回 true，导致 `side: consumer` 的规则也被应用到 provider URL。

   复现：`side=consumer&match={}` 的 v3 规则配置 provider，`loadbalance` 仍会被改成规则值。应在 side 不匹配时直接返回，并补充 consumer/provider 反向场景测试。

config_center/parser/configuration_parser.go
```diff
@@ -82,16 +82,19 @@ type ConditionMatch struct {
...
 		return false
 	}
-	if !c.Service.IsMatch(url.ServiceKey()) {
+	if c.Service != nil && !c.Service.IsMatch(url.ServiceKey()) {
 		return false
 	}
-	if !c.App.IsMatch(url.GetParam(constant.ApplicationKey, "")) {
+	if c.App != nil && !c.App.IsMatch(url.GetParam(constant.ApplicationKey, "")) {
 		return false
 	}
```
Alanxtl, line 100: 这里使用的是 `yaml:"app"`，但官方 v3 配置格式字段是 `application`。[[官方文档](https://dubbo.apache.org/en/overview/what/core-features/traffic/configuration-rule/)](https://dubbo.apache.org/en/overview/what/core-features/traffic/configuration-rule/)明确使用 `match.application`。

   YAML 中的 `application` 会被静默忽略；而本 PR 又把 nil matcher 当成“不限制”，所以限定 `app-a` 的规则会错误地应用到 `app-b`。建议兼容/改为 `yaml:"application"`，并增加 parser → configurator 的不匹配回归测试。

补充：`config-center.configVersion` 旧键兼容逻辑已在最新提交中处理。`go test ./...` 全量通过，但上述两个负向场景当前测试未覆盖。

Issue #3664, opened 2026-08-14
[BUG] Concurrent access to Environment.dynamicConfiguration races in common/config


lizining1231:
### Summary

`go test -race ./registry/directory/` intermittently reports a data race: a subscription goroutine left over from an earlier test reads the global `common/config.Environment.dynamicConfiguration` field while the current test goroutine writes it — no synchronization, no happens-before edge.

### Affected Locations

| Package | Test | Test Location | Related Source |
| --- | --- | --- | --- |
| registry/directory | `TestNewRegistryDirectoryConsumerListenerUsesResolvedApplicationName` | [directory_test.go#L191-L198](https://github.com/apache/dubbo-go/blob/master/registry/directory/directory_test.go#L191-L198) (write at L195) | Write: [environment.go#L100-L104](https://github.com/apache/dubbo-go/blob/master/common/config/environment.go#L100-L104); Read: [environment.go#L106-L111](https://github.com/apache/dubbo-go/blob/master/common/config/environment.go#L106-L111) via [router.go#L109](https://github.com/apache/dubbo-go/blob/master/cluster/router/tag/router.go#L109) → [chain.go#L115](https://github.com/apache/dubbo-go/blob/master/cluster/router/chain/chain.go#L115) → [directory.go#L413](https://github.com/apache/dubbo-go/blob/master/registry/directory/directory.go#L413) → [mock_registry.go#L116](https://github.com/apache/dubbo-go/blob/master/registry/mock_registry.go#L116) |

### Reproduction

```bash
# before the fix: deterministic (minimal stress test)
go test -race -run TestDynamicConfigurationRace ./common/config/

# CI: flaky under the full suite
go test -race ./registry/directory/
```

### Root Cause

- Global lock-free field: `Get/SetDynamicConfiguration` are plain field reads/writes.
- `MockRegistry.Subscribe` goroutine blocks on `listener.Next()` forever, surviving across tests (reader goroutine ID < writer goroutine ID proves cross-test).
- Lingering reader consumes backlogged events while the current test writes → race.

### Proposed Fix

- Add `sync.RWMutex` (write lock in Set, read lock in Get). 
- Regression test `TestDynamicConfigurationRace`: 100k reads vs 100k writes, fails before fix, passes after.

lizining1231:
此PR #3645 的CI复现过该问题，这是grep后的日志 ：[ci-race.txt](https://github.com/user-attachments/files/31085412/ci-race.txt)
[...]

PR #3668, opened 2026-08-16
fix(protocol/triple): SendHeader writes into ResponseHeader instead of RequestHeader

### LINKED ISSUE #3667: [Bug] protocol/triple: SendHeader writes response headers into RequestHeader instead of ResponseHeader
## Description

`SendHeader` is documented to "append response headers from a server handler and send them immediately" (same semantics as `SetHeader`, but flushed on the first `Send`). However, its implementation merges the user-supplied headers into `conn.RequestHeader()` rather than `conn.ResponseHeader()`, so the headers never reach the wire.

## Root cause

`protocol/triple/triple_protocol/header.go:307-314`

```go
func SendHeader(ctx context.Context, header http.Header) error {
	conn, ok := ctx.Value(handlerOutgoingKey{}).(StreamingHandlerConn)
	if !ok {
		return errorf(CodeInternal, "triple: handler outgoing context not found; SendHeader must be called within a Triple handler")
	}
	mergeHeaders(conn.RequestHeader(), header)   // ← should be conn.ResponseHeader()
	return conn.Send(nil)
}
```

Compare the sibling `SetHeader` at `header.go:263`, which correctly calls `conn.ResponseHeader()`.

## Impact

Taking the gRPC handler conn (`protocol_grpc.go`) as an example, `RequestHeader()` returns the inbound `hc.request.Header`, and `grpcHandlerConn.Send` only flushes `hc.responseHeader` into the response writer on the first `Send`. As a result:

1. **Response headers are silently dropped** — `SendHeader`'s payload is never sent to the client, contradicting both its doc comment and the grpc-style metadata contract.
2. **Request headers are polluted** — the user's response headers are merged into the shared inbound request header map, so subsequent `RequestHeader()` / `FromIncomingContext` reads observe headers that were never sent by the client. This can cause cross-talk / misinterpretation of inbound metadata in downstream handlers and interceptors.

## Why CI hasn't caught it

There is zero test coverage for `SendHeader`:

```
$ grep -rn "SendHeader" protocol/triple/triple_protocol/*_test.go
# (no matches)
```

`header_test.go` has `TestSetHeader` and `TestSetTrailer` but no `TestSendHeader`. The existing `mockHandlerConn.RequestHeader()` returns `nil` (`header_test.go:185`), so even a naive test would not surface the mismatch.

## Suggested fix

One-line change plus a regression test:

```go
func SendHeader(ctx context.Context, header http.Header) error {
	conn, ok := ctx.Value(handlerOutgoingKey{}).(StreamingHandlerConn)
	if !ok {
		return errorf(CodeInternal, "triple: handler outgoing context not found; SendHeader must be called within a Triple handler")
	}
	mergeHeaders(conn.ResponseHeader(), header)
	return conn.Send(nil)
}
```

And in `header_test.go`, add a `TestSendHeader` that asserts the headers land in `ResponseHeader()` and that `RequestHeader()` is **not** mutated. The mock should return a real, distinct `http.Header` from `RequestHeader()` so pollution can be detected.

## Notes

- Verified on latest `origin/develop` (`a31acfb6`): line 312 still reads `conn.RequestHeader()`.
- Related PRs (#3633 error return, #3649 / #3663 docs & tests) do not touch this line.
- I can submit a PR with the fix + test if maintainers agree on the direction.

Lcos-000:
## What & why

`SendHeader` is documented to "append response headers from a server handler and send them immediately" (same semantics as `SetHeader`, but flushed on the first `Send`). Its implementation merged the user-supplied headers into `conn.RequestHeader()` instead of `conn.ResponseHeader()`, so:

1. **Response headers were silently dropped** — `grpcHandlerConn.Send` only flushes `hc.responseHeader` into the response writer, so the user's headers never reached the client.
2. **Request headers were polluted** — the headers were merged into the shared inbound request header map, so subsequent `RequestHeader()` / `FromIncomingContext` reads observed headers the client never sent.

Align with the sibling `SetHeader` (`header.go:263`) by writing into `conn.ResponseHeader()`.

Closes #3667.

## Changes

- `protocol/triple/triple_protocol/header.go:312` — `conn.RequestHeader()` → `conn.ResponseHeader()`.
- `protocol/triple/triple_protocol/header_test.go` — upgrade `mockHandlerConn` to expose a real `RequestHeader` map and a `Send` call counter (the old `RequestHeader()` returned `nil`, which masked any pollution); add `TestSendHeader` (asserts headers land in `ResponseHeader()`, `RequestHeader()` is **not** mutated, and `Send` is called once) and `TestSendHeaderOutsideHandler` (asserts `CodeInternal`).
- `protocol/triple/triple_protocol/triple_ext_test.go` — add `TestSendHeaderInUnaryHandler`, an end-to-end test over a real `httptest` HTTP transport. The mock in `header_test.go` does not simulate `net/http`'s `WriteHeader` snapshot semantics, so it could not catch the "headers never reach the wire" failure mode. This test also covers the unary handler's double-`Send` path (`SendHeader`'s `conn.Send(nil)` followed by the framework's `conn.Send(msg)`) and verifies the response body stays decodable after the header flush.

## Verification

- Confirmed `TestSendHeader` (mock) and `TestSendHeaderInUnaryHandler` (e2e) both **FAIL** on the pre-fix code (`got: []` / `got: []`), proving the new tests reproduce the bug.
- After the one-line fix, both pass; the full `protocol/triple/triple_protocol` package test suite is green; `go vet` / `go build` clean.

## Notes

- Based on the latest `origin/develop` (a31acfb6).
- Scope is intentionally limited to the public `SendHeader` API. The compat layer's `compatHandlerStream.SendHeader` remains a documented no-op (handled by #3663) and is out of scope.
- A streaming e2e covering `grpcHandlerConn.Send`'s flush path would be a natural follow-up, but is left to a separate change to keep this fix focused.

### DIFF
--- a/protocol/triple/triple_protocol/header.go
+++ b/protocol/triple/triple_protocol/header.go
@@ -309,6 +309,6 @@ func SendHeader(ctx context.Context, header http.Header) error {
 	if !ok {
 		return errorf(CodeInternal, "triple: handler outgoing context not found; SendHeader must be called within a Triple handler")
 	}
-	mergeHeaders(conn.RequestHeader(), header)
+	mergeHeaders(conn.ResponseHeader(), header)
 	return conn.Send(nil)
 }
--- a/protocol/triple/triple_protocol/header_test.go
+++ b/protocol/triple/triple_protocol/header_test.go
@@ -165,26 +165,29 @@ func TestAppendToOutgoingContextPanicsOnOddKV(t *testing.T) {
 	})
 }
 
-// mockHandlerConn is a minimal StreamingHandlerConn for testing SetHeader and
-// SetTrailer.
+// mockHandlerConn is a minimal StreamingHandlerConn for testing SetHeader,
+// SendHeader and SetTrailer.
 type mockHandlerConn struct {
 	responseHdr  http.Header
 	responseTrlr http.Header
+	requestHdr   http.Header
+	sendCalls    int
 }
 
 func newMockHandlerConn() *mockHandlerConn {
 	return &mockHandlerConn{
 		responseHdr:  make(http.Header),
 		responseTrlr: make(http.Header),
+		requestHdr:   make(http.Header),
 	}
 }
 
 func (m *mockHandlerConn) Spec() Spec                    { return Spec{} }
 func (m *mockHandlerConn) Peer() Peer                    { return Peer{} }
 func (m *mockHandlerConn) Receive(any) error             { return nil }
-func (m *mockHandlerConn) RequestHeader() http.Header    { return nil }
+func (m *mockHandlerConn) RequestHeader() http.Header    { return m.requestHdr }
 func (m *mockHandlerConn) ExportableHeader() http.Header { return nil }
-func (m *mockHandlerConn) Send(any) error                { return nil }
+func (m *mockHandlerConn) Send(any) error                { m.sendCalls++; return nil }
 func (m *mockHandlerConn) ResponseHeader() http.Header   { return m.responseHdr }
 func (m *mockHandlerConn) ResponseTrailer() http.Header  { return m.responseTrlr }
 
@@ -208,6 +211,33 @@ func TestSetTrailer(t *testing.T) {
 	err := SetTrailer(ctx, http.Header{"X-Trailer": []string{"end"}})
 	assert.Nil(t, err)
 	assert.Equal(t, conn.responseTrlr.Get("X-Trailer"), "end")
+}
+
+// TestSendHeader verifies that SendHeader merges headers into the response
+// header buffer (not the request header) and triggers a flush via Send.
+func TestSendHeader(t *testing.T) {
+	t.Parallel()
+	conn := newMockHandlerConn()
+	ctx := context.WithValue(context.Background(), handlerOutgoingKey{}, conn)
+	err := SendHeader(ctx, http.Header{"X-Custom": []string{"value"}})
+	assert.Nil(t, err)
+	// Headers must land in the response header buffer.
+	assert.Equal(t, conn.responseHdr.Get("X-Custom"), "value")
+	// The request header must not be polluted.
+	assert.Equal(t, len(conn.requestHdr), 0)
+	// Send must be called exactly once to flush the headers.
+	assert.Equal(t, conn.sendCalls, 1)
+}
+
+// TestSendHeaderOutsideHandler verifies that SendHeader returns a CodeInternal
+// error when called outside a Triple handler context.
+func TestSendHeaderOutsideHandler(t *testing.T) {
+	t.Parallel()
+	err := SendHeader(context.Background(), http.Header{"X-Custom": []string{"value"}})
+	assert.NotNil(t, err)
+	tripleErr, ok := asError(err)
+	assert.True(t, ok)
+	assert.Equal(t, tripleErr.Code(), CodeInternal)
 }
 
 // TestSetHeaderOutsideHandler verifies that SetHeader returns a CodeInternal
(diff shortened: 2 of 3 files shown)


Lcos-000:
The Unit Test failure is `registry/directory.TestNewRegistryDirectoryConsumerListenerUsesResolvedApplicationName` — the log shows a data race in `common/config.Environment` (SetDynamicConfiguration vs GetDynamicConfiguration). This looks like the race already tracked in #3664, which has been on develop for a while. Should be unrelated to this change.

Alanxtl (approved):
lgtm

Alanxtl:
can u help us identify why intergration test fail to test this bug?
https://github.com/apache/dubbo-go-samples/tree/main/triple_header_trailer
https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-client/cmd/main.go line 94

Lcos-000:
I ve seen it,casuse the intergration-test never use the function `SendHeader`,so it can not test this bug,
if u use the `SendHeader`in the test https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go line 55,
the panic will happen in https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-client/cmd/main.go line 63,
and it would cause the requestheader be polluted.

Alanxtl:
> I ve seen it,casuse the intergration-test never use the function `SendHeader`,so it can not test this bug, if u use the `SendHeader`in the test https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go line 55, the panic will happen in https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-client/cmd/main.go line 63, and it would cause the requestheader be polluted.

But the integration test clearly runs `SendHeader` on line 55 of [https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go](https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go). The integration test will run it automatically.

Can you please help us create a pr to dubbo-go-samples to fix the integration test

Lcos-000:
> > I ve seen it,casuse the intergration-test never use the function `SendHeader`,so it can not test this bug, if u use the `SendHeader`in the test https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go line 55, the panic will happen in https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-client/cmd/main.go line 63, and it would cause the requestheader be polluted.
> 
> But the integration test clearly runs `SendHeader` on line 55 of https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go. The integration test will run it automatically.
> 
> Can you please help us create a pr to dubbo-go-samples to fix the integration test

Thk! Just a small clarification — line 55 is calling SetHeader, not SendHeader. I think there might be a minor oversight. Could u please take another look?
image

Lcos-000:
https://github.com/apache/dubbo-go-samples/pull/1139  here is the pr created to test the bug in `dubbo-go-samples`

Alanxtl:
> > > I ve seen it,casuse the intergration-test never use the function `SendHeader`,so it can not test this bug, if u use the `SendHeader`in the test https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go line 55, the panic will happen in https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-client/cmd/main.go line 63, and it would cause the requestheader be polluted.
> > 
> > 
> > But the integration test clearly runs `SendHeader` on line 55 of https://github.com/apache/dubbo-go-samples/blob/main/triple_header_trailer/go-server/cmd/main.go. The integration test will run it automatically.
> > Can you please help us create a pr to dubbo-go-samples to fix the integration test
> 
> Thk! Just a small clarification — line 55 is calling SetHeader, not SendHeader. I think there might be a minor oversight. Could u please take another look? image

okay, my mistake 😂

Alanxtl:
update to latest develop to fix the ci fail

PR #3639, opened 2026-08-10
fix(triple): guard httpSrv/http3Srv access with atomic pointers

lizining1231:
### Description
Fixes #3638 

Fixes the data race on Server.httpSrv and Server.http3Srv in the triple
protocol server, reported by go test -race ./server/ -run `TestCfgAPI_Export`.

Server.Start starts the HTTP/2 and HTTP/3 listeners in background goroutines
(startHttp2 / startHttp3 / startHttp2AndHttp3), which write s.httpSrv /
s.http3Srv, while Stop / GracefulStop run on a different goroutine and read
the same fields to call Close / Shutdown — a data race under the Go memory
model.

### Changes

- `httpSrv *http.Server` → `uatomic.Pointer[http.Server]`
- `http3Srv *http3.Server` → `uatomic.Pointer[http3.Server]`
- All write sites go through `Store`
- All read sites go through `Load`; `Stop` / `GracefulStop` load once into a
  local snapshot before nil-checking and closing (no TOCTOU)
- Adapt the existing QUIC config tests (`TestServer_HTTP3PathsUseQUICConfigHelper`)
  to `Load()` the atomic pointers before the `assert.Nil` checks, since the
  fields are now `uatomic.Pointer` value types that are never nil themselves

### Test

Add 13 tests that start the real HTTP/2, HTTP/3 and dual-protocol servers on
random ports:

| Group | Test | Description |
| --- | --- | --- |
| Start + Stop | 1. `TestServer_HTTP2_StartAndStop` | Starts an HTTP/2 server, asserts `httpSrv` is stored and `http3Srv` is empty; `Stop` makes `Run` return `http.ErrServerClosed` |
| | 2. `TestServer_HTTP2_StartAndStopWithTLS` | Starts HTTP/2 with TLS, covering the TLS branch of start and stop |
| | 3. `TestServer_HTTP3_StartAndStop` | Starts HTTP/3 (QUIC/UDP), asserts `http3Srv` is stored and `httpSrv` is empty; `Stop` makes `Run` return `http.ErrServerClosed` |
| | 4. `TestServer_HTTP2AndHTTP3_StartAndStop` | Starts both protocols, asserts both fields are non-empty; `Stop` exits cleanly |
| Start + GracefulStop | 1. `TestServer_HTTP2_StartAndGracefulStop` | HTTP/2 `GracefulStop` (`Shutdown` branch), covering the production destroy path |
[...]

PR #3671, opened 2026-08-16
fix(triple): filter http.ErrServerClosed in single-protocol server shutdown

### LINKED ISSUE #3670: [Bug] single-protocol server misreports normal Stop as serve failure
### Summary

In `protocol/triple`, a single-protocol (HTTP/2 or HTTP/3) server leaks `http.ErrServerClosed` from `Server.Run` after a normal `Stop`. The outer `startTransport` goroutine treats any non-nil error as a serve failure and logs `[Triple][Server] server serve failed, err=http: Server closed` every time the server is shut down normally — spurious error logs and monitoring noise. The dual-protocol path (`startHttp2AndHttp3`) already filters `http.ErrServerClosed` inside its errgroup, so the single-protocol and dual-protocol paths behave inconsistently.

### Affected Locations

| Package | Test | Test Location | Related Source |
| --- | --- | --- | --- |
| protocol/triple/triple_protocol | `TestServer_HTTP2_StartAndStop` / `TestServer_HTTP3_StartAndStop` / `TestServer_HTTP2_StartAndGracefulStop` / `TestServer_HTTP3_StartAndGracefulStop` / `TestServer_RepeatedStartStop` | https://github.com/apache/dubbo-go/blob/master/protocol/triple/triple_protocol/server_lifecycle_test.go#L170-L203 | Single-protocol: https://github.com/apache/dubbo-go/blob/master/protocol/triple/triple_protocol/server.go#L201-L220 (`startHttp2` returns err directly) and `#L222-L250` (`startHttp3`); Logging: https://github.com/apache/dubbo-go/blob/master/protocol/triple/server.go#L230-L241 (`startTransport` goroutine logs `serve failed` on any non-nil error) |

### Reproduction

```bash
# single-protocol HTTP/2 server, then Stop:
go test -run 'TestServer_HTTP2_StartAndStop' ./protocol/triple/triple_protocol/

# observed: Run returns http.ErrServerClosed after a normal Stop (test asserts it today),
# which makes the outer goroutine log "server serve failed, err=http: Server closed"
```

The existing tests currently **assert** the leak (`require.ErrorIs(..., http.ErrServerClosed)`), i.e. they pin the buggy behavior as expected instead of asserting a clean shutdown. The dual-protocol sibling test asserts `require.NoError(...)` — the inconsistency is baked into the test suite.

### Root Cause

- `startHttp2` (`server.go#L201-L220`) and `startHttp3` (`server.go#L222-L250`) return the `ListenAndServe` error verbatim, without filtering `http.ErrServerClosed` — the normal-shutdown signal from net/http.
- `startHttp2AndHttp3` (`server.go#L292-L313`) already filters `err != http.ErrServerClosed` inside the errgroup, so dual-protocol `Run` returns `nil` after a normal stop.
- `startTransport` (`protocol/triple/server.go#L230-L241`) logs `logger.Errorf("... server serve failed, err=%v", runErr)` for any non-nil `Run` error.
- Net effect: single-protocol deployments log a spurious error on every normal shutdown, while dual-protocol deployments do not — inconsistent behavior across the three `Run` switch branches.

### Proposed Fix

- Filter `http.ErrServerClosed` in `startHttp2` and `startHttp3` before returning, aligning them with `startHttp2AndHttp3`. Note: only `http.ErrServerClosed` is swallowed; genuine errors (e.g. `EADDRINUSE` on bind) still propagate.
- Update the 6 assertions in `server_lifecycle_test.go` (L178/L189/L203/L233/L249/L340) from `require.ErrorIs(..., http.ErrServerClosed)` to `require.NoError(...)`. The `if/else` in `TestServer_RepeatedStartStop` then becomes uniform and can be simplified.
- Verification: the updated tests fail on the pre-fix code (they assert `nil` while `Run` returns `ErrServerClosed`) and pass after the one-line fix; full `./protocol/triple/...` suite green.

lizining1231:
### Description
Fixes #3670

The single-protocol `Server.Run` paths returned `http.ErrServerClosed` verbatim after a normal `Stop`, so every clean shutdown surfaced the net/http normal-closure signal to the caller. The outer `startTransport` goroutine logs any non-nil error as `server serve failed`, so each shutdown produced a spurious error log; the dual-protocol path already swallowed the signal inside its errgroup, so single- and dual-protocol shutdown behaviour diverged.

Align the single-protocol paths with the dual-protocol one: filter `http.ErrServerClosed` in `startHttp2` / `startHttp3` before returning. Only the normal-closure signal is suppressed; genuine serve errors (e.g. `EADDRINUSE`) still propagate.

### Changes

- `triple_protocol/server.go` — filter `http.ErrServerClosed` in `startHttp2` / `startHttp3` before returning, matching the dual-protocol behaviour
- `triple_protocol/server_lifecycle_test.go` — switch the six `require.ErrorIs(..., http.ErrServerClosed)` assertions to `require.NoError`; fold the now-uniform `if/else` in `TestServer_RepeatedStartStop` into a single assertion; drop the stale comments that described only the dual-protocol swallowing behaviour

### Test

Add 2 tests guarding the boundary of the filter:

| Test | Description |
| --- | --- |
| `TestServerRunReturnsBindErrorWhenPortInUse` | Occupies the TCP port before `Run`, asserting that a genuine serve error (`EADDRINUSE`) is still propagated instead of being swallowed together with `http.ErrServerClosed` |
| `TestServerRunHTTP3ReturnsBindErrorWhenPortInUse` | Same boundary check on the HTTP/3 path: the QUIC endpoint fails to bind the occupied UDP port and `Run` still returns the error |

### Validation

- Before the fix, a normal `Stop` made `Run` return `http.ErrServerClosed` and the outer goroutine logged a spurious `server serve failed` error; after the fix `Run` returns `nil` and no error is logged
- The updated lifecycle tests fail on the pre-fix code and pass with the filter

### Checklist

### DIFF
--- a/protocol/triple/triple_protocol/server.go
+++ b/protocol/triple/triple_protocol/server.go
@@ -215,8 +215,10 @@ func (s *Server) startHttp2(tlsConf *tls.Config) error {
 	} else {
 		err = srv.ListenAndServe()
 	}
-
-	return err
+	if err != nil && err != http.ErrServerClosed {
+		return err
+	}
+	return nil
 }
 
 func (s *Server) startHttp3(tlsConf *tls.Config) error {
@@ -246,7 +248,11 @@ func (s *Server) startHttp3(tlsConf *tls.Config) error {
 
 	logger.Debugf("[Triple][Server] triple HTTP/3 Server starting on %v", s.addr)
 
-	return s.http3Srv.Load().ListenAndServe()
+	err = s.http3Srv.Load().ListenAndServe()
+	if err != nil && err != http.ErrServerClosed {
+		return err
+	}
+	return nil
 }
 
 func (s *Server) startHttp2AndHttp3(tlsConf *tls.Config) error {
--- a/protocol/triple/triple_protocol/server_lifecycle_test.go
+++ b/protocol/triple/triple_protocol/server_lifecycle_test.go
@@ -28,7 +28,6 @@ import (
 	"errors"
 	"math/big"
 	"net"
-	"net/http"
 	"syscall"
 	"testing"
 	"time"
@@ -175,7 +174,7 @@ func TestServer_HTTP2_StartAndStop(t *testing.T) {
 	require.Nil(t, srv.http3Srv.Load())
 
 	require.NoError(t, srv.Stop())
-	require.ErrorIs(t, waitForServerExit(t, errCh, 5*time.Second), http.ErrServerClosed)
+	require.NoError(t, waitForServerExit(t, errCh, 5*time.Second))
 }
 
 func TestServer_HTTP2_StartAndStopWithTLS(t *testing.T) {
@@ -186,7 +185,7 @@ func TestServer_HTTP2_StartAndStopWithTLS(t *testing.T) {
 	require.Nil(t, srv.http3Srv.Load())
 
 	require.NoError(t, srv.Stop())
-	require.ErrorIs(t, waitForServerExit(t, errCh, 5*time.Second), http.ErrServerClosed)
+	require.NoError(t, waitForServerExit(t, errCh, 5*time.Second))
 }
 
 func TestServer_HTTP3_StartAndStop(t *testing.T) {
@@ -200,7 +199,7 @@ func TestServer_HTTP3_StartAndStop(t *testing.T) {
 	require.Nil(t, srv.httpSrv.Load())
 
 	require.NoError(t, srv.Stop())
-	require.ErrorIs(t, waitForServerExit(t, errCh, 5*time.Second), http.ErrServerClosed)
+	require.NoError(t, waitForServerExit(t, errCh, 5*time.Second))
 }
 
 func TestServer_HTTP2AndHTTP3_StartAndStop(t *testing.T) {
@@ -215,8 +214,6 @@ func TestServer_HTTP2AndHTTP3_StartAndStop(t *testing.T) {
 	require.NotNil(t, srv.http3Srv.Load())
 
 	require.NoError(t, srv.Stop())
-	// startHttp2AndHttp3 swallows http.ErrServerClosed inside the errgroup,
-	// so Run returns nil after the servers are closed.
 	require.NoError(t, waitForServerExit(t, errCh, 5*time.Second))
 }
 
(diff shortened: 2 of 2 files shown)


lizining1231:
这个改动会修改我在 #3639 中写的部分测试断言，原先断言按照原先实际行为编写，现作修改

Alanxtl:
如果是因为错误而关闭了server是不是同样也走这个分支

lizining1231:
> 如果是因为错误而关闭了server是不是同样也走这个分支

是的
```go
if err != nil && err != http.ErrServerClosed {
    return err   // genuine errors still propagate
}
return nil       // only the normal-closure signal is swallowed
```
可以构造一个因为错误关闭server的测试：
```go
// TestZZNonNormalShutdownLogs verifies that a genuine serve error (port
// conflict) surfaces as a "server serve failed" log line through the real
// startTransport -> Run call chain. Temporary probe, deleted after the run.
func TestZZNonNormalShutdownLogs(t *testing.T) {
	// Occupy a TCP port so the upcoming ListenAndServe fails to bind.
	l, err := net.Listen("tcp", "127.0.0.1:0")
	require.NoError(t, err)
	defer l.Close()

	srv := NewServer(&global.TripleConfig{})
	srv.triServer = tri.NewServer(l.Addr().String(), &global.TripleConfig{})
	srv.startTransport(constant.CallHTTP2, nil)

	// Give the goroutine time to run Run() and log the failure.
	time.Sleep(500 * time.Millisecond)
}
```
运行的话就是正常看到报错日志：
```bash
lizining@Y:~/projects/dubbo-go$ go test -v -run TestZZNonNormalShutdownLogs ./protocol/triple/
=== RUN   TestZZNonNormalShutdownLogs
2026-08-17T14:17:09.038+0800    ERROR   triple/server.go:238  [Triple][Server] server serve failed, err=listen tcp 127.0.0.1:37831: bind: address already in use
dubbo.apache.org/dubbo-go/v3/protocol/triple.(*Server).startTransport.func1
        /home/lizining/projects/dubbo-go/protocol/triple/server.go:238
--- PASS: TestZZNonNormalShutdownLogs (0.50s)
PASS
ok      dubbo.apache.org/dubbo-go/v3/protocol/triple    (cached)
```

Alanxtl (approved):
lgtm

Issue #3644, opened 2026-08-11
[OSPP] 补齐 Dubbo-Go 注释与单测覆盖


Alanxtl:
## 背景
这批任务面向 OSPP 新人，目标是通过小 PR 补齐 Dubbo-Go 中面向用户的导出注释，并补充对应模块的单测覆盖。

# 需要在评论区里面评论【认领 + 任务序号】
# 未认领的 PR 不予接受
# AI 生成的 PR 不予接受
# 对 OSPP 感兴趣的同学一定要通过 README 里面的方式加群沟通
# **please note that a complete AI PR is not welcome.**

目标仓库：

dubbo-go https://github.com/apache/dubbo-go

提交 PR 的时候 @Alanxtl，我到时候会去看。

## 任务总览
| 序号 | 任务名称 | 负责人 | 任务链接 |
| --- | --- | --- | --- |
| 1 | 补齐 Metrics Options / Error Classifier 注释与单测覆盖 | @tangerineyu | https://github.com/apache/dubbo-go/pull/3665 |
| 2 | 补齐 Metrics Probe 注释与单测覆盖 | 待认领 | [查看任务](#2-补齐-metrics-probe-注释与单测覆盖) |
| 3 | 补齐 Registry Options 注释与单测覆盖 | @641-git641 | https://github.com/apache/dubbo-go/pull/3653 |
| 4 | 补齐 Service Discovery Metadata Customizer 注释与单测覆盖 | @DaWesen | https://github.com/apache/dubbo-go/pull/3678 |
| 5 | 补齐 Service Discovery Instance Customizer 注释与单测覆盖 | @DaWesen | https://github.com/apache/dubbo-go/pull/3676 |
| 6 | 补齐 Metadata Options 注释与单测覆盖 | @MaoMengww | https://github.com/apache/dubbo-go/pull/3657 |
| 7 | 补齐 Metadata Info 注释与单测覆盖 | @MaoMengww | https://github.com/apache/dubbo-go/pull/3666 |
| 8 | 补齐 Triple Header / Error 注释与单测覆盖 | @lizining1231 | https://github.com/apache/dubbo-go/pull/3649 |
| 9 | 补齐 Triple Stream 注释与单测覆盖 | @lizining1231  | https://github.com/apache/dubbo-go/pull/3663 |
| 10 | 补齐 Router Chain 注释与单测覆盖 | @SouthwestAsiaFloat | https://github.com/apache/dubbo-go/pull/3662 |
[...]

PR #3682, opened 2026-08-19
docs(router): complete script router comments and unit test coverage

MaoMengww:
### Description
- Add doc comments to the ScriptRouter lifecycle methods (Process, Route, Notify, ...) describing the script instance lifecycle, reuse (reference-counting) behavior, and failure handling
- Add doc comments to ScriptInstances, jsInstances, program, and the instance pool
- Add unit tests covering: instance creation failure, script execution failure, instance reuse, and concurrent instance fetching

#### Worth mentioning:

- This PR also fixes a potential nil-pointer panic: in router.go lines 120 and 124, *cfg.Enabled was dereferenced unconditionally. Since RouterConfig is populated by YAML decoding and Enabled has no default value, omitting the field in the rule would crash the router. The dereference is now guarded, dafault true, and a regression test has been added.
- The goja runtime pooling in this file is only half-implemented: insPool *sync.Pool (intended to store *goja.Runtime) only has Get calls but no Put, so every compilation creates a new runtime, degrading performance. if the maintainers agree — I'll open a separate issue for it and fix it.

Refs https://github.com/apache/dubbo-go/issues/3644 (task11)
### Checklist

### DIFF
--- a/cluster/router/script/router.go
+++ b/cluster/router/script/router.go
@@ -49,6 +49,7 @@ type ScriptRouter struct {
 	rawScript  string
 }
 
+// NewScriptRouter creates a new ScriptRouter.
 func NewScriptRouter() *ScriptRouter {
 	return &ScriptRouter{
 		applicationName: "",
@@ -56,6 +57,7 @@ func NewScriptRouter() *ScriptRouter {
 	}
 }
 
+// parseRoute decodes a YAML script rule into a RouterConfig.
 func parseRoute(routeContent string) (*global.RouterConfig, error) {
 	routeDecoder := yaml.NewDecoder(strings.NewReader(routeContent))
 	routerConfig := &global.RouterConfig{}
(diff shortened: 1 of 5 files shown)


cluster/router/script/router.go
```diff
@@ -117,11 +123,11 @@ func (s *ScriptRouter) Process(event *config_center.ConfigChangeEvent) {
 			logger.Error("[Router][Script] applicationName field must be set in config")
 			return
 		}
-		if !*cfg.Enabled {
+		if cfg.Enabled != nil && !*cfg.Enabled {
```
Alanxtl, line 126: cfg.Enabled == nil是不是也应该走这个分支呀
也就是 cfg.Enabled == nil || !*cfg.Enabled

cluster/router/script/router.go
```diff
@@ -117,11 +123,11 @@ func (s *ScriptRouter) Process(event *config_center.ConfigChangeEvent) {
...
 			logger.Infof("[Router][Script] enabled field equals false, this rule will be ignored, script=%s", cfg.Script)
 		}
 		// rewrite to ScriptRouter
-		s.enabled = *cfg.Enabled
+		s.enabled = cfg.Enabled == nil || *cfg.Enabled
```
Alanxtl, line 130: 这里不应该是cfg.Enabled != nil && *cfg.Enabled吗

cluster/router/script/router.go
MaoMengww: 我认为的是当Enabled没有填写时默认为true, 如果认为这样不合理的话我马上进行修改

cluster/router/script/router.go
MaoMengww: 同上

Alanxtl (approved):
lgtm

PR #3680, opened 2026-08-18
fix(filter/generic): guard $invoke arg type assertions to avoid panic

Lcos-000:
### Description
Fixes #3679

`genericServiceFilter.Invoke` (`filter/generic/service_filter.go:79,82`) performed unguarded type assertions on the generic-call arguments:

```go
mtdName := inv.Arguments()[0].(string)
// ...
args := inv.Arguments()[2].([]hessian.Object)
```

`IsGenericInvocation()` (`protocol/invocation/rpcinvocation.go:94-99`) only checks that `MethodName` is `$invoke`/`$invokeAsync` and that `len(Arguments) == 3`; it does **not** validate the element types. A malformed or malicious `$invoke` request with wrong-typed arguments reached these lines and panicked with `interface conversion: interface {} is X, not Y`.

`genericServiceFilter` is in `DefaultServiceFilters` (`common/constant/default.go:67-69`), so this was reachable by any inbound dubbo request with no opt-in required.

This PR replaces the two bare assertions with comma-ok form and returns a `perrors`-wrapped error via `result.RPCResult{Err}` (matching the existing error-return style used at `service_filter.go:92,99`), so a malformed `$invoke` now returns an error instead of panicking.

### Checklist
- [x] I confirm the target branch is `develop`
- [x] I have run `make fmt` to format my code
- [x] I have run `make test` to run local tests
- [x] I have added tests that prove my fix is effective or that my feature works

PR #3684, opened 2026-08-19
fix(invocation): guard ActualMethodName type assertion to avoid panic

### LINKED ISSUE #3683: [Bug] ActualMethodName() panics on malformed $invoke arguments
### Description

Follow-up to #3680.

`RPCInvocation.ActualMethodName()` (`protocol/invocation/rpcinvocation.go:85-91`) performs an unguarded type assertion on the first generic-call argument:

```go
func (r *RPCInvocation) ActualMethodName() string {
    if r.IsGenericInvocation() {
        return r.Arguments()[0].(string)   // panics if arg[0] is not a string
    } else {
        return r.MethodName()
    }
}
```

`IsGenericInvocation()` only checks `len(Arguments)==3`, not the element types. A malformed `$invoke` with a non-string `arg[0]` reaches this line and panics with `interface conversion: interface {} is X, not string`.

### Reproduction

```go
inv := invocation.NewRPCInvocation(constant.Generic,
    []any{123, []string{}, []hessian.Object{}},
    map[string]{constant.GenericKey: "true"})
// inv.ActualMethodName() -> panic: interface conversion: interface {} is int, not string
```

### Reachability

In the default `DefaultServiceFilters` chain, `genericServiceFilter` (fixed by #3680) runs **before** `otelServerFilter` and short-circuits malformed invocations, so `ActualMethodName()` is not reached. This is therefore **not reachable by default**.

However, `otelServerFilter` is appended to the filter chain when `otel.tracing.enable=true` (`server/action.go:388-389`), and the filter order is user-configurable via the service/provider `filter` setting (`server/action.go:380`). If a user places `otelServerTrace` before `generic_service`, or omits `generic_service` from a custom chain, the panic becomes reachable on the only server-side call site (`filter/otel/trace/filter.go:83`).

The remaining 6 call sites are client-side (cluster invokers / loadbalance) and handle locally-constructed invocations, so they are not externally reachable.

### Affected call sites

| Location | Side | Reachable? |
|---|---|---|
| `filter/otel/trace/filter.go:83` (otelServerFilter.Invoke) | server | Only with non-default filter ordering + tracing enabled |
| `filter/otel/trace/filter.go:126` (otelClientFilter.Invoke) | client | No |
| `cluster/cluster/failover/cluster_invoker.go:64` | client | No |
| `cluster/cluster/failfast/cluster_invoker.go:49` | client | No |
| `cluster/cluster/forking/cluster_invoker.go:66` | client | No |
| `cluster/cluster/zoneaware/cluster_invoker.go:88` | client | No |
| `cluster/loadbalance/p2c/loadbalance.go:118` | client | No |

### Environment

- dubbo-go version: `develop` @ `9757e773`

Lcos-000:
### Description
Fixes #3683

`RPCInvocation.ActualMethodName()` (`protocol/invocation/rpcinvocation.go:89`) performed an unguarded type assertion on the first generic-call argument:

```go
func (r *RPCInvocation) ActualMethodName() string {
    if r.IsGenericInvocation() {
        return r.Arguments()[0].(string)   // panics if arg[0] is not a string
    }
    return r.MethodName()
}
```

`IsGenericInvocation()` only checks `len(Arguments)==3`, not the element types. A malformed `$invoke` with a non-string `arg[0]` reached this line and panicked with `interface conversion: interface {} is X, not string`.

This is the same class of bug fixed in #3680 for `genericServiceFilter.Invoke`. In the default `DefaultServiceFilters` chain, `genericServiceFilter` runs before `otelServerFilter` and short-circuits malformed invocations, so `ActualMethodName()` is not reached by default. However, the filter order is user-configurable, and if a custom chain places `otelServerTrace` before `generic_service` (or omits `generic_service`), the panic becomes reachable on the server side via `filter/otel/trace/filter.go:83`.

This PR replaces the bare assertion with comma-ok form. Since `ActualMethodName() string` is an interface method on `base.Invocation` with no error return, it falls back to `r.MethodName()` on type mismatch (instead of returning an error like #3680 did). The interface signature is unchanged; all 7 call sites are unaffected.

### Checklist

### DIFF
--- a/protocol/invocation/rpcinvocation.go
+++ b/protocol/invocation/rpcinvocation.go
@@ -86,7 +86,8 @@ func (r *RPCInvocation) ActualMethodName() string {
 	if r.IsGenericInvocation() {
 		// IsGenericInvocation only guarantees len(Arguments)==3, not the element types;
 		// guard the assertion so a malformed $invoke falls back to MethodName() instead of panicking.
-		mtdName, ok := r.Arguments()[0].(string)
+		args := r.Arguments()
+		mtdName, ok := args[0].(string)
 		if !ok {
 			return r.MethodName()
 		}
--- a/protocol/invocation/rpcinvocation.go
+++ b/protocol/invocation/rpcinvocation.go
@@ -84,10 +84,15 @@ func (r *RPCInvocation) MethodName() string {
 // ActualMethodName gets actual invocation method name. It returns the method name been called if it's a generic call
 func (r *RPCInvocation) ActualMethodName() string {
 	if r.IsGenericInvocation() {
-		return r.Arguments()[0].(string)
-	} else {
-		return r.MethodName()
+		// IsGenericInvocation only guarantees len(Arguments)==3, not the element types;
+		// guard the assertion so a malformed $invoke falls back to MethodName() instead of panicking.
+		mtdName, ok := r.Arguments()[0].(string)
+		if !ok {
+			return r.MethodName()
+		}
+		return mtdName
 	}
+	return r.MethodName()
 }
 
 // IsGenericInvocation gets if this is a generic invocation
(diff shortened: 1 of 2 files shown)


Alanxtl (approved):
lgtm

AlexStocks (approved):

AlexStocks:
please take a look of the comment from copilot.

Lcos-000:
i ve accepted 2 suggestions form copilot

PR #3705, opened 2026-08-26
refactor: replace Uber atomic with local wrappers

xiaobaicai66695:
## Summary

- add `common/dubboutil/atomic`, a `sync/atomic`-backed compatibility package for the Uber atomic types used by dubbo-go
- preserve the existing constructor and method names so production call sites only need import-path changes
- keep zero-value initialization concurrency-safe and keep copied wrappers sharing the initialized atomic state instead of copying standard-library atomic values
- replace all project source imports of `go.uber.org/atomic`, including the independent dubbogo-cli module
- remove `go.uber.org/atomic` as a direct dependency from both Go modules

Part of #3693.

## Implementation notes

The local wrappers cover `Bool`, `Int32`, `Int64`, `Uint32`, `Uint64`, `Pointer[T]`, `Duration`, `Float64`, and `Time`, including the Uber-compatible constructors, CAS aliases, swap/add/sub helpers, JSON methods, and string methods used by the project.

The wrapped values are standard-library typed atomics. A single low-level pointer CAS initializes their shared backing state on first use, preserving concurrent zero-value behavior without copying a typed atomic after first use.

`go.uber.org/atomic` can still appear as an indirect module dependency because existing third-party dependencies such as dubbo-getty and gost require it; dubbo-go production source no longer imports it directly.

## Validation

- `make fmt`
- `make check-fmt`
- `git diff --check`
- `go vet ./...`
- `go test -race ./common/dubboutil/atomic ./global ./graceful_shutdown ./protocol/base ./filter/adaptivesvc/limiter ./server ./remoting ./remoting/zookeeper ./registry`
- `cd tools/dubbogo-cli && go vet ./...`
- `cd tools/dubbogo-cli && go test ./...`
- related cluster, shutdown, server, remoting, registry, filter, and Triple packages were exercised

A full Windows `go test ./...` run also completed apart from existing environment-specific failures: Unix `/tmp` assumptions, Windows bind-error text assertions in Triple lifecycle tests, and etcd tests that expect no service on local port 2379.


AlexStocks:
please fix the file confliction

xiaobaicai66695:
> please fix the file confliction

done

PR #3714, opened 2026-08-29
fix(dubbo): accept bare-ms timeout attachment so failover retries keep the configured timeout

### LINKED ISSUE #3710: [Bug]: Failover retries lose the configured timeout and fall back to the hard-coded 3s (dubbo protocol)
### What happened?

With failover cluster (default) on the dubbo protocol, the 2nd and subsequent attempts
(retries) silently lose the configured timeout: it collapses to 0 and is replaced by
the getty client's hard-coded 3s default (`remoting/getty/getty_client.go#L229-L231`).
E.g. with `timeout=10s`, every retry aborts at ~3s with a premature read timeout.

Root cause: `getTimeout` writes and reads the `timeout` attachment in two formats.

- `protocol/dubbo/dubbo_invoker.go#L178` writes it as a bare millisecond integer: `"10000"`
- `protocol/dubbo/dubbo_invoker.go#L166` re-reads it with `time.ParseDuration`, which
  requires a unit → `ParseDuration("10000")` fails, the error is discarded, timeout = 0.

Failover makes this reachable on every retry because it passes the same invocation
object to each attempt (`cluster/cluster/failover/cluster_invoker.go#L88`).

### How can we reproduce it?

Failing unit test (fails on current `develop`):

```go
func TestGetTimeoutAcrossAttemptsOnSameInvocation(t *testing.T) {
	url, _ := common.NewURL("dubbo://127.0.0.1:20880/org.apache.dubbo.UserProvider?timeout=10s")
	invoker := NewDubboInvoker(url, nil)
	inv := invocation.NewRPCInvocation("GetUser", nil, nil)

	assert.Equal(t, 10*time.Second, invoker.getTimeout(inv)) // attempt 1: OK
	assert.Equal(t, 10*time.Second, invoker.getTimeout(inv)) // retry: FAILS, actual 0s
}
```

Suggested fix — make the read accept the bare-ms format it writes:

```go
if d, err := time.ParseDuration(attachTimeout); err == nil {
	timeout = d
} else if ms, err := strconv.Atoi(attachTimeout); err == nil {
	timeout = time.Duration(ms) * time.Millisecond
}
```

I'd be happy to submit a PR with the fix and the regression test above. Please assign this issue to me.

### Environment

dubbo-go 3.3.2 / current develop

eye-gu:
### Description
Fixes #3710

### Checklist

### DIFF
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -163,7 +163,12 @@ func (di *DubboInvoker) Invoke(ctx context.Context, ivc base.Invocation) result.
 func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration {
 	timeout := di.timeout                                                //default timeout
 	if attachTimeout, ok := ivc.GetAttachment(constant.TimeoutKey); ok { //check invocation timeout
-		timeout, _ = time.ParseDuration(attachTimeout)
+		// the timeout below is written back as bare milliseconds, so accept both formats
+		if d, err := time.ParseDuration(attachTimeout); err == nil {
+			timeout = d
+		} else if ms, err := strconv.Atoi(attachTimeout); err == nil {
+			timeout = time.Duration(ms) * time.Millisecond
+		}
 	} else { // check method timeout
 		methodName := ivc.MethodName()
 		if di.GetURL().GetParamBool(constant.GenericKey, false) {
--- a/protocol/dubbo/dubbo_invoker.go
+++ b/protocol/dubbo/dubbo_invoker.go
@@ -163,11 +163,10 @@ func (di *DubboInvoker) Invoke(ctx context.Context, ivc base.Invocation) result.
 func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration {
 	timeout := di.timeout                                                //default timeout
 	if attachTimeout, ok := ivc.GetAttachment(constant.TimeoutKey); ok { //check invocation timeout
-		// the timeout below is written back as bare milliseconds, so accept both formats
 		if d, err := time.ParseDuration(attachTimeout); err == nil {
 			timeout = d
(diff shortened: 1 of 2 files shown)


protocol/dubbo/dubbo_invoker.go
```diff
@@ -163,7 +163,12 @@ func (di *DubboInvoker) Invoke(ctx context.Context, ivc base.Invocation) result.
 func (di *DubboInvoker) getTimeout(ivc *invocation.RPCInvocation) time.Duration {
 	timeout := di.timeout                                                //default timeout
 	if attachTimeout, ok := ivc.GetAttachment(constant.TimeoutKey); ok { //check invocation timeout
-		timeout, _ = time.ParseDuration(attachTimeout)
+		// the timeout below is written back as bare milliseconds, so accept both formats
+		if d, err := time.ParseDuration(attachTimeout); err == nil {
+			timeout = d
+		} else if ms, err := strconv.Atoi(attachTimeout); err == nil {
+			timeout = time.Duration(ms) * time.Millisecond
```
AlexStocks: [P1] 裸毫秒转 time.Duration 前需要校验可表示范围

strconv.Atoi 只保证字符串能放进 int，不保证乘以 time.Millisecond 后仍能放进 time.Duration。64 位下附件值 9223372036855 会成功解析，但相乘后溢出为负的 -2562047h47m16.854551616s；Invoke 把这个值传给 Getty 后，timeout <= 0 分支又会替换成硬编码 3s，重新出现本 PR 要修复的提前超时。请改用 time.ParseDuration(attachTimeout + "ms")，或在 ParseInt 后显式检查 time.Duration 的毫秒上下界，仅在可表示时覆盖默认 timeout；同时补 max-valid-ms 与 max+1 的边界测试。

eye-gu: 已修复：改用 time.ParseDuration(attachTimeout + "ms") 解析裸毫秒，超范围值保持默认 timeout，并补了 max-ms 与 max+1 的边界测试

AlexStocks (approved):

PR #3732, opened 2026-09-17
fix(router): prevent data race on router cache in TestRouteCacheGenerationRace

bodapatisaikrishna:
## What is the purpose of the change

In `cluster/router/chain/chain_test.go`, `genCheckRouter` was created to mirror `TagRouter`'s cache guard under concurrency. However, unlike `TagRouter` which protects its cached reference using an atomic pointer (`atomic.Pointer[router.Cache]`), `genCheckRouter` stored `cache` in a plain field without atomic synchronization.

During concurrent `SetInvokers` (which rebuilds the cache and invokes `accessor.SetCache(...)`) and `Route()` (which reads `r.cache`), Go's race detector detected unsynchronized read and write access on `r.cache`, causing `TestRouteCacheGenerationRace` to fail under `-race` and requiring it to be added to `RACE_SKIP_TESTS` in `Makefile`.

## Brief changelog

- Update `genCheckRouter.cache` to `atomic.Pointer[router.Cache]`, ensuring safe concurrent reads in `Route()` and stores in `SetCache()`.
- Remove `TestRouteCacheGenerationRace` from `RACE_SKIP_TESTS` in `Makefile` so it is actively checked by `make test-race`.
- Format `cluster/router/script/instance/js_instance_test.go` so `make check-fmt` succeeds.

## Verifying this change

- Verified `go test -race -v -run TestRouteCacheGenerationRace ./cluster/router/chain/...` passes with zero race warnings.
- Ran 10 consecutive executions of `TestRouteCacheGenerationRace` under `-race` with 100% pass rate.
- Verified `make check-fmt` passes cleanly.


Alanxtl:
1. the target branch should be develop
2. pls fix the ci fail

bodapatisaikrishna:
Hi @Alanxtl, thanks for the review!

1. Updated the base branch to `develop`.
2. Rebased onto `upstream/develop` and resolved the CI format issue by removing the unintended formatting diff in `js_instance_test.go`. The patch now cleanly modifies only `cluster/router/chain/chain_test.go` (using `atomic.Pointer[router.Cache]` to avoid unsynchronized access) and `Makefile` (enabling `TestRouteCacheGenerationRace` under `-race`). `gofmt` and race tests pass cleanly.

Alanxtl (approved):
lgtm

AlexStocks (approved):

